Issue #10, opened 2022-05-05
Supports both Ethereum & Fuel signature recovery


Braqzen:
Similar to the [EIP-191 Signed Data Standard](https://eips.ethereum.org/EIPS/eip-191) the wallet should support recovery for both 

- [ ] #13
- [ ] #14

Braqzen:
@K1-R1 can the two linked issues in the description be closed on just one of them?

Issue #6, opened 2022-05-05
Can transfer native assets


Braqzen:
The multi-signature **wallet** should be able to transfer coins that have been sent to it to

- [ ] #8
- [ ] #9

PR #25, opened 2022-05-06
Using nonce to prevent duplicate execution

### LINKED ISSUE #5: Use a nonce to prevent double spends
Once a Tx has reached the required number of approvals it can be executed.

The nonce is meant to prevent multiple executions which could result e.g. having funds sent multiple times to an address until a contract is drained of an asset.
The execution should only be played out at most once and never again.

Braqzen:
Closes #5


Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
+
+members = [
+    "multisig-wallet",
+]
+
```
adlerjohn, line 6: I don't think we really want this to be a Rust workspace, since it's not actually a bunch of Rust projects. It's a bunch of Sway projects.

multisig-wallet/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+name = "multisig-wallet"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+
+[dependencies]
+std = { git = "https://github.com/FuelLabs/sway-lib-std", branch = "master" }
```
adlerjohn, line 8: Delete to use implicit `std` that's matching the compiler version used. There's no guarantee `master`'s stdlib compiles with any release version of `forc`. Also regen lockfile.

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,176 @@
+contract;
+
+use std::{
+    address::Address,
```
adlerjohn, line 4: Format.

adlerjohn (commented):
Could you add a CI workflow to build, check formatting, and test all Sway projects in this repo?

Edit: added in #26, which can be merged into this branch.

multisig-wallet/Forc.lock
```diff
@@ -0,0 +1,12 @@
+[[package]]
```
adlerjohn, line 1: Hmm. Actually, since we're using tagged releases of the compiler in CI, how about deleting Forc lockfiles and adding `Forc.lock` to the `.gitignore`? Since these projects are meant to showcase maintained examples rather that being deployed to production as-is, keeping up with dependency updates is a desirable property.

adlerjohn (commented):
Add a `.gitattributes` file for Sway, ref: https://github.com/FuelLabs/sway/blob/460d7afd30f4dba845371c96bed0ec887ca5e718/.gitattributes

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Syntax highlighting of sway files as rust
+*.sw linguist-language=Rust
+
+# Syntax highlighting of ir files as LLVM
+*.ir linguist-language=LLVM
```
adlerjohn, line 5:
```suggestion
```

multisig-wallet/.gitignore
```diff
@@ -0,0 +1,3 @@
+out
+target
+Forc.lock
```
adlerjohn, line 3: Redundant to the top-level gitignore
```suggestion
```

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,160 @@
...
+        storage.sentinel = SentinelTransaction {
+            to: hash_u64(1, HashMethod::Sha256), value: hash_u64(2, HashMethod::Sha256), data: hash_u64(3, HashMethod::Sha256), executed: hash_u64(4, HashMethod::Sha256), approvals: hash_u64(5, HashMethod::Sha256)
+        };
+
+        storage.threshold = threshold;
+        storage.state = 1;
+        true
+    }
+
+    fn submit(destination: ContractId, value: u64, data: b256) -> bool {
```
adlerjohn, line 49: Starting simple with just asset transfers, you need an asset ID, an amount, and a destination which could either be a contract or an address. And no data.

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,160 @@
...
+            store(hash_pair(nonce_hash, storage.sentinel.value, HashMethod::Sha256), value);
+            store(hash_pair(nonce_hash, storage.sentinel.data, HashMethod::Sha256), data);
+        } else {
+            panic(0);
+        };
+
+        true
+    }
+
+    fn approve(nonce: u64) -> bool {
```
adlerjohn, line 70: Approvals should be done over transaction hashes (which includes a nonce) rather than a bare nonce. The latter is vulnerable to re-orgs changing what the nonce represents.

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,160 @@
...
+        assert(storage.state == 1);
+        assert(nonce != 0 && nonce <= storage.nonce);
+
+        let sender: Result<Sender, AuthError> = msg_sender();
```
adlerjohn, line 74: Signature authentication should be used instead of transaction-level authentication. This would allow multiple signatures to be submitted in a single tx.

multisig-wallet/src/main.sw
Braqzen: Will be done in [Can transfer native assets](https://github.com/FuelLabs/sway-applications/issues/6)

multisig-wallet/src/main.sw
Braqzen: Are you referring to [Supports both Ethereum & Fuel signature recovery](https://github.com/FuelLabs/sway-applications/issues/10) for auth or are you referring to something else?

adlerjohn: No, I mean rather than using `msg_sender` (of which there can only be one), use a signature over the action. That way you can include as many signatures as you want.

Braqzen: Doesn't that mean that we need to be able to recover the signature in order to match it against some owner in the multisig?

adlerjohn: Yes. You can use https://github.com/FuelLabs/sway/blob/a34792b696dac1c70984c8bca66b7412fd0482b2/sway-lib-std/src/ecr.sw#L31 for that. For now you can just used Fuel signatures and leave EIP-191 signature support for a future PR.

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,151 @@
...
+    /// let tx_hash = contract.get_transaction_hash(destination_contract, value, data).call.await.unwrap();
+    /// ```
+    fn get_transaction_hash(to: ContractId, value: u64, data: b256) -> b256 {
+        _assert_is_initialized(storage.nonce);
+        // TODO: data > b256?
+        // TODO: this is probably not following the EIP-191 signing standard. What do?
+        let to_hash = hash_value(to.value, HashMethod::Keccak256);
+        let data_hash = hash_value(data, HashMethod::Keccak256);
+        let value_hash = hash_u64(value, HashMethod::Keccak256);
+        let nonce_hash = hash_u64(storage.nonce, HashMethod::Keccak256);
```
adlerjohn, line 124: EIP-191 _signature_ support will be done in the standard library, separately. But that's orthogonal to how to hash the message that's signed. For that, using SHA-256.

simonr0204:
Is this still a WIP or is it waiting for a review ?

adlerjohn (approved):

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// Executes a Tx if the required signatures meet the restrictions on ownership and threshold approval
+    ///
+    /// # Panics
+    ///
+    /// - When the constructor has not been called to initialize the contract
+    /// - When the public key cannot be recovered from a signature
+    /// - When the signer is not an owner
+    /// - When the signatures are not in ascending order (0x1 < 0x2 < 0x3...)
```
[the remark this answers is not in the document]
nfurfaro: I see how the comparison is done below, which must mean Ord is implemented for `B512`.

But what would force `signer1.value < signer2.value` to be the case?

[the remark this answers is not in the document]
nfurfaro: Ok. Same question then I guess, how are the addresses being in ascending order enforced when an address is just a hash of a pub key derived from randomness ultimately?

multisig-wallet/src/main.sw
```diff
@@ -0,0 +1,141 @@
...
+    /// Initializes the contract by setting the owners, owner weightings and initializes the nonce
+    ///
+    /// # Panics
+    ///
+    /// - When the constructor is called more than once
+    fn constructor(owner1: Address, owner2: Address) -> bool {
+        require(storage.nonce == 0, Error::CannotReinitialize);
+
+        // TODO: when vectors are implemented change owners to be a Vec<Address>
+        store(owner1.value, true);
```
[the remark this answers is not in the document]
Braqzen: `store` takes a `b256` as the key and I need to know the truth value of whether the value has been marked as an owner.

multisig-wallet/src/main.sw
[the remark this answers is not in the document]
Braqzen: > I see how the comparison is done below, which must mean Ord is implemented for `B512`.
> 
> But what would force `signer1.value < signer2.value` to be the case? I'm sure it's here, just not seeing it at first pass. The incrementing nonce must be a part of this, but the signature is otherwise not sequential I would think.

My reference implementation was to follow [this execute function](https://github.com/SilentCicero/MultiSignatureWallet#reference-implementation-solidity)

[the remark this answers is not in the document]
adlerjohn: You can enforce they're in ascending order because they're public.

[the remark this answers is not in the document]
Braqzen: The signatures would need to be placed in the function in such an order that the recovered address values are ascending in order otherwise it would error out because of a possible single owner trying to execute the Tx with their own signature multiple times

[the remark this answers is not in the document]
adlerjohn: Oh, if there are two sigs then indeed an inequality is sufficient. But eventually this will support an arbitrary number of signers.

PR #26, opened 2022-05-07
Add CI

adlerjohn:
Adds CI workflow. Note that a matrix of projects is used where you have to manually add projects. This should be okay since we're only going to have a relatively small number of projects.

To merge into https://github.com/FuelLabs/sway-applications/tree/braqzen-%235 #25

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,93 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  release:
+    types: [published]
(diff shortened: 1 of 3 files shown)


Braqzen (approved):
Given that the `env` section uses `FORC_VERSION: 0.12.1` and we release frequently I assume that we might be updating this CI often?

adlerjohn:
Yes, but that's fine. Minor (as opposed to patch) releases of `forc` have been on a cadence of once every 2 weeks or thereabouts.

Issue #1, opened 2022-05-05
Multi-Signature Wallet Application


Braqzen:
This issue is meant to document the various sub tasks that are required in order to create a fully functional multi-signature wallet written in Sway.

There are various components and they are grouped into categories.

**The Basics**

- [x] #2
- [x] #3
- [x] #4
- [x] #5
- [x] #6
- [x] #7

**Ethereum Signature Capability**

- [x] #10
- [x] #11
- [x] #12

**Frontend**

- [ ] #18

**Contract Calls**

- [x] #22

**Documentation**

- [x] #23
- [x] #24

bitzoic:
Closing in favor of https://github.com/FuelLabs/sway-applications/issues/18

PR #27, opened 2022-05-21
added foundation for `sway-tac-toe`

eureka-cpu:
At present this only uses JS for its functionality, will change this to using Sway in a future commit. This should also solve the one known problem this has, which is that the logic allows for clicking the same square twice and instead of maintaining the correct player it's possible for the same user to have multiple turns back to back. This version also doesn't implement multi-device interactivity, which is something I'd assume we'd want if anyone is to actually play with someone else.


Braqzen (changes requested):
1. I'd like to see an issue which details the information regarding the application similar to #1 or perhaps any other "sub issue" so that there is some documentation instead of a whole new piece of work hitting master out of nowhere.
2. When I pull in your work it all gets added directly under "sway-applications" however there will be many applications for example my multisig work is under `sway-applications/multisig`. Can you create a directory under "sway-applications" and move all of your work into that so it's not at the top level e.g. `sway-applications/tic-tac-toe`
3. Can you add some documentation as to how this code should be run? When looking through the files the index.html has a little bit of info however when I run `npm start` it says that I do not have `react-scripts` or something to that extent. I _could_ look into how to set up a react project myself but that shouldn't be a thing that an outsider does and instead it should be documented so that anyone can follow the steps and get your project running.
4. A little nitpic, the css is small but I prefer [scss](https://sass-lang.com/documentation) (sass is fine too). Is it a good idea to move towards that now before the project grows or will it stay sufficiently small that the little bit of css is fine and adding scss/sass is unnecessary?
5. The App.js file has a massive "src" image. Is that needed / can it be replaced with a loading from a file or something?
6. Since you want to add sway into this instead of it being a pure JS application can you mark it as a Draft until the code is ready to go into master?

eureka-cpu:
>     4. A little nitpic, the css is small but I prefer [scss](https://sass-lang.com/documentation) (sass is fine too). Is it a good idea to move towards that now before the project grows or will it stay sufficiently small that the little bit of css is fine and adding scss/sass is unnecessary?
>     5. The App.js file has a massive "src" image. Is that needed / can it be replaced with a loading from a file or something?

4. As far as the CSS goes, you're 100% right. Once this project becomes bigger a definitive choice will need to be made on this, for now since the amount of styling is minimal and there is only one `component` I kept it all in once place (`App.css`) and just used default CSS, but this is not best practice.

5. Also correct, the original image is from `SwaySwap` and somewhere in rendering it is turned into a `base64` image. Since I was just trying to get this up quickly, I went ahead and used the `base64` version instead but it does look pretty ugly and will definitely be changed.

eureka-cpu:
>     2. When I pull in your work it all gets added directly under "sway-applications" however there will be many applications for example my multisig work is under `sway-applications/multisig`. Can you create a directory under "sway-applications" and move all of your work into that so it's not at the top level e.g. `sway-applications/tic-tac-toe`

2. Would maybe `sway-applications/games/sway-tac-toe` suffice?

Braqzen:
> > ```
> > 2. When I pull in your work it all gets added directly under "sway-applications" however there will be many applications for example my multisig work is under `sway-applications/multisig`. Can you create a directory under "sway-applications" and move all of your work into that so it's not at the top level e.g. `sway-applications/tic-tac-toe`
> > ```
> 
> 2. Would maybe `sway-applications/games/sway-tac-toe` suffice?

If there are going to be multiple mini-games then sure. This project could be one of those games. I have no issues with that.

eureka-cpu:
> > > ```
> > > 2. When I pull in your work it all gets added directly under "sway-applications" however there will be many applications for example my multisig work is under `sway-applications/multisig`. Can you create a directory under "sway-applications" and move all of your work into that so it's not at the top level e.g. `sway-applications/tic-tac-toe`
> > > ```
> > 
> > 
> > 
> > 2. Would maybe `sway-applications/games/sway-tac-toe` suffice?
> 
> If there are going to be multiple mini-games then sure. This project could be one of those games. I have no issues with that.

I think the intention is to have a few at least. Checkers, maybe chess eventually (:

Braqzen:
This has been a draft for a long time. I'm thinking of closing it.

eureka-cpu:
Closing this for now until the Sway contract is ready, feel free to reopen when ready @matt-user @Rubyt0x

PR #38, opened 2022-05-25
Added approval threshold for execution

### LINKED ISSUE #3: Enforces a threshold on the number of approvals required to execute a transaction
The **multi-signature wallet** should use a threshold on the number of approvals in order to allow the execution of a Tx.

This means that a value should be set in storage and only when the threshold for a specified Tx (by the signers) has been reached it will become executable.

Braqzen:
Closes #3


adlerjohn (commented):
Looks fine to me, but would like @bitzoic and @matt-user to review.

bitzoic (approved):
Looks good!

matt-user (approved):
Looks good

PR #40, opened 2022-05-26
Added weightings to approval

### LINKED ISSUE #4: Weighted signers
A weight of a signer is the number of approvals that they "give" when they approve a Tx.

A basic application is where each signer has a weight of 1 approval (their own) and thus if there are 5 signers with an approval threshold of 3 then 3 signers must approve the Tx.

This issue specifies that a signer can have a greater approval than 1 and thus it would be possible to approve a Tx (in the aforementioned example) with only 2 signers.

Braqzen:
Closes #4

### DIFF
--- a/multisig-wallet/src/main.sw
+++ b/multisig-wallet/src/main.sw
@@ -65,7 +65,7 @@ impl MultiSignatureWallet for Contract {
     fn constructor(owner1: Address, owner2: Address, owner1_weight: u64, owner2_weight: u64, threshold: u64) -> bool {
         require(storage.nonce == 0, Error::CannotReinitialize);
         require(storage.threshold != 0, Error::ThresholdCannotBeZero);
-        
+
         // TODO: when vectors are implemented change owners to be a Vec<Address>
         require(owner1_weight != 0, Error::WeightingCannotBeZero);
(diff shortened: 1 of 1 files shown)


multisig-wallet/src/main.sw
```diff
@@ -91,27 +101,40 @@ impl MultiSignatureWallet for Contract {
             Result::Ok(address) => address.value, _ => revert(42), 
         };
 
-        require(get::<bool>(signer1) && get::<bool>(signer2), Error::NotAnOwner);
         require(~b256::min() < signer1 && signer1 < signer2, Error::IncorrectSignerOrdering);
 
+        let signer1_weight = get::<u64>(signer1);
+        let signer2_weight = get::<u64>(signer2);
+
+        require(signer1_weight != 0 && signer2_weight != 0, Error::NotAnOwner);
```
matt-user, line 109: Would here be a good place to use the is_owner function defined below?  Or is the is_owner function more for external calls?

matt-user (approved):
Looks good.  I just left a couple of questions regarding sway and the is_owner function.

bitzoic (approved):

PR #41, opened 2022-05-26
Added transfer of assets to outputs & contracts

### LINKED ISSUE #6: Can transfer native assets

### LINKED ISSUE #8: An address of a user (wallet / output)
Transferring assets can be done to outputs and contracts.

This issue is meant to support transfers from the contract to an output (address).

### LINKED ISSUE #9: Contract
Transferring assets can be done to outputs and contracts.

This issue is meant to support transfers from the contract to another contract.

Braqzen:
Closes #8 
Closes #9 
Closes #6


multisig-wallet/src/main.sw
```diff
@@ -4,29 +4,34 @@ use std::{
...
+    storage::{get, store},
+    token::{force_transfer, transfer_to_output}
 };
 
 use core::num::*;
 
 abi MultiSignatureWallet {
     fn constructor(owner1: Address, owner2: Address, owner1_weight: u64, owner2_weight: u64, threshold: u64) -> bool;
-    fn execute_transaction(to: ContractId, value: u64, data: b256, signature1: B512, signature2: B512) -> bool;
+    fn execute_transaction(to: Sender, value: u64, data: b256, signature1: B512, signature2: B512) -> bool;
```
bitzoic, line 23: Maybe I'm missing something, but get_transaction_hash and _get_transaction_hash both return a b256. Why do execute_transaction and transfer take b512 signatures?

Braqzen: The hashing functions take some parameters and produce a hash which is meant to be signed with the signature of the owner(s). More information regarding the signature structure can be found in [EIP-2098: Compact Signature Representation](https://eips.ethereum.org/EIPS/eip-2098).

Braqzen: Oh, and as for the "why", it is to make sure that when the public key is recovered from the signature that key is regarded as an owner in the multisig. This is to prevent random people from interacting with the contract.

multisig-wallet/src/main.sw
```diff
@@ -147,11 +147,51 @@ impl MultiSignatureWallet for Contract {
...
+        // Hardcoded value, that passes the checks above, until the loop below is unblocked
+        let approval_count = signer1_weight + signer2_weight;
+
+        require(storage.threshold <= approval_count, Error::ApprovalThresholdNotReached);
+
+        storage.nonce = storage.nonce + 1;
+        log(tx_hash);
+
+        match to {
+            Sender::Address(address) => transfer_to_output(value, asset_id, address), Sender::ContractId(contract) => force_transfer(value, asset_id, contract),
```
matt-user, line 191: Why do we need a force_transfer branch here?  Is it because not all contracts may have implemented some sort of transfer/receive function?

multisig-wallet/src/main.sw
```diff
@@ -147,11 +147,51 @@ impl MultiSignatureWallet for Contract {
...
+        // TODO: change to Vec<B512> once implemented and then iterate instead of hardcoding length
+        let signer1: b256 = match ec_recover_address(signature1, tx_hash) {
+            Result::Ok(address) => address.value, _ => revert(42), 
+        };
+
+        let signer2: b256 = match ec_recover_address(signature2, tx_hash) {
+            Result::Ok(address) => address.value, _ => revert(42), 
+        };
+
+        require(~b256::min() < signer1 && signer1 < signer2, Error::IncorrectSignerOrdering);
```
matt-user, line 175: Why is it required that the recovered address of signer1 is less than signer2?  Also could we implement some check here say:
```
if signer1 < signer2 {
    swap(signer1, signer2)
}
```

Braqzen: Otherwise you can submit your own signature multiple times to game the system and send any Tx through that you want. Making sure that they are increasing in size means that they cannot be 0 and cannot be the same address and thus likely to come from different owners... or at least the same owner but with different addresses

multisig-wallet/src/main.sw
Braqzen: There are two functions

1. transfer_to_output()
    - Is used to transfer to an `Address`
2. force_transfer()
    - Is used to transfer to a `ContractId`

They are equivalent but the name `force_transfer` might be confusing since `transfer_to_output` is also forceful and cannot be stopped. Both of those functions act in the same way as if I were to drop a letter into your mail box.

Braqzen: To my knowledge, in Sway contracts do not have a `receive` function like in `Solidity`

nfurfaro: @matt-user ensures that the receiving contract cannot possibly execute any code upon receipt of the coins. This will work for any contract, unlike calling a `receive` function and forwarding coins for example, where the receiving contract may not even implement a `receive` function, or may choose to revert the call, etc...
Forwarding coins with a call can be done visa any function that doesn't explicitly forbid it, but with `force_transfer` we don't even need to call a function on a contract.

adlerjohn: Note that without an explicit fallback function, if you try to call a contract in Sway without a function selector (or with a selector that's not in the contract's ABI), the ABI dispatcher will _revert_. This means there's no much point in using call to transfer assets to a contract here. If you want to specifically call a function to receive funds, then you can call it.

.github/workflows/ci.yml
```diff
@@ -50,14 +51,67 @@ jobs:
       - name: Run Cargo.toml linter
         run: git ls-files | grep Cargo.toml$ | xargs --verbose -n 1 cargo-toml-lint
 
-  build:
+  escrow:
```
adlerjohn, line 54: The above comment will also obviate the need for this, which is unmaintainable.

README.md
```diff
@@ -6,9 +6,6 @@
     <a href="https://github.com/FuelLabs/sway-applications/actions/workflows/ci.yml" alt="CI">
         <img src="https://github.com/FuelLabs/sway-applications/actions/workflows/ci.yml/badge.svg" />
     </a>
-    <a href="https://crates.io/crates/forc/0.14.4" alt="forc">
```
adlerjohn, line 9: Ditto

escrow/README.md
```diff
@@ -0,0 +1,22 @@
+<p align="center">
+    <img src="./.docs/escrow-logo.png" height="120">
+</p>
+
+<p align="center">
```
adlerjohn, line 5: This is an update to escrow. which should be in a separate PR.

.github/workflows/ci.yml
```diff
@@ -10,7 +10,8 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  FORC_VERSION: 0.14.4
+  ESCROW_FORC_VERSION: 0.14.5
+  MULTISIG_FORC_VERSION: 0.15.1
```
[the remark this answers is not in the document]
Braqzen: The projects are all meant to be independent of each other. If 1 project is dormant for a while because it doesn't require any updates and then other projects get updates that contain breaking changes then using the same version will break other projects which is entirely not the point. They should all be self contained and use their own versions otherwise an update to 1 project may require updates to every other project when it is unnecessary

.github/workflows/ci.yml
Braqzen: I agree that this is spammy but I'm unsure of how to use distinct versions in the matrix. Perhaps a script that outputs the version and pipes it to the required command... not ideal

.github/workflows/ci.yml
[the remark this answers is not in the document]
adlerjohn: No, the entire repo should be of projects that all work with the latest version of the compiler. It wouldn't be "an update to 1 project," it would be an update to the entire repo.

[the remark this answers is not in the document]
adlerjohn: Thought experiment: a month from now, what use is a contract that only compiles with `v0.14.5`? Almost nothing. What use is the same contract that compiles with the version of the compiler at that time? Significant!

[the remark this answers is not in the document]
Braqzen: The point is to eventually containerise each application. They would be entirely independent from everything else. The version can be locked in the container until it requires an update otherwise it can remain there as it is on the older version of the compiler. There is no point in creating additional work and constantly updating something that does not need to be updated. As more applications are added, and with the rate of change of the compiler, any 1 small change will require days to weeks of work just to bump a version in 1 application that needs it. It's just wrong.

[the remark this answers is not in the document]
Braqzen: Unless ofc you expect to maintain these as actual production (and in use) applications and not just examples that a user is meant to reference, clone and update for their own use.
If you want to keep updating the apps and redeploying the code for each update... okay. That seems wasteful but then it makes sense to keep everything up to date.

multisig-wallet/src/main.sw
```diff
@@ -1,172 +1,195 @@
...
+    }
 
-        //     let signer_weight = get::<u64>(signer);
-        //     require(signer_weight != 0, Error::NotAnOwner);
+    /// Transfers assets to outputs & contracts if the signatures meet the threshold requirement
+    ///
+    /// # Panics
+    ///
+    /// - When the constructor has not been called to initialize the contract
+    /// - When the balance of the asset being sent is less than the balance in the contract
```
matt-user, line 110: Shouldn't this be "When the balance of the asset in the contract is less than the amount being transferred." ?

matt-user (changes requested):

multisig-wallet/src/main.sw
Braqzen: How about now?

matt-user (approved):

Issue #47, opened 2022-05-27
General purpose NFT contract


bitzoic:
**WIP**

A general purpose NFT contract that will have the following attribute:

- [x] the contract should represent a collection of NFTs, not just a single token
- [x]  every token minted should have a unique, immutable token id
- [x]  there should be a way to find the owner of any token given its id.
- [x]  access control for minting - the contract should store an Identity as the minter, which allows for either an address or a minter contract to be set as the entity with minting rights
- [x]  an event should be emitted when approvals are made, an NFT is minted or burned, and when transfers are made

The contract ABI will follow the format below:
```
fn allowMint(minter: Identity) -> bool;
fn approve(to: Identity, token_id: b256) -> bool;
fn balance_of(owner: Identity) -> u64;
fn burn(token_id: b256) -> bool ;
fn constructor(owner: Identity, access_control: bool, token_count: u64) -> bool;
fn getApproved(token_id: b256) -> Identity;
fn getTotalSupply() -> u64;
fn isApprovedForAll(owner: Identity, operator: Identity) -> bool;
fn mint(to: Identity, amount: u64) -> bool ;
fn owner_of(token_id: b256) -> Option<Identity>;
fn setApprovalForAll(owner: Identity, operator: Idenity) -> bool;
fn transferFrom(from: Identity, to: Identity, token_id: b256) -> bool;
```

nfurfaro:
A few things that I'd like to see in an NFT contract to help keep it useful and flexible:
- [ ] the contract should represent a collection of NFTs, not just a single token
- [ ] every token minted should have a unique, immutable token id
- [ ] there should be a way to find the owner of any token given its id.
- [ ] access control for minting - the contract should store an `Identity` as the `minter`, which allows for either an address or a minter contract to be set as the entity with minting rights

nfurfaro:
A good next step for this would be to pin down what the `abi` looks like.

The specific implementation can follow, and will probably be able to benefit from some cool new functionality in Sway (ie: vec, improved storage interface).

nfurfaro:
Something along the lines of this as a start (may not be complete yet):
```
library nft;

use std::identity::Identity;


abi NFT {
    fn mint(to: Identity, amount: u64) ;
    fn burn(token_id: b256) ;
    fn owner_of(token_id: b256) -> Option<Identity>;
    fn balance_of(owner: Identity) -> u64;
    fn transferFrom(from: Identity, to: Identity, token_id: b256)
}
```
[...]

PR #48, opened 2022-05-27
Basic nft contract

bitzoic:
Note that this is a one-of-one NFT library. Support for tokenized NFTs will require mapping support.

Please be nit-picky in the review, the more feedback the better!


nft/simple_nft/tests/harness.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    // TODO: Test for wallet with 0 address
+    // #[tokio::test]
+    // #[should_panic]
+    // async fn panics_when_minted_to_zero_address() {
+    //     let (deploy_wallet, owner1, owner2) = setup().await;
+
+    //     let empty_address = Metadata {
+    //         nft: owner1.nft,
+    //         wallet: LocalWallet::, // Something needs to happen here
```
Braqzen, line 109: A wallet cannot have `0 addresses`

nft/simple_nft/tests/harness.rs
```diff
@@ -0,0 +1,393 @@
...
+            .value
+        );
+    }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn panics_when_not_initalized() {
+        let (_deploy_wallet, owner1, _owner2) = setup().await;
+
+        assert!(
```
Braqzen, line 171: no need to assert on a panic. the `unwrap()` will panic so the `.value` is not needed

Braqzen: same in other places

nft/simple_nft/tests/harness.rs
```diff
@@ -0,0 +1,393 @@
...
+            owner1
+                .nft
+                .transfer(owner1.wallet.address(), owner2.wallet.address())
+                .call()
+                .await
+                .unwrap()
+                .value
+        );
+    }
+}
```
Braqzen, line 393:
```suggestion
}

```

nft/nft_lib/src/main.sw
```diff
@@ -0,0 +1,139 @@
...
+///
+/// The function will panic when:
+/// - The NFT has already been initalized
+/// - The new owner is the 0 address
+pub fn mint_to_address(new_owner: Address) -> bool {
+    require(storage.state == 0, Error::OwnerAlreadySet);
+    require(new_owner.value != NATIVE_ASSET_ID, Error::AddressCannotBeZero);
+
+    storage.data = Data {
+        owner: new_owner, approved: ~Address::from(NATIVE_ASSET_ID)
```
Braqzen, line 44: I don't understand why the 0th address is able to approve things

nft/nft_lib/src/main.sw
```diff
@@ -0,0 +1,139 @@
...
+// TODO: Add mappings when supported for token ids
+
+struct Data {
+    owner: Address,
+    approved: Address,
+}
+
+storage {
+    data: Data,
+    state: u64,
```
matt-user, line 29: Can we make this an enum?  It would be more readable.

nft/nft_lib/src/main.sw
matt-user, line 44: Isn't NATIVE_ASSET_ID alreadt an address?  Do we need the ~Address::from part here?

nft/nft_lib/src/main.sw
```diff
@@ -0,0 +1,139 @@
...
+/// # Panics
+///
+/// The function will panic when:
+/// - The sender is not the owner
+/// - The NFT has not be initalized
+pub fn burn() -> bool {
+    require(storage.state == 1, Error::NFTNotInitalized);
+
+    let sender: Result<Sender, AuthError> = msg_sender();
+    if let Sender::Address(address) = sender.unwrap() {
```
matt-user, line 61: I think you can turn this into a match.  You can checkout the mutlisig for an example.  Although I don't know if match if preferable here.

Braqzen (changes requested):
Overall, I'm not a fan of this implementation at all and that is being said while disregarding all of the comments that I have made below.

This library doesn't really achieve anything and it acts somewhat like a contract. Tbh I'm surprised that `storage` works in the library. I have not tried that before.

This library, if it is to be just a library, doesn't need to be in sway-applications and instead belongs more in the standard lib but then again we already have equivalent functions that actually transfer etc.
Again, it doesn't really achieve what it sets out to do.

I would change the entire implementation to be a single contract which has an owner set. This contract is able to mint 1 coin to some address but first it needs to check if it currently has an NFT / if it has ever received one (because they may have transferred it).
In order to mint they need to send some asset into the NFT and only the owner can withdraw that asset (basically purchase the NFT from the contract). 

If you do this then you have a contract that can receive some asset and in turn mint a single coin to an address. This is still hardly an NFT but at least from this you can turn it into an application with a front end.

We can discuss this further.

Oh, and there is not CI added so we do not know if this project builds and the tests pass.
Moreover, the linked issue needs a lot more detail in order for it to become a fully fledged application. A simple ticket / issue is insufficient in the description of an entire application.

nft/nft_lib/src/main.sw
```diff
@@ -0,0 +1,139 @@
...
+///
+/// # Panics
+///
+/// The function will panic when:
+/// - The sender is not the owner
+/// - The owner is the to address
+/// - The NFT has not been initalized
+pub fn set_transfer_approval(to: Address) -> bool {
+    require(storage.state == 1, Error::NFTNotInitalized);
+
```
matt-user, line 127: Probably add a requirement that the to address cannot be the zero address similar to what you did in the mint function.

matt-user (changes requested):
Nice work!

nft/nft_lib/src/main.sw
Braqzen: It is not an address, it's a [constant](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/constants.sw#L3)

nft/nft_lib/src/main.sw
Braqzen: I don't believe that an enum makes sense here

nft/nft_lib/src/main.sw
Braqzen: Yes, matching is preferable

nft/nft_lib/src/main.sw
adlerjohn: The Sway compiler can't yet do constant evaluation, so constants (such as `NATIVE_ASSET_ID`) must be primitives (such as `b256` in the case of `NATIVE_ASSET_ID`).

nft/nft_lib/src/main.sw
```diff
@@ -0,0 +1,139 @@
+library nft_lib;
```
adlerjohn, line 1: Libraries dealing with storage as @Braqzen pointed out aren't really well-supported yet. Just making this into a template contract that any can fork, rather than a library that needs to be imported, would be better for now.

nfurfaro: Agreed, we should make this into a contract rather than a lib.

nfurfaro: Agree, let's make this a contract rather than a library.

nfurfaro:
@bitzoic There are a few things I think are key here that are not yet implemented or can't be done yet due to missing features like vec. 
I've added a few things I'd like to see in a general-purpose NFT contract to this issue: https://github.com/FuelLabs/sway-applications/issues/47 
I think we should keep this in draft status until we hash out the details in the linked issue.

nft/nft_lib/src/main.sw
matt-user: My thinking is that an enum may be preferable to track the state of the contract instead of using integers.  So instead of state being 0 to represent the uninitialized state it would be like State::Uninitialized.  Or something like that.

Braqzen: enums are not supported in storage the last time I have checked so this is a workaround for now

nft/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = "0.14"
+fuels-abigen-macro = "0.14"
+tokio = { version = "1.12", features = ["rt", "macros"] }
+fuel-tx = "0.12.1"
```
adlerjohn, line 12: Use major.minor instead of major.minor.patch for deps, so that you can automatically pick up patches.
```suggestion
fuel-tx = "0.12"
```

nft/nft_lib/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["Cam Fuel"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuel-gql-client = { version = "0.6", default-features = false }
+fuel-tx = "0.9"
+fuels = "0.12"
+fuels-abigen-macro = "0.12"
+tokio = { version = "1.12", features = ["rt", "macros"] }
```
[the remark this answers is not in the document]
adlerjohn: FYI now you can use `fuelup`: https://fuellabs.github.io/sway/latest/introduction/installation.html#installing-from-pre-compiled-binaries

nft/src/main.sw
```diff
@@ -0,0 +1,488 @@
...
+
+    /// The state of the contract. Can either be initalized or not
+    state: u64,
+
+    /// The total number of tokens that have been minted
+    token_count: u64,
+
+    /// The price to purchase a single token. Can only be set on
+    /// the initalization of the contract
+    token_price: u64,
```
nfurfaro, line 69: I don't think the general-purpose base contract needs to have a notion of token price.
Not all NFTs needs be for sale, and the pricing logic is easily added in custom implementation s that require it.

nfurfaro: To further that thought, even when the token has a price, the price may be a hardcoded constant, or may be provided by an oracle, or may be calculated using some flavour of bonding-curve, etc...

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    fn is_approved_for_all(owner: Identity, operator: Identity) -> bool {
+        require(storage.state != 0, InitError::NFTNotInitalized);
+
+        let hash: b256 = get_owner_operator_hash(owner, operator);
+        storage.operator_approval.get(hash)
```
Braqzen, line 267:
```suggestion
        storage.operator_approval.get(sha256(owner, operator))
```

Braqzen: Should there be validation on the owner / operator here?

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /// Mints an NFT
+    ///
+    /// # Panics
+    /// 
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized 
+    /// - The amount is set to 0
```
Braqzen, line 276: It doesn't look like you need to block the user from minting 0. I would allow it, let them waste their gas

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    /// - More NFTs than supply is minted
+    /// - The to Identity is not valid
+    /// - The sender is not approved to mint
+    /// - The sender sent the wrong asset
+    /// - The sender did not pay enough tokens
+    fn mint(to: Identity, amount: u64) -> bool {
+        require(storage.state != 0, InitError::NFTNotInitalized);
+        require(amount != 0, InputError::MintAmountCannotBeZero);
+        require(
+            storage.token_supply >= (storage.token_count + amount),
```
Braqzen, line 286:
```suggestion
            storage.token_count + amount <= storage.token_supply, 
```

Braqzen: I think you should consider if overflows are something to be considered here and if this should be rewritten. cc @nfurfaro

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            (storage.access_control && storage.allowed_minters.get(sender)), 
+            AccessError::SenderDoesNotHaveAccessControl
+        );
+
+        /// Mint as many tokens as the sender has paid for
+        let mut i = 0;
+        while i < amount {
+            /// Create a new token id in sequential order
+            let token_id: u64 = storage.token_count + 1;
+            let empty_identity: Identity = Identity::Address(~Address::from(NATIVE_ASSET_ID));
```
Braqzen, line 305:
```suggestion
            let empty_identity = ~Address::from(NATIVE_ASSET_ID);
```

Shouldn't this work?

Braqzen: Also, consider where this is used. If the 0x000 address becomes a contract that can perform work then this may be an issue. There has been some mention of this

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+
+            /// Create the metadata for this new token with the owner 
+            let meta_data: MetaData = MetaData {
```
Braqzen, line 308:
```suggestion
            let meta_data = MetaData {
```

You don't need to type everything unless the compiler cannot figure it out or you're doing something confusing that you want to make easier to understand for the reader although at that point you should reconsider what you are doing

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+        );
+
+        /// Mint as many tokens as the sender has paid for
+        let mut i = 0;
+        while i < amount {
+            /// Create a new token id in sequential order
+            let token_id: u64 = storage.token_count + 1;
+            let empty_identity: Identity = Identity::Address(~Address::from(NATIVE_ASSET_ID));
+
+            /// Create the metadata for this new token with the owner
```
Braqzen, line 307:
```suggestion
            // Create the metadata for this new token with the owner 
```

`///` is for documentation, `//` is for comments

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            /// Create the metadata for this new token with the owner 
+            let meta_data: MetaData = MetaData {
+                owner: to, approved: empty_identity
+            };
+            storage.meta_data.insert(token_id, meta_data);
+            storage.owners.insert(to, token_id);
+            
+            /// Increase the balance of the new owner
+            let mut balance = storage.balances.get(to);
+            storage.balances.insert(to, balance + 1);
```
Braqzen, line 316:
```suggestion
            storage.balances.insert(to, storage.balances.get(to) + 1);
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            !storage.access_control || 
+            (storage.access_control && storage.allowed_minters.get(sender)), 
+            AccessError::SenderDoesNotHaveAccessControl
+        );
+
+        /// Mint as many tokens as the sender has paid for
+        let mut i = 0;
+        while i < amount {
+            /// Create a new token id in sequential order
+            let token_id: u64 = storage.token_count + 1;
```
Braqzen, line 304:
```suggestion
            storage.token_count = storage.token_count + 1;
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            let empty_identity: Identity = Identity::Address(~Address::from(NATIVE_ASSET_ID));
+
+            /// Create the metadata for this new token with the owner 
+            let meta_data: MetaData = MetaData {
+                owner: to, approved: empty_identity
+            };
+            storage.meta_data.insert(token_id, meta_data);
+            storage.owners.insert(to, token_id);
```
Braqzen, line 312:
```suggestion
            storage.meta_data.insert(storage.token_count, meta_data);
            storage.owners.insert(to, storage.token_count);
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            
+            /// Increase the balance of the new owner
+            let mut balance = storage.balances.get(to);
+            storage.balances.insert(to, balance + 1);
+
+            /// Increment the token count and the number of tokens minted in this transaction
+            storage.token_count = storage.token_count + 1;
```
Braqzen, line 319:
```suggestion
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+        /// Ensure that the sender is on the approved mint list
+        require(
+            !storage.access_control || 
+            (storage.access_control && storage.allowed_minters.get(sender)), 
+            AccessError::SenderDoesNotHaveAccessControl
+        );
+
+        /// Mint as many tokens as the sender has paid for
+        let mut i = 0;
+        while i < amount {
```
Braqzen, line 302:
```suggestion
        let mut count = 0;
        while count < amount {
```

could use `index` too

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            storage.meta_data.insert(token_id, meta_data);
+            storage.owners.insert(to, token_id);
+            
+            /// Increase the balance of the new owner
+            let mut balance = storage.balances.get(to);
+            storage.balances.insert(to, balance + 1);
+
+            /// Increment the token count and the number of tokens minted in this transaction
+            storage.token_count = storage.token_count + 1;
+            i = i + 1;
```
Braqzen, line 320:
```suggestion
            count = count + 1;
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+
+            log(MintEvent{owner: to, token_id});
```
Braqzen, line 322: Perhaps this should have some logic to keep track of how many are minted and only emit a log once at the end. Logging multiple times is not a good idea.

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    // Uncomment when https://github.com/FuelLabs/fuels-rs/issues/375 is resolved
+    /// Returns the owner of a given token id
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    // fn owner_of(token_id: u64) -> Option<Identity> {
```
Braqzen, line 335: similar comments to get_approved

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    /// - The operator address provided not a valid identity
+    /// - The address has already been approved
+    /// - The sender is not the owner
+    fn set_approval_for_all(owner: Identity, operator: Identity) -> bool {
+        require(storage.state != 0, InitError::NFTNotInitalized);
+        require_identity_is_valid(operator);
+
+        let hash: b256 = get_owner_operator_hash(owner, operator);
```
Braqzen, line 356:
```suggestion
        let hash = sha256(owner, operator);
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    // }
+
+    /// Gives the operator identity approval to transfer any tokens owned by 
+    /// the owner identity. This can be dangerous.
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    /// - The operator address provided not a valid identity
```
Braqzen, line 349:
```suggestion
    /// - The operator address is not a valid identity
```

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /// Returns the balance of the specified owner
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    fn balance_of(owner: Identity) -> u64 {
+        require(storage.state != 0, InitError::NFTNotInitalized);
```
Braqzen, line 152: remove require

Braqzen (changes requested):
I'll make further comments on the remaining functions once these changes are made

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+            require(entity != ~Address::from(NATIVE_ASSET_ID), 
+                InputError::InputAddressCannotBeZero
+            );
+        },
+        Identity::ContractId(entity) => {
+            require(entity != ~ContractId::from(NATIVE_ASSET_ID), 
+                InputError::InputAddressCannotBeZero
+            );
+        },
+    };
```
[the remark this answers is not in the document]
bitzoic: I like this a lot better. Since you don't have the type specified and know both have a value, just declare a variable

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+
+    // Uncomment when https://github.com/FuelLabs/fuels-rs/issues/375 is resolved
+    /// Returns the approved address
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been initalized
+    // fn get_approved(token_id: u64) -> Option<Identity> {
+    //     require(storage.state != 0, InitError::NFTNotInitalized);
```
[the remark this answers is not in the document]
bitzoic: If a token is burned, doesn't exist, or is not owned by anyone the 0 address returns. I will change this to return `None`

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+    // }
+
+    /// Returns the tokens owned by the address
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The NFT contract has not been intialized
+    fn get_tokens(identity: Identity) -> u64 {
+        require(storage.state != 0, InitError::NFTNotInitalized);
```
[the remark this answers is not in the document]
bitzoic: I would rather have no one be able to interact with the contract at all until the contract has been initialized.

nft/src/main.sw
bitzoic: I guess I'm too nice lol

nft/src/main.sw
```diff
@@ -0,0 +1,474 @@
...
+        require_identity_is_valid(to);
+
+        let sender: Identity = unwrap_identity(msg_sender());
+
+        /// Ensure that the sender is on the approved mint list
+        require(
+            !storage.access_control || 
+            (storage.access_control && storage.allowed_minters.get(sender)), 
+            AccessError::SenderDoesNotHaveAccessControl
+        );
```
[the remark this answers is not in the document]
bitzoic: This is better

nft/src/main.sw
bitzoic: Not sure what I would use in place of the 0 address. As far as I'm aware Sway or Rust doesn't have a `null` in the way that C++ does. In solidity the 0 address is reserved. If a contract were to receive the 0 address that would let them do a lot of damage.

nft/src/main.sw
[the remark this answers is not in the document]
Braqzen: A token can be owned... by no one?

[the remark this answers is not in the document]
bitzoic: When I'm referring to not owned by anyone I mean not yet minted sorry

nft/src/main.sw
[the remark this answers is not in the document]
bitzoic: Good point

nft/src/main.sw
```diff
@@ -10,49 +10,56 @@ use abi::NFT;
...
     result::*,
     revert::revert,
     storage::StorageMap,
+    vec::Vec,
 };
 
 storage {
-    /// Determines if there is a whitelist to mint. Only set on the
-    /// initalization of the contract
+    /// Determines if there is an allow list to call the mint function
```
nfurfaro, line 30: use whitelist **or** allow list, consistently.

nft/src/main.sw
```diff
@@ -10,49 +10,56 @@ use abi::NFT;
...
     operator_approval: StorageMap<b256, bool>,
 
     /// TODO: Use a Vec here to support multiple ownerships
     /// Maintains the token ids owned by the specified identity
+    /// Map(Identity => owned tokens)
     owners: StorageMap<Identity, u64>,
 
     /// The total number of tokens that have been minted
+    /// This should only be incremented
     token_count: u64,
```
nfurfaro, line 63: It may be more useful to make this `total_suppply` and also decrement it if a token is burned?

nft/src/main.sw
```diff
@@ -10,49 +10,56 @@ use abi::NFT;
...
-    /// Determines if there is a whitelist to mint. Only set on the
-    /// initalization of the contract
+    /// Determines if there is an allow list to call the mint function
+    /// Only set on the initalization of the contract
     access_control: bool,
 
-    /// The identity that has permission to add identities to the whitelist
+    /// Stores the identity that has permission to add identities to the whitelist
+    /// Only set on the initalization of the contract
     access_control_address: Identity,
```
nfurfaro, line 36: maybe just `admin: Identity` ?
In the future, we may want to expand the scope of what this Identity can do beyond managing access control.

nft/src/main.sw
Braqzen: Whitelist is a known and established term. I was waiting for this to be marked as a PR before mentioning that

nft/src/main.sw
```diff
@@ -62,84 +69,94 @@ storage {
...
         storage.allowed_minters.insert(minter, allow);
     }
 
-    /// Gives approval to the 'to' Identity to transfer the specified token
+    /// Gives approval to the 'to' Identity to transfer the specified token on the owner's behalf.
+    ///
+    /// # Arguments
     ///
-    /// # Panics
+    /// * `to` - The `Identity` which will be allowed to transfer the token.
```
nfurfaro, line 104: perhaps use `spender` or `approved` rather than `to` for clearer semantics.

nft/src/main.sw
```diff
@@ -10,49 +10,56 @@ use abi::NFT;
...
     meta_data: StorageMap<u64, Option<MetaData>>,
 
-    /// Stores a b256 hash of the (owner, operator) and stores whether the
+    /// Maps a b256 hash of the (owner, operator) identities and stores whether the
     /// operator is allowed to transfer ALL tokens on the owner's behalf 
+    /// Map(hash => approved)
     operator_approval: StorageMap<b256, bool>,
 
     /// TODO: Use a Vec here to support multiple ownerships
     /// Maintains the token ids owned by the specified identity
```
nfurfaro, line 57: This looks like it currently maps an `Identity` to the number of tokens it owns, rather than to the token_ids it owns. I'm guessing StorageVec is needed here.

bitzoic: Right now since we do not have `StorageVec` it can only keep track of a single token, but it does store which token_ids are owned. In the case of transfer it currently just sets to 0 which is problematic if the `Identity` owns multiple tokens

nft/src/main.sw
bitzoic: Do you think it would be useful to keep track of the original total supply? If someone in the future were to check whether a token_id is less than the total_supply, burning a token could throw this off hence why I've only left it to be set in the constructor.

nft/src/abi.sw
```diff
@@ -0,0 +1,32 @@
+library abi;
+
+use std::{identity::Identity, option::Option};
+
+abi NFT {
+    #[storage(read, write)]
+    fn allow_mint(minter: Identity, allow: bool);
```
nfurfaro, line 7: I feel that this function is out of place in the abi, and would prefer a simplified (but still flexible) approach.
e.g: Set a single `minter: Identity` in the constructor.

This minter can be an EOA for simple scenarios, or can be a dedicated `Minter` contract which can modulate access (using its own access control scheme to add/remove Identities which can mint NFT _through_ the `Minter` contract.
This would allow for arbitrarily complex access control schemes around minting, while keeping the base NFT contract clean and minimal.

nfurfaro: The dedicated `Minter` contract could also be upgraded while retaining the same ContractId, allowing yet another path for making changes without complicating the base NFT.

bitzoic: I like this approach a lot. I will likely create an example `Minter` contract in the near future

bitzoic:
In order to merge this in, I have removed the need for a `StorageVec`. This was originally used to return a `Vec` of all tokens owned by an `Identity`. This can be done instead by calling the `total_supply()` function and then checking the tokens with `owner_of()` while the tokens found is less than the `balance_of()` the `Identity`. This also reduces the contract storage.

NFT/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "NFT"
+version = "0.0.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = "0.16"
+fuels-abigen-macro = "0.16"
```
adlerjohn, line 10: I was under the impression `fuels-abigen-macro` was no longer needed as a separate import as of `0.16`: https://github.com/FuelLabs/fuels-rs/pull/393

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+    /// This should only be incremented.
+    token_count: u64,
+
+    /// The total supply tokens that can be minted. Can only be set on the initalization of the
+    /// contract.
+    token_supply: u64,
+}
+
+impl NFT for Contract {
+    /// Constructor for the NFT. Calling this function will instantiate the `total_supply`, the `admin`
```
adlerjohn, line 56: Did we decide on putting docstrings on implementations? The issue with this is that a downstream contract of this contract only has access to the ABI file, not the implementation (of which there may be many). They wouldn't actually be able to see these docstrings.

As such, documentation should go on the ABI declaration.

NFT/Cargo.toml
Braqzen: Correct

NFT/src/main.sw
Braqzen: Interesting. I was only considering the implementation since that is what you are reading and it makes it easier to match up the docs with the code since you can see it right there.

At the same time I dislike the idea of not having docs on the functions themselves. It feels like it might become disjoint if not carefully maintained. It's also not a good idea to duplicate in the abi and impl.

I will make adjustments so that the docs are on the ABI instead when I get around to upgrading the contracts in some issues - some of which are already as drafts.

README.md
```diff
@@ -51,6 +51,7 @@ sway-applications/
 - [Escrow](./escrow) is a third party that keeps an asset on behalf of multiple parties
 - [Fundraiser](./fundraiser/) is a program allowing users to pledge towards a goal
 - [Multi-Signature Wallet](./multisig-wallet) is a wallet that requires multiple signatures to execute a transaction
+- [NFT](./NFT) is a general purpose Non-Fungible Token contract
```
matt-user, line 54: This suggestion may be unnecessary @Braqzen what do you think?
```suggestion
- [Non-Fungible Token (NFT)](./NFT) is a general purpose Non-Fungible Token contract
```

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+    /// Determines if only the `admin` is allowed to call the mint function.
+    /// This is only set on the initalization of the contract.
+    access_control: bool,
+
+    /// Stores an `Option` of the `Identity` that has permission to mint if `access_control` is set to true.
+    /// Will store `None` if this contract does not have `access_control` set.
+    /// Only the `admin` is allowed to change the `admin` of the contract.
+    admin: Option<Identity>,
+
+    /// Used of O(1) lookup of the number of tokens owned by each `Identity`.
```
matt-user, line 35:
```suggestion
    /// Used for O(1) lookup of the number of NFT tokens owned by each `Identity`.
```

bitzoic: I've been primarily referring to NFT as the contract as a whole and tokens as the individual pieces of ownership within the contract.

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+    /// * `to` - The `Identity` which will own the minted tokens.
+    /// * `amount` - The number of tokens to be minted in this transaction.
+    ///
+    /// # Reverts
+    ///
+    /// * When the sender attempts to mint more tokens than total supply.
+    /// * When the sender is not the admin and `access_control` is set.
+    #[storage(read, write)]fn mint(to: Identity, amount: u64) {
+        // The current number of tokens minted plus the amount to be minted cannot be
+        // greater than the total supply
```
matt-user, line 100: I would add some words here describing how the mint will also fail if the contract has not been initialized because the total supply would be 0.

NFT/tests/harness.rs
```diff
@@ -0,0 +1,633 @@
...
+            let minter = Identity::Address(owner2.wallet.address());
+            let admin = Option::Some(Identity::Address(owner1.wallet.address()));
+            constructor(&deploy_wallet.nft, &admin, true, 1).await;
+
+            mint(&owner2.nft, &minter, 1).await;
+        }
+    }
+}
+
+mod set_admin {
```
matt-user, line 183: In general, I would match the order of your test cases to the order they are defined in your abi and contract.  I would also order your revert test cases the order in which they appear in your contract (not saying these are out of order just mentioning it)

matt-user (changes requested):
Hopefully @Braqzen will chime in if any of these suggestions are misguided.

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+        require(storage.token_supply >= (storage.token_count + amount), InputError::NotEnoughTokensToMint);
+
+        // Ensure that the sender is on the admin if this is a controlled access mint
+        require(!storage.access_control || (storage.admin.is_some() && msg_sender().unwrap() == storage.admin.unwrap()), AccessError::SenderDoesNotHaveAccessControl);
+
+        // Mint as many tokens as the sender has asked for
+        let mut index = 0;
+        let mut minted_tokens: Vec<u64> = ~Vec::new::<u64>();
```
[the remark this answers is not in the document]
bitzoic: No not anymore, this is left over from an earlier version where it was required.

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+    /// This increments or decrements when minting, transfering ownership, and burning tokens.
+    /// Map(Identity => balance)
+    balances: StorageMap<Identity,
+    u64>, /// Stores the `Metadata` for each token based on the token's `u64` id
+    /// Map(token_id => Metadata)
+    meta_data: StorageMap<u64,
+    Option<MetaData>>, /// Maps a `b256` hash of the (owner, operator) identities and stores whether the
+    /// operator is allowed to transfer ALL tokens on the owner's behalf.
+    /// Map(hash => approved)
+    operator_approval: StorageMap<b256,
```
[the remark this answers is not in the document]
bitzoic: I guess this is a habit from Solidity. In terms of efficiency, I'm not sure there are any benefits. I believe any tuples are hashed by default but I'll let someone else with more knowhow chime in here. If this is the case I believe the tuple will be more readable.

NFT/src/main.sw
```diff
@@ -0,0 +1,368 @@
...
+    ///
+    /// # Reverts
+    ///
+    /// * When `token_id` does not map to an existing token.
+    /// * When the 'approved' `Identity` is the token's owner.
+    /// * When the sender is not the token's owner.
+    #[storage(read, write)]fn approve(approved: Option<Identity>, token_id: u64) {
+        // Ensure this is a valid token
+        let meta_data = storage.meta_data.get(token_id);
+        require(meta_data.is_some(), InputError::TokenDoesNotExist);
```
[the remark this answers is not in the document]
bitzoic: Utility functions don't inherently have access to storage. I could add something like this however:

```
pub fn token_metadata(meta_data: Option<MetaData>) -> MetaData{
    require(meta_data.is_some(), InputError::TokenDoesNotExist);
    meta_data.unwrap()
}
```

Where calling it would look like `let mut meta_data = token_metadata(storage.meta_data.get(token_id));`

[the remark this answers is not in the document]
matt-user: I personally think this would be preferable, but feel free to leave it as it is.

NFT/src/main.sw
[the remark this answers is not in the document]
matt-user: Yes I agree the tuple is more readable.  However, I don't know whether there is an efficiency difference either lol.

NFT/src/main.sw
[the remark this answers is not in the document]
bitzoic: Just tried it out and it lowers the contract size by a few bytes so I'll take it!

NFT/src/main.sw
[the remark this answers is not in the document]
bitzoic: Again lowers contract byte size woop woop!

bitzoic:
Blocked by https://github.com/FuelLabs/sway/issues/2232

NFT/src/main.sw
nfurfaro: In general, I agree that it makes more sense to add docstrings to the ABI declaration (even though I like to prefer reading it right on the implementation in practice), and also that we don't want duplicated docs for maintenance reasons. 
On the other hand, the implementation should add additional documentation regarding details specific to the implementation ( i.e: additional cases which may cause a revert).

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// Determines if only the `admin` is allowed to call the mint function.
+    /// This is only set on the initalization of the contract.
+    access_control: bool,
+
+    /// Stores an `Option` of the `Identity` that has permission to mint if `access_control` is set to true.
+    /// Will store `None` if this contract does not have `access_control` set.
+    /// Only the `admin` is allowed to change the `admin` of the contract.
+    admin: Option<Identity>,
+
+    /// Used of O(1) lookup of the number of tokens owned by each `Identity`.
```
nfurfaro, line 36: maybe "Use of ..." ?

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// operator is allowed to transfer ALL tokens on the owner's behalf.
+    /// Map((owner, operator) => approved)
+    operator_approval: StorageMap<(Identity,
+    Identity), bool>, /// The total number of tokens that have been minted.
+    /// This should only be incremented.
+    token_count: u64,
+
+    /// The total supply tokens that can be minted. Can only be set on the initalization of the
+    /// contract.
+    token_supply: u64,
```
nfurfaro, line 53: I feel there's something lacking here. 
First, I don't think ? there's any storage variable tracking the _total current_ tokens in existence?
I might have missed something as it very early, but I might suggest something like:
```
supply_cap: u64; // The total supply tokens that can be minted.

total_supply: u64 // The number of tokens currently in existence, Incremented on mint, decremented on burn.

tokens_ever_minted: u64; // The number of tokens that have been minted.
```

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// * When the `token_supply` is set to 0.
+    /// * When `access_control` is set to true and no admin was given.
+    #[storage(read, write)]fn constructor(admin: Option<Identity>, access_control: bool, token_supply: u64) {
+        // This function can only be called once so if the token supply is already set it has
+        // already been called
+        require(storage.token_supply == 0, InitError::CannotReinitialize);
+        // The number of tokens that can be minted cannot be 0
+        require(token_supply != 0, InputError::TokenSupplyCannotBeZero);
+        // Access control is set to true but there was no admin given
+        require((access_control && admin.is_some()) || admin.is_none(), InitError::AccessControlSetAndAdminIsNone);
```
nfurfaro, line 79: Have you tested this with a case where `access_control` is true, but `admin.is_none()`?
It looks like the first '&&' condition will fail, but then the second would pass.

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// # Arguments
+    ///
+    /// * `to` - The `Identity` which will own the minted tokens.
+    /// * `amount` - The `u64` number of tokens to be minted in this transaction.
+    ///
+    /// # Reverts
+    ///
+    /// * When the sender attempts to mint more tokens than total supply.
+    /// * When the sender is not the admin and `access_control` is set.
+    #[storage(read, write)]fn mint(to: Identity, amount: u64) {
```
nfurfaro, line 101: Maybe not 100% necessary in the initial version here, but an improvement would be to have both a `mint` and a `mint_batch` function.
This would help minimize the minting cost for those who only want to mint a single token.
Eliminating the loop for single mints may be a small gas saving, but these things have a way of adding up.

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+        while index < amount {
+            // Increment the token count
+            storage.token_count = storage.token_count + 1;
+
+            // Create the metadata for this new token with the owner
+            let meta_data = ~MetaData::new(Option::None(), to);
+            storage.meta_data.insert(storage.token_count, Option::Some(meta_data));
+
+            // Increase the balance of the new owner
+            storage.balances.insert(to, storage.balances.get(to) + 1);
```
nfurfaro, line 121: I would take this out of the loop so you're not writing to storage more than once.
Just do `storage.balances.insert(to, storage.balances.get(to) + amount);`

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+
+            // Push to minted tokens Vec
+            minted_tokens.push(storage.token_count);
```
nfurfaro, line 124: _maybe_ same as above? 
In general, try to minimize storage read/writes where possible.

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// to `None`. After the token has been burned, no one will be able to fetch any data
+    /// about this token or have control over it.
+    ///
+    /// # Arguments
+    ///
+    /// * `token_id` - The `u64` ID of the token which is to be burned.
+    ///
+    /// * Reverts
+    ///
+    /// * When `token_id` does map to an existing token.
```
nfurfaro, line 145: "... does NOT map ..."

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    #[storage(read, write)]fn burn(token_id: u64) {
+        // Ensure this is a valid token that has already been minted and exists
+        let mut meta_data = token_metadata(storage.meta_data.get(token_id));
+
+        // Ensure the sender owns the token that is provided
+        let owner = msg_sender().unwrap();
+        require(meta_data.owner == owner, AccessError::SenderNotOwner);
+
+        // Burn this token by setting the `token_id` Metadata mapping to `None`
+        storage.meta_data.insert(token_id, Option::None());
```
nfurfaro, line 156: I'm not sure here about the gas cost implications of setting the metadata to `None`...
Are we effectively clearing (overwriting with 0's) all slots used by a Metadata struct (up to 4 slots here, 2 for each `Identity`)? 
If so, then we should use a different approach for registering a token as burned (i.e: keep `owner` separate so that just the owner can be set to `None`.
cc @mohammadfawaz

mohammadfawaz: @nfurfaro is right. We're not effectively overwriting with zeros thought because `None` is not the 0th enum variant, but we still write `size_of(Option<MetaData>)` bytes which you can divide by 32 and round up to figure out how many storage slots it requires.

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+        log(TransferEvent {
+            from, sender, to, token_id
+        });
+    }
+
+    /// Gives approval to the 'approved' `Identity` to transfer the specified token on the owner's behalf.
+    ///
+    /// # Arguments
+    ///
+    /// * `approved` - The `Identity` which will be allowed to transfer the token.
```
Braqzen, line 218: 3 args, 2 params

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// # Reverts
+    ///
+    /// * When `token_id` does not map to an existing token.
+    /// * When the 'approved' `Identity` is the token's owner.
+    /// * When the sender is not the token's owner.
+    #[storage(read, write)]fn approve(approved: Option<Identity>, token_id: u64) {
+        // Ensure this is a valid token
+        let mut meta_data = token_metadata(storage.meta_data.get(token_id));
+
+        // The owner cannot approve themselves
```
Braqzen, line 231: Why can't they?

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+        storage.meta_data.insert(token_id, Option::Some(meta_data));
+
+        // Log the approval event
+        log(ApprovalEvent {
+            owner: sender, approved, token_id
+        });
+    }
+
+    /// Gives the `operator` `Identity` approval to transfer ALL tokens owned by
+    /// the `owner` `Identity`. This can be dangerous.
```
Braqzen, line 249:
```suggestion
    /// Gives the `operator` approval to transfer ALL tokens owned by the `owner`
```

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `owner` - The `Identity` which owns tokens.
```
Braqzen, line 253: 2 args, 3 out of order params

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    /// * `operator` - The `Identity` which may transfer all tokens owned by the `owner`.
+    ///
+    /// # Reverts
+    ///
+    /// * When the sender is not the `owner`.
+    #[storage(read, write)]fn set_approval_for_all(owner: Identity, operator: Identity, allow: bool) {
```
Braqzen, line 259: why is the owner param even needed? just use msg_sender()

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+        // Only the owner is allowed to set an operator for themselves
+        require(owner == msg_sender().unwrap(), AccessError::SenderNotOwner);
+
+        // Set the identity to have or not have approval to transfer all tokens owned
```
Braqzen, line 263: Comment needs work or just remove it

NFT/src/main.sw
bitzoic: Yes there is a test for this in the harness.rs

NFT/src/main.sw
bitzoic: I think this would be good as a new issue.

NFT/src/main.sw
bitzoic: I like this much better

Braqzen (changes requested):
More comments on subsequent passes, including testing.

NFT/src/main.sw
bitzoic: I've moved updating `storage.balances` and `storage.token_count` outside of the while loop but I don't think it's possible to move inserting `meta_data` to outside of the while loop.

NFT/src/errors.sw
```diff
@@ -0,0 +1,28 @@
...
+}
+
+pub enum InitError {
+    AccessControlSetAndAdminIsNone: (),
+    CannotReinitialize: (),
+    NFTNotInitalized: (),
+}
+
+pub enum InputError {
+    InputAddressCannotBeZero: (),
```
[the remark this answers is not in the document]
bitzoic: Also removed

NFT/src/events.sw
```diff
@@ -0,0 +1,56 @@
+library events;
```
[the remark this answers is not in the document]
Braqzen: Nice. That means I have to update that in the contibuting.md that recently went in

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+    Identity), bool>, /// The total number of tokens that have been minted.
+    /// This should only be incremented.
+    token_count: u64,
+
+    /// The total supply tokens that can be minted. Can only be set on the initalization of the
+    /// contract.
+    token_supply: u64,
+}
+
+impl NFT for Contract {
```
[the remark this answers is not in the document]
bitzoic: In previous reviews it's been discussed that implementations should be in ordered what is needed most and logical. For example, in the Fundraiser any getters are at the end of the implementation. Are we going alphaetizie EVERYTHING direction?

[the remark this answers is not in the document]
Braqzen: Yeah, I've written down my reasoning in the contrib doc iirc. It's easier to be consistent than to leave some room for uncertainty "it applies here but maybe not there". Unless there's a specific reason not to then alphabetically order everything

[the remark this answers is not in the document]
bitzoic: Sweet makes everything nice and easy.

NFT/src/main.sw
```diff
@@ -0,0 +1,359 @@
...
+
+            // Push to minted tokens Vec
+            minted_tokens.push(storage.token_count);
+
+            // Increment the number of tokens minted in this transaction
+            index = index + 1;
+        }
+
+        log(MintEvent {
+            owner: to, token_ids: minted_tokens
```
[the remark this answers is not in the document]
bitzoic: Not sure what you mean by referencing a range. Two `u64` stating the start and the end?

NFT/src/main.sw
nfurfaro: I did a quick test for this, which indicates that we're writing 3 slots for TokenMetadata:
```
script;

use std::{
    address::Address,
    identity::Identity,
    intrinsics::size_of_val,
    logging::log,
    option::Option,
};

pub struct TokenMetaData {
    /// The user that is allowed to transfer this token.
    approved: Option<Identity>,
    /// The user that owns this token.
    owner: Identity,
}

fn main() -> u64 {

    let address_1 = ~Address::from(0x1111111111111111111111111111111111111111111111111111111111111111);
    let approved_addr = Identity::Address(address_1);
    let approved_opt = Option::Some(approved_addr);
    let address_2 = ~Address::from(0x2222222222222222222222222222222222222222222222222222222222222222);
    let owner = Identity::Address(address_2);

    let metadata = TokenMetaData {
        approved: approved_opt,
        owner: owner,
    };

    let size = size_of_val(metadata);
    let div = size / 32;
    let mod = size % 32;

    let total_slots = if mod > 0 {
        div + 1
    } else {
        div
    };

    total_slots
}
```

nfurfaro: I suggest breaking up the `TokenMetadata` into 2 separate storage variables which can be updated independently of each other, as the organizational benefit of grouping them is eclipsed by the extra gas cost while managing them.

Braqzen: What is the original gas cost vs your proposed gas cost i.e. the difference in cost?

nfurfaro: I can't speak to the actual cost, but we would eliminate reading/writing 1 slot with the proposed change (so ~33% savings), **each time** we need to read or write **either** of the 2 TokenMetaData fields.

(The change proposed [here](https://github.com/FuelLabs/fuel-specs/issues/324#issuecomment-1146847164)
would eliminate another storage slot, but is orthogonal to this PR.)

bitzoic: I've personally never been a fan of the way Solidity NFTs use a URI and store data off-chain, but I understand the reasoning. I'd like to limit that as much as possible. As for these two methods of storing the URI, I defintely prefer the URI-shortener to have a uniform length. I will have to dig deeper into this tomorrow or any other possible solutions.

adlerjohn: Can we defer this to a future PR? Interesting discussion, but this level of gas golfing isn't a req for functionality.

NFT/tests/harness.rs
```diff
@@ -0,0 +1,739 @@
...
+            assert_eq!(balance_of(&owner1.nft, &minter).await, 0);
+
+            mint(10, &owner1.nft, &minter).await;
+
+            assert_eq!(balance_of(&owner1.nft, &minter).await, 10);
+            assert_eq!(total_supply(&owner1.nft).await, 10);
+            assert_eq!(max_supply(&owner1.nft).await, 10);
+            for itterator in 0..10 {
+                assert_eq!(
+                    owner_of(&owner1.nft, itterator).await,
```
Braqzen, line 401:
```suggestion
            for iterator in 0..10 {
                assert_eq!(
                    owner_of(&owner1.nft, iterator).await,
```

is this a typo or intentional because `iterator` is a keyword?

Braqzen (changes requested):
1. Aside from this being possibly moved into [sway-libs](https://github.com/FuelLabs/sway-libs)... could you update the logo so that it automatically changes based on the user's theme i.e. light/dark? You can see the example [here](https://raw.githubusercontent.com/FuelLabs/sway-applications/master/fundraiser/README.md)
2. I don't remember how much it matters but could you bump the SDK to use the latest version? I think it is v0.17 atm
3. `data_structures.sw` has an empty struct for meta data. Could you add some fields into it like a storage vec in place of a string and whatever other structures a token usually contains? (storage vec is almost in master of sway)
4. Storage contains `max_supply`, `total_supply` & `tokens_ever_minted`. Without reading the rest of the contract I do not understand why that last one exists. I assume that you care about the max supply and the current supply but I don't understand why it matters how many tokens have been minted overall. Could you elaborate?
6. The function in `utils.sw` is never called in `main.sw`

Looks pretty good. Some minor comments here and there. I think after this we'll need an expert to look at the token for a final confirmation cc @nfurfaro

NFT/tests/harness.rs
```diff
@@ -0,0 +1,739 @@
...
+                owner_of(&owner1.nft, 0).await,
+                Option::Some(operator.clone())
+            );
+
+            assert_eq!(balance_of(&owner1.nft, &minter).await, 0);
+            assert_eq!(balance_of(&owner2.nft, &operator).await, 1);
+        }
+    }
+
+    mod reverts {
```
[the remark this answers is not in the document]
bitzoic: There wouldn't be anything different in such a test as it uses a single require statement for both cases. If the sender is not the owner or approved the test passes.

Renaming it `panics_when_sender_is_not_owner_or_approved`

nfurfaro:
>Storage contains max_supply, total_supply & tokens_ever_minted. Without reading the rest of the contract I do not understand why that last one exists. I assume that you care about the max supply and the current supply but I don't understand why it matters how many tokens have been minted overall. Could you elaborate?

I would say that `tokens_ever_minted` is not strictly necessary for all use-cases.
It does enable the case where:
- we have a hard supply cap set by `max_supply` (as in there can only ever be 1000 tokens of this type minted)
- we allow tokens to be burned (which reduces `total_supply`, so we can't enforce the supply cap by using `total_supply` as it tracks the number of tokens currently in existence)

For the base token, we could probably remove `tokens_ever_minted` and let users add it if they really need it, and keep things simple here.

NFT/src/interface.sw
```diff
@@ -0,0 +1,229 @@
...
+pub struct MintEvent {
+    /// The owner of the newly minted tokens.
+    owner: Identity,
+
+    /// The starting range of token ids that have been minted in this transaction.
+    token_id_start: u64,
+
+    /// The total number of tokens minted in this transaction.
+    total_tokens: u64,
+}
```
matt-user, line 43: Is there a reason you moved the events into the contract interface?  I don't think this is a bad thing necessarily, but I think we should be consistent about how we structure our applications.  The reason being is it will be easier for new developers to find code they are looking for if our applications are structured similarly.

NFT/src/main.sw
```diff
@@ -1,368 +1,242 @@
...
-        require(meta_data.owner == sender, AccessError::SenderNotOwner);
-
-        // Burn this token by setting the `token_id` Metadata mapping to `None`
-        storage.meta_data.insert(token_id, Option::None());
+        require(token_owner.unwrap() == sender, AccessError::SenderNotOwner);
 
-        // Reduce the balance of tokens for the owner
+        storage.owners.insert(token_id, Option::None());
         storage.balances.insert(sender, storage.balances.get(sender) - 1);
+        storage.total_supply = storage.total_supply - 1;
```
matt-user, line 114:
```suggestion
        storage.total_supply -= 1;
```

matt-user (changes requested):
There is a lot of commented out code in your test harness.  I would either delete the commented code or write a comment somewhere at the top of your harness.rs explaining why the code is commented

NFT/src/interface.sw
bitzoic: Since a lot of other applications will be importing the NFT interface they may need access to events hence why they are moved here. I will be adding an NFT interface to the sway-libs repository and then refactoring this application to use it once it is implemented which should clean this up.

bitzoic:
> There is a lot of commented out code in your test harness. I would either delete the commented code or write a comment somewhere at the top of your harness.rs explaining why the code is commented

I've added a comment stating that I am awaiting SDK features

matt-user (approved):
Looks good to me!

Braqzen (changes requested):
1. The `data_structures.sw` contains an `impl` for the `TokenMetaData` which uses the name of the struct rather than `Self` which means if the struct is renamed then there's one more change to be made while using `Self` means that you do not have to worry about the name of the struct. Here is an example.

**Before**

```rust
impl TokenMetaData {
    fn new() -> Self {
        TokenMetaData {
            name: "Example",
        }
    }
}
```

**After**

```rust
impl TokenMetaData {
    fn new() -> Self {
        Self {
            name: "Example",
        }
    }
}
```

2. `main.sw` imports `std::chain::auth::AuthError` but does not seem to use it
3. There are some places where the syntax `var = var + 1` is used rather than `var += 1`
4. The documentation on the `abi` for `approved` & `owner_of` is missing the `revert` case
5. Minor nit, in the `harness.rs` could you change the variables that are unused to be `_` instead of e.g. `_deploy_wallet`?
6. Could you also adjust the "TODO" comment there and change the commented out tests to not be commented out and instead add a `#[ignore]` attribute? This will ignore the test so it will still show up when the tests are run but it will not run them. This makes it easier to notice that there are tests that must be addressed at some point rather than having to dig into the code itself to find commented out tests
7. I would personally change the `Metadata` field `nft` to be `contract` since, imo, it's clearer that a contract is being passed in / interacted with as opposed to `nft` or `handle` etc.
8. It's not clear to me what the `NftBuilder` is and where it comes from. Can you elucidate?
9. The `approve` module for the `revert` / `panics_when_not_initialized` test is a bit misleading imo. That test panics because the token owner is not set rather than something not being initialized. Could you rename the test to be more representative of the panic case? Actually, looking at the next test that covers it so I would remove this test
    - At least the renaming applies to the `revert` tests for
        - `burn`
        - `mint`
        - `set_admin`
11. For any success case that deals with only 1 NFT could you add another test so that the tests prove that the contract works for more than 1 NFT?
12. Can you change any `assert_eq!` macros that use `{` / `}` to `(` / `)`?
13. In `is_approved_for_all` `success` tests could you add a pre-condition to prove that the `owner` / `operator` combo was first false and then your test changed it to true?
    - This kind of tests the `set_approval_for_all` but it's important to check both the positive and negative test cases for the `is_...` call
    - Could you also add this pre-condition to the `set_appr...` `success` test cases?
14. Should a user be able to mint 0?
15. Do you think that there should be another function to get the current admin?
16. I believe that the `revert` tests for `transfer_from` only test the first `sender == token_owner` case in the second `require` since when that fails the other checks are no longer performed. Could you add the remaining test cases?

bitzoic:
> 8. It's not clear to me what the `NftBuilder` is and where it comes from. Can you elucidate?

NFTBuilder is part of the updated SDK. There are examples [here](https://github.com/FuelLabs/fuels-rs/blob/master/examples/contracts/src/lib.rs). It threw me off at first as well and took an embarrassingly long time to figure out.

> 13. Should a user be able to mint 0?

Yes, they can waste their gas.

> 15. I believe that the `revert` tests for `transfer_from` only test the first `sender == token_owner` case in the second `require` since when that fails the other checks are no longer performed. Could you add the remaining test cases?

If `sender == token_owner` fails, then `(approved.is_some() && sender == approved.unwrap())` is checked. If that fails then `(from == token_owner && storage.operator_approval.get((from, sender)))` is checked because this is just a large or case. All 3 are tested in the same case and cannot be tested in separate cases.

Braqzen (dismissed):
Looks reasonable to me. I would wait on 1 or 2 more reviews 1 of which being from @nfurfaro

NFT/src/interface.sw
```diff
@@ -0,0 +1,244 @@
...
+
+    /// Sets the inital state and unlocks the functionality for the rest of the contract.
+    ///
+    /// This function can only be called once.
+    ///
+    /// # Arguments
+    ///
+    /// * `access_control` - Determines whether only the admin can call the mint function.
+    /// * `admin` - The user which has the ability to mint if `access_control` is set to true and change the contract's admin.
+    /// * `token_supply` - The maximum supply of tokens that can ever be minted.
```
nfurfaro, line 112: I would rename this param to `max_supply` for consistency.

NFT/src/main.sw
```diff
@@ -0,0 +1,250 @@
...
+    /// Map(token_id => owner)
+    owners: StorageMap<u64,
+    Option<Identity>> = StorageMap {
+    }, /// The total number of tokens that ever have been minted.
+    /// This is used to assign token identifiers when minting. This will only be incremented.
+    tokens_minted: u64 = 0,
+
+    /// The number of tokens currently in existence.
+    /// This is incremented on mint and decremented on burn. This should not be used to assign
+    /// unqiue identifiers due to the decementation of the value on burning of tokens.
```
nfurfaro, line 65: typo: "decrementation"

nfurfaro (commented):
Looks good! 2 small nits, but otherwise ready to approve on my end.

nfurfaro (approved):
Looks Good!

Braqzen (approved):

PR #50, opened 2022-05-28
Added CONTRIBUTING.md

### LINKED ISSUE #44: Create a CONTRIBUTING.md
The README.md contains some info about contributing however that is the incorrect place to store such information.
Moreover, it is not particularly descriptive.
Expand on the contributions in this file.

Braqzen:
Closes #44


CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+
+If you are looking to do more than just raise an issue then check to see if there are any issues labelled under `good first issue` and no one is assigned to them.
+
+These issues should be relatively easy to do and they are there specifically to let others jump in and have a go.
+
+Before you go off and make a pull request (PR) make sure to first fork the repository and work in your local repo. 
+
+Once you are ready to submit a PR make sure to mark it as a draft first until it is ready to go into the master branch.
+
+When making a pull request there are many things to consider depending on the type of request.
```
adlerjohn, line 29: Condense into some paragraphs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+
+Here are some things to consider otherwise your PR may require multiple changes.
+
+- [ ] Have you formatted your code via `cargo fmt` and `forc fmt` ?
+- [ ] What is your test coverage like?
+- [ ] Is your code clean and high quality?
+- [ ] Have you extensively documented the code itself but also added documentation regarding its usage?
+- [ ] Does your code require any updates to the CI and if so do all of the cases still pass?
+- [ ] Are your commit messages descriptive and address small changes or do you bulk update everything into one giant commit (bad idea)?
```
matt-user, line 38: It may be a good idea to provide some reference to what a good commit message/commit history looks like.

CONTRIBUTING.md
```diff
@@ -0,0 +1,38 @@
...
+Once you are ready to submit a PR make sure to mark it as a draft first until it is ready to go into the master branch.
+
+When making a pull request there are many things to consider depending on the type of request.
+
+Here are some things to consider otherwise your PR may require multiple changes.
+
+- [ ] Have you formatted your code via `cargo fmt` and `forc fmt` ?
+- [ ] What is your test coverage like?
+- [ ] Is your code clean and high quality?
+- [ ] Have you extensively documented the code itself but also added documentation regarding its usage?
```
matt-user, line 36: I think grammatically this should be "Have you not only extensively documented the code, but also added documentation regarding its usage?" or an alternative "Have you extensively documented the code and also added documentation regarding its usage?"

matt-user (changes requested):
I just have some small nit comments.

CONTRIBUTING.md
Braqzen: I don't think so, there would be too many use cases and there's lots of content online about it that they should read up on. Moreover, I think line 38 is self evident.

Obviously "fixed error" is bad while "fixed an error that cause [behaviour] to occur" is a bit better. At the same time changing multiple distinct things which may spam many dozen lines is also bad to have as a single commit msg

adlerjohn: You'd be surprised at what people think are good commit messages vs not 😂. Provided some links can only help.

adlerjohn (approved):

bitzoic (approved):
Definitely will be referring to this in the future

PR #54, opened 2022-05-31
Upgrade the Escrow to be feature rich

### LINKED ISSUE #53: Upgrade the Escrow to be feature rich
Upgrade contract for better integration into an application that has a UI.

Braqzen:
## Summary

These notes are messy... if it's difficult to parse the content (or if I have missed something out) then let me know and I'll spend more time to structure this properly.

Changes
- Whoever creates an escrow is deemed to be the seller
- The seller must deposit an arbiter fee (when creating an escrow) in case there is a dispute
    - The arbiter can choose up to the limit of the fee how much they will take for resolving the dispute
    - If the buyer / seller complete the transaction then the seller gets back their deposit
- A buyer must be specified by the seller and they cannot be the arbiter or the seller
- There is a deadline after which the seller can take the funds if the buyer has not transferred for any reason
- Buyer can dispute to lock the escrow
    - A dispute prevents the seller from taking payment after the deadline
    - Only the arbiter can resolve and choose which party (buyer or seller) gets the deposit from the buyer
- Only the seller can propose a new arbiter
    - Similar to creating an escrow another deposit fee must be sent
    - If the seller proposes more than once in a row (before the buyer accepts) then the previous arbiter proposal deposit will be returned to the seller
    - If the buyer accepts a proposal then the original deposit made when creating the escrow will be returned to the seller
- Arbiter, buyer and seller can all be outputs or contracts
- When creating an escrow the seller can specify any number of assets to allow the buyer to choose a method of payment
- Logging added for the front end to pick up on
- Expanded documentation
- Each escrow only exists once. It cannot be reused. After completion a user must create a new escrow from the same contract "factory"
- Buyer can only deposit once and it must be within the deadline.
    - After depositing the buyer cannot ever get the funds back themselves
    - They can transfer to seller, seller can return funds or they can dispute for arbiter resolution
- If the seller has proposed an arbiter (which requires another deposit) and the escrow transaction is completed then they will automatically get that additional deposit back
- Seller can withdraw their collateral if the deadline has been surpassed and the buyer has not deposited

## Related
Closes #53 

## Notes

- Replaced vec with array of size 2. Need to update to use vec when SDK supports it
- Code is made specifically for the UI despite it allowing `Identity` interaction. This means that no getters are added for this improvement which means the tests do not check storage at all and instead simply check if the test panics or not. A future upgrade will need to implement contract support and thus update the tests to be more thorough
- Moreover, some `take_payment` & `withdraw_collateral` tests require SDK block manipulation therefore they are incomplete
- Factory contract will eventually be simplified again to be a single contract per escrow once parallelization is cheaper
- Used SDK version 16 because I rather do that another time :)


escrow/src/data_structures.sw
```diff
@@ -0,0 +1,83 @@
+library data_structures;
+
+use std::{contract_id::ContractId, identity::Identity, option::*};
+use core::ops::Eq;
+
+pub enum Initialized {
```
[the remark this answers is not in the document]
Braqzen: Good catch

escrow/src/main.sw
```diff
@@ -1,257 +1,355 @@
...
+        // arbiter can choose their payment amount up to the deposit from the seller
+
+        let mut escrow = storage.escrows.get(identifier);
+
+        require(escrow.state == State::Pending, StateError::StateNotPending);
+        require(escrow.disputed, StateError::NotDisputed);
+        require(msg_sender().unwrap() == escrow.arbiter.address, UserError::Unauthorized);
+        require(user == escrow.buyer.address || user == escrow.seller.address, UserInputError::InvalidRecipient);
+        require(escrow.buyer.asset.is_some() && 0 < escrow.buyer.deposited_amount, StateError::CannotResolveBeforeDesposit);
+        require(payment_amount <= escrow.arbiter.fee_amount, ArbiterInputError::PaymentTooLarge);
```
matt-user, line 239: IMO we should always just pay the arbiter their max fee amount, and then we don't have to worry about sending the seller the difference.

escrow/src/main.sw
```diff
@@ -1,257 +1,355 @@
...
-            }
-        } else {
-            revert(0);
+    #[storage(read, write)]fn create_escrow(arbiter: Arbiter, assets: [Asset;
+    2], buyer: Identity, deadline: u64) {
+        // The assertions ensure that assets are specified with a none-zero amount, the arbiter is
+        // not the buyer / seller, the arbiter has a fee that they can take upon resolving a dispute
+        // and the escrow deadline is set in the future
+
+        // require(0 < assets.len(), AssetInputError::UnspecifiedAssets);
```
matt-user, line 101: I would add a general comment somewhere to this file explaining that the commented lines will be uncommented once the SDK supports vecs.

matt-user (changes requested):
I'm not a fan of the current arbiter fee mechanism.  I think it would be much simpler to send the arbiter their fee as part of the buyer deposit whenever they resolve a dispute  This would reduce transfers and refunds between the seller and the contract, and arbiter proposal checks.  The only downside I can think of is the arbiter has to get paid with the same asset as the seller.

escrow/src/events.sw
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Unique escrow identifier
+    identifier: u64,
+}
+
+pub struct ResolvedDisputeEvent {
+    /// Unique escrow identifier
+    identifier: u64,
+
+    /// The user that can interact with the escrow
```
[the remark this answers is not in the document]
Braqzen: I'll change this to explicitly mention the buyer / seller instead

escrow/src/main.sw
```diff
@@ -1,257 +1,355 @@
...
-    IncorrectAssetAmount: (),
-    IncorrectAssetId: (),
-    StateNotInitialized: (),
-    StateNotPending: (),
-    UnauthorizedUser: (),
-    UserHasAlreadyDeposited: (),
-}
+        // TODO: incomplete compiler defaults the Option<Arbiter> to not be None therefore fee check
+        // https://github.com/FuelLabs/sway/issues/2326
+        require(arbiter.is_some() && 0 < arbiter.unwrap().fee_amount, StateError::ArbiterHasNotBeenProposed);
```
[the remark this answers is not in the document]
Braqzen: It is because the user can call this function before proposing has been set by the seller. This will allow the buyer to reset the arbiter to a zero'd data structure which may cause unexpected side effects.

escrow/src/main.sw
Braqzen: That's a subjective choice for a feature. I have opted to allow the arbiter to pick their fee instead of limiting it to more basic functionality which transfers the full amount in all cases.

escrow/src/interface.sw
```diff
@@ -0,0 +1,160 @@
...
+    /// * When the escrow is not in the State::Pending state
+    /// * When the caller is not the buyer
+    /// * When the caller deposits more than once
+    /// * When the caller sends an incorrect amount of an asset for the specified asset in the escrow
+    /// * When the caller deposits an asset that has not been specified in the escrow
+    #[storage(read, write)] fn deposit(identifier: u64);
+
+    /// Changes a flag in the escrow marking it as disputed which results in the escrow being locked
+    ///
+    /// Once the escrow is locked the seller cannot take the payment after those conditions are met
```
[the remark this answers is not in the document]
Braqzen: I've adjusted the comment. The seller cannot take payment at any point during or after a dispute. The arbiter must resolve in either direction, the buyer can transfer to the seller or the seller can return back to the buyer

Braqzen:
> I'm not a fan of the current arbiter fee mechanism. I think it would be much simpler to send the arbiter their fee as part of the buyer deposit whenever they resolve a dispute This would reduce transfers and refunds between the seller and the contract, and arbiter proposal checks. The only downside I can think of is the arbiter has to get paid with the same asset as the seller.

One party must pay a fee for the arbiter which means either the buyer or the seller must deposit collateral. In the real world, afaik, it's the seller that has to have insurance so that any buyer can interact with them. I find the buyer needing to pay X amount more to be friction for the buyer and a huge negative in terms of UX. Just thinking about it makes me not want to interact with that system. 
The buyer should have an easy experience while the seller is the one taking on the overhead.
Imagine if you had to put down collateral for any purchase that involves a third party...

I'm unsure about simplicity if your choice were to be implemented.
This is an edge case scenario but even if it did happen then you are correct that either the buyer must deposit an addition X amount of an asset (which the arbiter may not be happy with) or we need another function for the buyer to deposit a different asset which comes back to the same problem as with the seller but potentially worse as it might require 1 more function than we currently have, but maybe not.

If you'd like to come up with some code as a proposition for how much simpler it would be to make the buyer take on the fee then I'd take a look however given the first point it would have to be **super** good for the contract otherwise I cannot accept such a bad UX for the buyer.

matt-user:
> > I'm not a fan of the current arbiter fee mechanism. I think it would be much simpler to send the arbiter their fee as part of the buyer deposit whenever they resolve a dispute This would reduce transfers and refunds between the seller and the contract, and arbiter proposal checks. The only downside I can think of is the arbiter has to get paid with the same asset as the seller.
> 
> One party must pay a fee for the arbiter which means either the buyer or the seller must deposit collateral. In the real world, afaik, it's the seller that has to have insurance so that any buyer can interact with them. I find the buyer needing to pay X amount more to be friction for the buyer and a huge negative in terms of UX. Just thinking about it makes me not want to interact with that system. The buyer should have an easy experience while the seller is the one taking on the overhead. Imagine if you had to put down collateral for any purchase that involves a third party...
> 
> I'm unsure about simplicity if your choice were to be implemented. This is an edge case scenario but even if it did happen then you are correct that either the buyer must deposit an addition X amount of an asset (which the arbiter may not be happy with) or we need another function for the buyer to deposit a different asset which comes back to the same problem as with the seller but potentially worse as it might require 1 more function than we currently have, but maybe not.
> 
> If you'd like to come up with some code as a proposition for how much simpler it would be to make the buyer take on the fee then I'd take a look however given the first point it would have to be **super** good for the contract otherwise I cannot accept such a bad UX for the buyer.

So my thought process was the party that receives the deposited funds will pay the arbiter (with the deposited funds).  If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa.  To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter.  This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction.  One counterargument is: won't the seller just increase the price of their goods to account for the arbiter fee?  However, the seller can do this with the existing design as well.

In terms of code simplicity, we can remove all instances of transferring the fee amount back to the seller.  This is because either:
1.) The user (either buyer or seller) will get unspent fees back if the arbiter chooses to not select the full payment amount when resolving
2.) No transfers are necessary when a new arbiter is proposed/accepted (this is because the arbiter is now getting paid from the deposited funds rather than separately)

Braqzen:
> > > I'm not a fan of the current arbiter fee mechanism. I think it would be much simpler to send the arbiter their fee as part of the buyer deposit whenever they resolve a dispute This would reduce transfers and refunds between the seller and the contract, and arbiter proposal checks. The only downside I can think of is the arbiter has to get paid with the same asset as the seller.
> > 
> > 
> > One party must pay a fee for the arbiter which means either the buyer or the seller must deposit collateral. In the real world, afaik, it's the seller that has to have insurance so that any buyer can interact with them. I find the buyer needing to pay X amount more to be friction for the buyer and a huge negative in terms of UX. Just thinking about it makes me not want to interact with that system. The buyer should have an easy experience while the seller is the one taking on the overhead. Imagine if you had to put down collateral for any purchase that involves a third party...
> > I'm unsure about simplicity if your choice were to be implemented. This is an edge case scenario but even if it did happen then you are correct that either the buyer must deposit an addition X amount of an asset (which the arbiter may not be happy with) or we need another function for the buyer to deposit a different asset which comes back to the same problem as with the seller but potentially worse as it might require 1 more function than we currently have, but maybe not.
> > If you'd like to come up with some code as a proposition for how much simpler it would be to make the buyer take on the fee then I'd take a look however given the first point it would have to be **super** good for the contract otherwise I cannot accept such a bad UX for the buyer.
> 
> So my thought process was the party that receives the deposited funds will pay the arbiter (with a portion of the deposited funds). If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa. To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter. This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction. One counterargument is: won't the seller just increase the price of their goods to account for the arbiter fee? However, the seller can do this with the existing design as well.
> 
> In terms of code simplicity, we can remove all instances of transferring the fee amount back to the seller. This is because either: 1.) The user (either buyer or seller) will get unspent fees back if the arbiter chooses to not select the full payment amount when resolving 2.) No transfers are necessary when a new arbiter is proposed/accepted (this is because the arbiter is now getting paid from the deposited funds rather than separately)

> If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa
There is no "sellers end" unless the seller deposits. It's just the deposit from the buyer minus the fee however an arbitration fee may be much greater than the fee for the trade so the buyer would need to deposit a large amount for something cheap. That's not good.

> To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter. 
Quite the opposite for me. The deposited funds are for the product the seller is providing and not an additional fee to the arbiter. I cannot think of a use case where I would have ever needed to pay for a product and for someone else to come in and sort something out because the seller messed up. Let's suppose that happened and I, as the buyer, have deposited for the product and the fee, the arbiter will take some fee and now I am penalised for the bad behaviour of the seller. The seller has never deposited anything so they can do whatever they want only at the cost of creating the escrow.
If the buyer deposits for only the product that they are paying for and the seller is depositing an arbitration fee in case the buyer disputes then there is a convoluted resolution if the buyer is acting out. The arbiter takes a fee from the seller and sends the funds to the seller. At this point the seller would need to refund the buyer a portion which hopefully is not greater than the fee. They could just keep the funds at that point but 1) there may be legal things going on externally which prevent this or 2) if you're a malicious buyer then I have no interest in protecting your funds. If the seller keeps them at that point then you shouldn't have been malicious.
Either way a just resolution can be reached in all cases except when the buyer is not malicious and the arbiter colludes with the seller and you lose your deposit and don't get the product / service which again may have legal ramifications.

> This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction
Yes, for the seller but not the buyer. The buyer is a random person interacting with some service. They should not have to deal with the overhead of managing arbiters etc. The seller and buyer agree before the escrow or even after (since they can change it) and hopefully they never have to go into resolution. The seller is the business-person that has to navigate the legal landscape. In your scenario it's completely flipped and now every buyer must know about every service in all detail in order to interact with them rather than just showing up and depositing.

As for the last part I was implying that I'd like to see the code rewritten to be completely functional and simpler as per the suggestion rather than having to rethink the whole thing myself based on your suggestion.

matt-user:
> > So my thought process was the party that receives the deposited funds will pay the arbiter (with a portion of the deposited funds). If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa. To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter. This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction. One counterargument is: won't the seller just increase the price of their goods to account for the arbiter fee? However, the seller can do this with the existing design as well.
> > In terms of code simplicity, we can remove all instances of transferring the fee amount back to the seller. This is because either: 1.) The user (either buyer or seller) will get unspent fees back if the arbiter chooses to not select the full payment amount when resolving 2.) No transfers are necessary when a new arbiter is proposed/accepted (this is because the arbiter is now getting paid from the deposited funds rather than separately)
> 
> > If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa
> 
> There is no "sellers end" unless the seller deposits. It's just the deposit from the buyer minus the fee however an arbitration fee may be much greater than the fee for the trade so the buyer would need to deposit a large amount for something cheap. That's not good.

By "sellers end" I just mean their revenue (the funds from the buyer).  In general I assumed the arbiter fee would be a relatively small percentage of the total trade.  Why would anyone (buyer or seller) pay an arbiter fee greater than the value of the underlying trade.  They would be better off just taking the loss of the assets than paying a greater sum trying to recover them.
> > To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter.
> 
> Quite the opposite for me. The deposited funds are for the product the seller is providing and not an additional fee to the arbiter. I cannot think of a use case where I would have ever needed to pay for a product and for someone else to come in and sort something out because the seller messed up. Let's suppose that happened and I, as the buyer, have deposited for the product and the fee, the arbiter will take some fee and now I am penalised for the bad behaviour of the seller. The seller has never deposited anything so they can do whatever they want only at the cost of creating the escrow. If the buyer deposits for only the product that they are paying for and the seller is depositing an arbitration fee in case the buyer disputes then there is a convoluted resolution if the buyer is acting out. The arbiter takes a fee from the seller and sends the funds to the seller. At this point the seller would need to refund the buyer a portion which hopefully is not greater than the fee. They could just keep the funds at that point but 1) there may be legal things going on externally which prevent this or 2) if you're a malicious buyer then I have no interest in protecting your funds. If the seller keeps them at that point then you shouldn't have been malicious. Either way a just resolution can be reached in all cases except when the buyer is not malicious and the arbiter colludes with the seller and you lose your deposit and don't get the product / service which again may have legal ramifications.
> 
> > This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction
> 
> Yes, for the seller but not the buyer. The buyer is a random person interacting with some service. They should not have to deal with the overhead of managing arbiters etc. The seller and buyer agree before the escrow or even after (since they can change it) and hopefully they never have to go into resolution. The seller is the business-person that has to navigate the legal landscape. In your scenario it's completely flipped and now every buyer must know about every service in all detail in order to interact with them rather than just showing up and depositing.
> 
Don't the buyers already have to deal with some overhead of managing arbiters?  They still have to accept proposed arbiters.  Not to mention that sellers will just raise their prices to cover their arbitration costs.
> As for the last part I was implying that I'd like to see the code rewritten to be completely functional and simpler as per the suggestion rather than having to rethink the whole thing myself based on your suggestion.

After reading your comments I'm having second thoughts about my suggestion.  It's apparent to me now that my suggestion is not an obvious improvement.  That being said I think the ideal solution would be to charge the seller the arbitration fee when a dispute is ruled in favor of the buyer, and charge the buyer the arbitration fee when a dispute is resolved in favor of the seller.  However, I think the current design is good enough.

escrow/src/main.sw
```diff
@@ -1,12 +1,15 @@
...
+// TODO: commented-out code is there because the SDK does not support Vec yet so an array is used
+
 // Our library dependencies
 dep data_structures;
 dep errors;
 dep events;
 dep interface;
 
 use data_structures::{Arbiter, Asset, Buyer, EscrowInfo, Seller, State};
+// use data_structures::*;
```
matt-user: remove line as I don't think it applies to the TODO at the top of the file

matt-user (changes requested):
One small nit and then we should be good

Braqzen:
> > > So my thought process was the party that receives the deposited funds will pay the arbiter (with a portion of the deposited funds). If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa. To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter. This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction. One counterargument is: won't the seller just increase the price of their goods to account for the arbiter fee? However, the seller can do this with the existing design as well.
> > > In terms of code simplicity, we can remove all instances of transferring the fee amount back to the seller. This is because either: 1.) The user (either buyer or seller) will get unspent fees back if the arbiter chooses to not select the full payment amount when resolving 2.) No transfers are necessary when a new arbiter is proposed/accepted (this is because the arbiter is now getting paid from the deposited funds rather than separately)
> > 
> > 
> > > If the arbiter resolves the dispute in favor of the seller it will come out of the sellers end and vis versa
> > 
> > 
> > There is no "sellers end" unless the seller deposits. It's just the deposit from the buyer minus the fee however an arbitration fee may be much greater than the fee for the trade so the buyer would need to deposit a large amount for something cheap. That's not good.
> 
> By "sellers end" I just mean their revenue (the funds from the buyer). In general I assumed the arbiter fee would be a relatively small percentage of the total trade. Why would anyone (buyer or seller) pay an arbiter fee greater than the value of the underlying trade. They would be better off just taking the loss of the assets than paying a greater sum trying to recover them.
> 
> > > To me it makes the most sense to just pay the arbiter with the deposited funds in the contract, rather than depositing funds separately for the arbiter.
> > 
> > 
> > Quite the opposite for me. The deposited funds are for the product the seller is providing and not an additional fee to the arbiter. I cannot think of a use case where I would have ever needed to pay for a product and for someone else to come in and sort something out because the seller messed up. Let's suppose that happened and I, as the buyer, have deposited for the product and the fee, the arbiter will take some fee and now I am penalised for the bad behaviour of the seller. The seller has never deposited anything so they can do whatever they want only at the cost of creating the escrow. If the buyer deposits for only the product that they are paying for and the seller is depositing an arbitration fee in case the buyer disputes then there is a convoluted resolution if the buyer is acting out. The arbiter takes a fee from the seller and sends the funds to the seller. At this point the seller would need to refund the buyer a portion which hopefully is not greater than the fee. They could just keep the funds at that point but 1) there may be legal things going on externally which prevent this or 2) if you're a malicious buyer then I have no interest in protecting your funds. If the seller keeps them at that point then you shouldn't have been malicious. Either way a just resolution can be reached in all cases except when the buyer is not malicious and the arbiter colludes with the seller and you lose your deposit and don't get the product / service which again may have legal ramifications.
> > > This would also allow for the seller to create Escrows with no up front capital, which in my opinion reduces friction
> > 
> > 
> > Yes, for the seller but not the buyer. The buyer is a random person interacting with some service. They should not have to deal with the overhead of managing arbiters etc. The seller and buyer agree before the escrow or even after (since they can change it) and hopefully they never have to go into resolution. The seller is the business-person that has to navigate the legal landscape. In your scenario it's completely flipped and now every buyer must know about every service in all detail in order to interact with them rather than just showing up and depositing.
> 
> Don't the buyers already have to deal with some overhead of managing arbiters? They still have to accept proposed arbiters. Not to mention that sellers will just raise their prices to cover their arbitration costs.
> 
> > As for the last part I was implying that I'd like to see the code rewritten to be completely functional and simpler as per the suggestion rather than having to rethink the whole thing myself based on your suggestion.
> 
> After reading your comments I'm having second thoughts about my suggestion. It's apparent to me now that my suggestion is not an obvious improvement. That being said I think the ideal solution would be to charge the seller the arbitration fee when a dispute is ruled in favor of the buyer, and charge the buyer the arbitration fee when a dispute is resolved in favor of the seller. However, I think the current design is good enough.

> Why would anyone (buyer or seller) pay an arbiter fee greater than the value of the underlying trade.

Yeah, I didn't want to make any assumptions in the code that could potentially force one amount or another.

> Don't the buyers already have to deal with some overhead of managing arbiters? They still have to accept proposed arbiters. 

My thoughts are that they only have to approve them off-chain (this could be as simple as looking up an address on some website from a firm) and if another proposition is done during the lifetime of the escrow then it's the same process but now they have to approve a call to change the address and fee amount.
This only reduces the friction in so much as the buyer has to agree that they seem reputable enough to accept and thus no further research or insurance etc. has to be done - but it could be and arguably it should be. Nevertheless, it's the job of the seller to provide a service which as you've mentioned can be, and usually is, factored into the fee.

> That being said I think the ideal solution would be to charge the seller the arbitration fee when a dispute is ruled in favor of the buyer, and charge the buyer the arbitration fee when a dispute is resolved in favor of the seller. However, I think the current design is good enough.

When a resolution is performed by the arbiter the arbiter can select an amount they would like to take as payment from the collateral of the seller. With this action they select either the buyer or seller as the party that the deposit from the buyer goes to. This means that if the buyer is at fault the full deposit goes to the seller and some portion of the fee goes to the arbiter (in most cases it's probably going to be all of it otherwise some gets returned to the seller). On the other hand if the seller is at fault then the buyer gets their deposit back and the seller pays some fee again.
In effect, the buyer can be charged "an arbitration fee" since they lose their deposit.
I agree that the design isn't perfect. It took a while to think of something functional that isn't inherently too bad for anyone involved. One of those ideas was to deal with a % fee which was too much of a pain so a flat rate is set.
Another idea was to have both parties deposit collateral too.

escrow/src/main.sw
Braqzen: This has been resolved in [this commit](https://github.com/FuelLabs/sway-applications/pull/54/commits/b966c5d460bd68af8b564b89a8fb044b3416ce74) a little higher up

matt-user: lol whoops

matt-user (approved):

escrow/src/main.sw
```diff
@@ -1,257 +1,346 @@
...
+        require(escrow.disputed, StateError::NotDisputed);
+        require(msg_sender().unwrap() == escrow.arbiter.address, UserError::Unauthorized);
+        require(user == escrow.buyer.address || user == escrow.seller.address, UserInputError::InvalidRecipient);
+        require(escrow.buyer.asset.is_some() && 0 < escrow.buyer.deposited_amount, StateError::CannotResolveBeforeDesposit);
+        require(payment_amount <= escrow.arbiter.fee_amount, ArbiterInputError::PaymentTooLarge);
+
+        escrow.state = State::Completed;
+        storage.escrows.insert(identifier, escrow);
+
+        transfer(escrow.buyer.deposited_amount, escrow.buyer.asset.unwrap(), user);
```
bitzoic, line 235: How does the seller get their original assets back? Who do they go to? Since the `escrow.state` is set to `State::Completed` isn't it locked in the contract forever?

bitzoic (commented):
Will the arbiter become an `Option` once supported by the SDK? 
A large part of the ethos of defi is getting rid of the middle man, settling things on-chain, and stricly with code. Introducing someone in the middle and depending on an external party to resolve a conflict is what blockchain is supposed to circumnavigate. While it may be useful in some cases, I believe we should be careful in reintroducing the same probelms we see in the real world today.

Braqzen:
> Will the arbiter become an `Option` once supported by the SDK? A large part of the ethos of defi is getting rid of the middle man, settling things on-chain, and stricly with code. Introducing someone in the middle and depending on an external party to resolve a conflict is what blockchain is supposed to circumnavigate. While it may be useful in some cases, I believe we should be careful in reintroducing the same probelms we see in the real world today.

> Will the arbiter become an Option once supported by the SDK?
Nope, there must always be an arbiter under this implementation so no options. There may be a spin off contract which is much simpler in the future but that will require careful consideration again.

An escrow requires an arbiter otherwise it doesn't make sense to use an escrow because if either party can just pull out their collateral / deposit at any time then there's no protection in the exchange for either party. Perhaps there are implementations of escrows that have somehow circumvented this but afaik an escrow always requires a 3rd party otherwise it cannot work.
If you know of any implementations then let me know. Initially I wanted to completely remove the 3rd party i.e. no arbiter, no oracle etc. but I couldn't think of a safe way to implement the escrow without 1 party screwing over another in some situation.

escrow/src/interface.sw
```diff
@@ -0,0 +1,161 @@
+library interface;
+
+dep data_structures;
+
+use data_structures::{Arbiter, Asset};
+use std::{identity::Identity, vec::Vec};
```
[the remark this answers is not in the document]
Braqzen: Right, I forgot to remove that when I have switched back to the array for testing purposes

escrow/src/interface.sw
```diff
@@ -0,0 +1,161 @@
...
+    /// # Arguments
+    ///
+    /// * `identifier` - Identifier used to find a specific escrow
+    ///
+    /// # Reverts
+    ///
+    /// * When the escrow is not in the State::Pending state
+    /// * When the caller is not the buyer
+    /// * When the arbiter has not been proposed by the seller
+    #[storage(read, write)]fn accept_arbiter(identifier: u64);
```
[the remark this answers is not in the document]
Braqzen: I'm unsure what you mean by the last comment nevertheless once a script factory is a thing then identifiers will be unneccessary so that will be removed.

escrow/src/main.sw
```diff
@@ -1,257 +1,346 @@
...
 
-    #[storage(read, write)]fn approve() -> bool;
+    /// Number of created escrows
+    /// Used as an identifier for O(1) look-up in mappings
+    escrow_count: u64 = 0,
+}
 
-    #[storage(read, write)]fn withdraw() -> bool;
+impl Escrow for Contract {
+    #[storage(read, write)]fn accept_arbiter(identifier: u64) {
```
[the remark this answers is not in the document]
Braqzen: No need since the escrow can only have 1 value set at a time so the user should be able to see the proposition and all they have to do is accept it otherwise I'd need to implement some juggling of any number of arbiters being proposed at a time and then handle all of those fund transfers.

escrow/src/main.sw
```diff
@@ -1,257 +1,346 @@
...
+        storage.escrows.insert(identifier, escrow);
+
+        transfer(escrow.buyer.deposited_amount, escrow.buyer.asset.unwrap(), user);
+        transfer(payment_amount, escrow.arbiter.asset, escrow.arbiter.address);
+
+        if payment_amount != escrow.arbiter.fee_amount {
+            transfer(escrow.arbiter.fee_amount - payment_amount, escrow.arbiter.asset, escrow.seller.address);
+        }
+
+        // If there is a previous proposal then we must transfer those funds back to the seller
```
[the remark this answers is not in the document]
Braqzen: It could be but the issue is the storage which updates the proposal. That cannot be done in a library therefore I would need to duplicate the if conditional check multiple times in order to update the proposal at which point it's probably the same amount of lines anyway so I do not see the benefit

escrow/src/main.sw
```diff
@@ -1,257 +1,346 @@
...
-    #[storage(read)]fn get_user_data(user: Address) -> (bool, bool) {
-        // require(storage.state != State::Void, Error::StateNotInitialized);
-        require(storage.state != 0, Error::StateNotInitialized);
-        require(user == storage.buyer.address || user == storage.seller.address, Error::UnauthorizedUser);
-
-        if user == storage.buyer.address {
-            (storage.buyer.deposited, storage.buyer.approved)
-        } else {
-            (storage.seller.deposited, storage.seller.approved)
+    #[storage(read, write)]fn take_payment(identifier: u64) {
```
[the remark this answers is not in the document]
matt-user: My understanding is that the buyer receives the selling asset off chain, hence the need for an arbiter

escrow/src/main.sw
```diff
@@ -1,257 +1,346 @@
...
+        require(arbiter.asset == msg_asset_id(), ArbiterInputError::AssetDoesNotMatch);
+
+        // If there is a previous proposal then we must transfer those funds back to the seller
+        let proposal = storage.arbiter_proposal.get(identifier);
+        if proposal.is_some() && 0 < proposal.unwrap().fee_amount {
+            transfer(proposal.unwrap().fee_amount, proposal.unwrap().asset, escrow.seller.address);
+        }
 
-        true
+        storage.arbiter_proposal.insert(identifier, Option::Some(arbiter));
```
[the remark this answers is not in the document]
Braqzen: In Solidity we perform all state changes first to prevent re-entrency / use a modifier. In Sway, when transferring, that is unnecessary because the transfer does not allow re-entrency since it's not a standard contract call. It simply performs a transfer so there is no vector for re-entrency when transferring. That being said it's still a good idea to keep that pattern in mind.

escrow/src/main.sw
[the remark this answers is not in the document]
Braqzen: The way the process works is the buyer deposits an asset into the escrow and then they wait for something to happen off-chain. Once that occurs they can use the `transfer_to_seller()` to send their deposit to the seller. If for whatever reason they do not do that and the deadline is surpassed then the seller can `take_payment()` instead. If the buyer does not want to allow that, because the trade did not occur as expected, they can `dispute()` at which point an arbiter must come in to resolve OR if the arbiter does not resolve in time the seller can `return_deposit()` or the buyer can still `transfer_to_seller()`

escrow/src/main.sw
Braqzen: When a resolution is performed by the arbiter the arbiter gets to select who the deposit from the buyer goes to i.e. the buyer or seller. Since the arbiter has performed work they get to be rewarded with the collateral from the seller. The arbiter gets to select up to the maximum amount of collateral from the seller and the difference will be sent back to the seller so for example if the seller has some asset of quantity 100 and the arbiter thinks that their work was only worth 80 then the remaining 20 go back to the user (as seen on line 238).

This leaves a potential deposit from the seller for a change of arbiter and thus on line 243 the code checks to see if the seller has made proposal for a change of arbiter (which requires collateral) and if the buyer has not accepted then it's stuck there so we must now send that back.

escrow/src/main.sw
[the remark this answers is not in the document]
bitzoic: Ah I see, the off-chain portion went over my head. This makes a lot more sense.

escrow/src/data_structures.sw
```diff
@@ -0,0 +1,115 @@
...
+
+pub struct Buyer {
+    /// Address identifying the buyer
+    address: Identity,
+
+    /// The asset that the user has currently deposited in the contract
+    asset: Option<ContractId>,
+
+    // Minor data duplication allows us to forego validating unique assets upon escrow creation
+    // otherwise the same asset with different values can be added which, if handled incorrectly,
```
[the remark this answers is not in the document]
Braqzen: I don't agree with this suggestion because it seems to break the sentence incorrectly into two, furthermore, it appears to duplicate line 38 onto line 39 as well

bitzoic:
> > Will the arbiter become an `Option` once supported by the SDK? A large part of the ethos of defi is getting rid of the middle man, settling things on-chain, and stricly with code. Introducing someone in the middle and depending on an external party to resolve a conflict is what blockchain is supposed to circumnavigate. While it may be useful in some cases, I believe we should be careful in reintroducing the same probelms we see in the real world today.
> 
> > Will the arbiter become an Option once supported by the SDK?
> > Nope, there must always be an arbiter under this implementation so no options. There may be a spin off contract which is much simpler in the future but that will require careful consideration again.
> 
> An escrow requires an arbiter otherwise it doesn't make sense to use an escrow because if either party can just pull out their collateral / deposit at any time then there's no protection in the exchange for either party. Perhaps there are implementations of escrows that have somehow circumvented this but afaik an escrow always requires a 3rd party otherwise it cannot work. If you know of any implementations then let me know. Initially I wanted to completely remove the 3rd party i.e. no arbiter, no oracle etc. but I couldn't think of a safe way to implement the escrow without 1 party screwing over another in some situation.

After realizing that this involves off-chain transactions, yes a 3rd party is indeed needed. The original escrow involved only on-chain assets hence my confusion.

This should be included in the documentation. I also believe in that case it would be a good idea to have 2 different escrow applications, one for on-chain assets and another for off-chain in the same manner that we have different auction applications.

escrow/src/main.sw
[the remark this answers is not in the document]
bitzoic: The one problem that I see is the seller creating a new arbiter proposal a block before the buyer calls `accept_arbiter` not giving the front end time to update. This may be a stretch but a possible attack vector and something worth considering.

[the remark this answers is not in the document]
Braqzen: I don't think that can be solved on-chain. We don't know block times so I cannot select an arbitrary number of blocks and that may also mess up the deadline. If I did know the times that wouldn't help much anyway imo. This is an edge case as it would need to occur right before the acceptance which is unlikely to happen or front running must be done and the chance of some node allowing this specifically is also unlikely so it's possible but very slim.

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Actions that users are able to perform
+
+This sub-section details what a user is able to do e.g. click a button and "x, y, z" happens.
+
+### Arbiter
+
+#### `resolve_dispute()`
+
+1. When an escrow is in dispute the arbiter of the escrow can perform a resolution in either direction of the buyer or seller
```
matt-user, line 36:
```suggestion
1. When an escrow is in dispute the arbiter of the escrow can perform a resolution in either the direction of the buyer or seller
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### `accept_arbiter()`
+
+1. The buyer is able to accept a proposition by the seller to change the arbiter or change the fee for the arbiter
+
+> **NOTE** They are not required to do so however there may be instances where changing the arbiter / fee is favourable
+
+#### `deposit()`
+
+1. The buyer must deposit into the escrow one asset from the options provided by the seller
```
matt-user, line 51:
```suggestion
1. The buyer must deposit into the escrow one asset from the list of approved assets provided by the seller
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+2. The escrow accepts a 1 time deposit therefore the buyer cannot go back and deposit a different asset
```
matt-user, line 52:
```suggestion
2. The escrow accepts a 1 time deposit.  Therefore the buyer cannot go back and deposit a different asset
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+
+> **NOTE** Once a buyer deposits they are unable to get the asset out by themselves. Either the arbiter must resolve in their favour or the seller must return the deposit. This is a safety mechanism.
+
+#### `dispute()`
+
+1. The buyer is able to `dispute()` the escrow which prevents the seller from taking their deposit after the deadline
+2. Disputing means that there are 3 ways to move the deposit from the buyer out of the escrow
```
matt-user, line 59:
```suggestion
2. Disputing means that the contract now allows 3 ways to move the deposit from the buyer out of the escrow
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+   1. The buyer transfers to seller
+   2. The seller returns the deposit to the buyer
+   3. The arbiter may come in and resolve the dispute in favour of the buyer or seller
```
matt-user, line 62: It may be useful to a reader to include the functions that perform the operations mentions (`transfer_to_seller`) etc

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. When the buyer is satisfied with the exchange then they are able to transfer the funds to the seller
+
+> **NOTE** They are not required to do so however being a good samaritan is nice. The seller can take the payment later themselves if the escrow is not in dispute
+
+### Seller
+
+#### `create_escrow()`
+
+1. Any user that creates an escrow is automatically considered as the seller
```
matt-user, line 74:
```suggestion
1. Any user that creates an escrow is automatically considered the seller
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+2. Creating an escrow requires collateral as payment for a possible arbitration
```
matt-user, line 75:
```suggestion
2. Creating an escrow requires depositing collateral as payment for a possible arbitration
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+3. When creating an escrow the seller can choose any assets that they are willing to take as payment from the buyer
```
matt-user, line 76:
```suggestion
3. When creating an escrow the seller provides a list of assets that they are willing to accept as payment from the buyer
```

escrow/SPECIFICATION.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### `propose_arbiter()`
+
+1. The seller may propose a change to the arbiter or arbiter fee
+   1. This requires a new deposit as collateral
+2. This can be done an unlimited number of times
+   1. If this is done more than once in a row before the buyer accepts then the previous collateral will automatically be returned to the seller
+3. If the buyer accepts then the previous collateral will be returned to the seller
+4. If the buyer does not accept then upon completion of the escrow the funds will be returned to the seller
```
matt-user, line 85:
```suggestion
4. If the buyer does not accept then upon completion of the escrow the unused collateral will be returned to the seller
```

matt-user (changes requested):
Some small nits that I think improve the readability and information provided by the spec

escrow/SPECIFICATION.md
Braqzen: This is incorrect

escrow/SPECIFICATION.md
Braqzen: This implies the seller has done something to approve an asset. It doesn't really make sense especially in the context of the buyer actually having a function to approve

escrow/SPECIFICATION.md
Braqzen: This is also incorrect

escrow/SPECIFICATION.md
Braqzen: Technically the contract allowed 3 ways before & 3 ways after the dispute but now one of the options has changed so "now" doesn't really make sense here

escrow/SPECIFICATION.md
Braqzen: > to include the functions that perform the operations mentions

I do not understand.

I am planning on adding diagrams which show a visual contract flow of all the use cases

matt-user: I mean it may be useful to mention the functions that perform the 3 different ways to move the deposit from the buyer out of the escrow.  `transfer_to_seller`, `resolve_dispute`, and `return_deposit`.  Although if you add diagrams in the future this may not be a useful addition.

escrow/SPECIFICATION.md
matt-user: I plugged both options into an online grammar tool and they both work.  I guess it's just a matter of opinion.

escrow/SPECIFICATION.md
matt-user: 1. hmm I'm not sure this is incorrect.  I think the "therefore" needs some sort of punctuation around it since you are using it to connect two [independent clauses](https://www.wikihow.com/Use-Therefore-in-a-Sentence), but also grammar is not my strongest lol.

matt-user: Also I think the original sentence works well enough if you don't feel like changin.

Braqzen: Same here. Looking at that link I don't know why they end the sentence and start with "Therefore". I'm pretty sure that is incorrect or rather it can be done but that's not the correct way to use it. Afaik it's meant to connect and the common use may / must involve a comma at times

matt-user: Feel free to add punctuation or don't.  I think readers will be able to understand the original.

simonr0204: Deep into bike-shedding territory here, but technically, it's either:

> "The escrow accepts a one-time deposit.  Therefore, the buyer ... " 

or

> "The escrow accepts a one-time deposit; therefore, the buyer ... "

Braqzen: hm, I thought something like the second one was the correct way to do it while the first is incorrect. Guess I was never taught the first way

matt-user (approved):
Nice additions!

matt-user (approved):

simonr0204 (approved):
Wow, that was a lot of reading :)   

I'm coming to this after some already pretty thorough reviews so all looks good to me!

The main test harness is a pretty massive file, but I see you've already abstracted as much of the setup out as is reasonable. I worry slightly about maintainability if something changes in the SDK but I guess there's a balance between a large main test file and endless abstractions.

PR #67, opened 2022-06-04
Fundraiser program

### LINKED ISSUE #65: Create a fundraiser program that is fully tested
# Requirements

1. As an author I should be able to 
    - Specify the asset that is used to measure against a target goal amount
    - Specify how long the crowdfund will last
    - Claim the amount once the goal is reached
2. As a non-author I should be able to
    - Pledge any amount of the specified asset if the fund has not expired
    - Withdraw my pledge at any point after my pledge unless the goal has been reached

Extension: The contract should be a single contract that can spawn any number of independent fundraisers / campaigns. This can be done via handling data through various mappings.

Braqzen:
# Summary

## Type of change

- New feature

## Changes

- Added to CI
- Rough Specification outlining the application (WIP)
- Allows the user to create a campaign which other users can pledge to
- Allows the user to cancel their campaign and claim upon the goal being reached
- Allows users to pledge to campaigns if they pledge with the asset that the campaign accepts (before the deadline and if the campaign has not been cancelled / claimed)
- Allows users to unpledge any amount up to whatever they have pledged as long as the campaign has not been claimed
- Various utility functions in preparation for integration with the user interface (soon TM)

## Notes

- Contract does not support campaign discovery (ref the spec for terminology) since the spec needs to be fleshed out
- Need to update README.md eventually so that it does not contain just a skeleton

# Related Issues

Closes #65


adlerjohn:
For posterity, can you add which Sway issues need to be resolved as blockers in the PR description?

.github/scripts/build.sh
```diff
@@ -7,4 +7,6 @@ if [ $PROJECT = 'escrow' ]; then
     forc build --path $PROJECT/tests/artifacts/asset
 elif [ $PROJECT = 'multisig-wallet' ]; then
     forc build --path $PROJECT
+elif [ $PROJECT = 'fundraiser' ]; then
```
adlerjohn, line 10: Maybe alphabetize this one to avoid merge conflicts with everyone putting stuff at the end.

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     strategy:
       matrix:
-        project: ["escrow", "multisig-wallet"]
+        project: ["escrow", "multisig-wallet", "fundraiser"]
```
adlerjohn, line 48: Ditto.

adlerjohn:
Is this one ready for review?

adlerjohn:
> Aside from that, yeah. I could mark it as ready.

Please do! Doesn't have to be perfect, can always be improved in iterations.

Braqzen:
Adding @digorithm @emilyaherbert @mohammadfawaz @otrho and @sezna to specifically comment on my testing. I'd like to know if that is okay or if there are better conventions that I could uptake.

Braqzen:
Furthermore, I'll be waiting to see reviews specifically from @nfurfaro @matt-user @bitzoic and @SwayStar123 since these are the latest and greatest conventions that I'd like to follow in sway-apps (until better ones are found).

digorithm (dismissed):
I really liked the way you organized the tests! Separating concerns/paths with modules and especially the way you separated the `abigen!` + setup stuff from the core test logic. Really awesome, I will definitely, at some point, get some of these structuring ideas to the SDK docs!

Note: this approval is specifically for the Rust SDK code, I'll defer the Sway-related reviews to the other requested folks.

fundraiser/tests/harness.rs
```diff
@@ -0,0 +1,1350 @@
...
+// TODO: Until the SDK supports block manipulation changing tests may break them because of the
+//       specifically selected block deadlines so your test might be correct but the deadline is
+//       messing up the test
+//
+//  - claim_pledges
+//      - revert_when_claiming_before_deadline (need SDK to manipulate block height)
+//  - pledges
+//      - revert_when_pledging_after_deadline (need SDK to manipulate block height)
+//
+//      When logging is deserialized in the SDK, check logs are correct
```
digorithm, line 26: @Voxelot Here's an example of the need for the block manipulation/configuration thing we talked about recently. More specifically: https://github.com/FuelLabs/sway-applications/pull/67/files#diff-d3c232d3492ac06233511f72153a04c8e079fe6c141469787897a2b712af319dR460. 

Tagging you so you have some more motivation/context for that feature; it'll require some major SDK + client coordination to get this out.

fundraiser/src/main.sw
```diff
@@ -0,0 +1,448 @@
...
+        // The user should not be able to continue to pledge if the campaign has been cancelled
+        // Given the logic below it's unnecessary but it makes sense to stop them
+        require(!campaign_info.cancelled, CampaignError::CampaignHasBeenCancelled);
+
+        // Use the user's pledges as an ID / way to index this new pledge
+        let user = sender_identity();
+        let pledge_count = storage.pledge_count.get(user);
+
+        // First time pledging to any campaign
+        if pledge_count == 0 {
```
bitzoic, line 257: Is checking for 0 necessary? if `pledge_count` is 0 then `storage.pledge_count.insert(user, pledge_count + 1);` from below would accomplish the same thing as `storage.pledge_count.insert(user, 1);`

Braqzen: Yeah, you're correct. My reasoning was that at first they have not pledged otherwise there is a pledge but it might be to a new campaign or an existing campaign so I would have an initial check and then another branch between those latter two. The logic can be updated (including the comments) to use the pledge index instead.

fundraiser/src/utils.sw
```diff
@@ -0,0 +1,17 @@
...
+
+use errors::UserError;
+
+pub fn sender_identity() -> Identity {
+    // TODO: use msg_sender().unwrap() (and remove this function) when that becomes functional
+    let sender: Result<Identity, AuthError> = msg_sender();
+    sender.unwrap()
+}
+
+pub fn validate_id(id: u64, count: u64) {
```
matt-user, line 15: Is there a reason why you made this require check its own function and the other require checks not?

matt-user (approved):
Overall it looks good.  I'll definitely take note of your test structure src/ structure when implementing sway programs.

fundraiser/tests/harness.rs
Voxelot: Should we be using timestamps instead of block numbers? I know there were some spec discussions about this, but I think timestamps would be safer for dapps as well as better for testing imo

fundraiser/src/utils.sw
Braqzen: Keeping the code DRY (don't repeat yourself). The other checks do not get used over and over. This one does. It makes it easier to manage if it were to be changed.

fundraiser/src/main.sw
```diff
@@ -0,0 +1,438 @@
...
+    /// # Reverts
+    ///
+    /// * When `asset` is the BASE_ASSET
+    /// * When the `deadline` is not ahead of the current block height
+    /// * When the `target_amount` is 0
+    /// * When an AuthError is generated
+    #[storage(read, write)]fn create_campaign(asset: ContractId, beneficiary: Identity, deadline: u64, target_amount: u64) {
+        // Prevent a user from ever having the ability to accept the base asset in case that becomes
+        // a contract - which may be used as a burned address similar to ETH 0x0000...
+        require(asset.value != BASE_ASSET_ID, CreationError::CannotUseBaseAsset);
```
bitzoic, line 79: I've previously added `BASE_ASSET_ID` checks to my contracts in the case of something being burned, but have gone the route of instead using an `Option` and setting to `None`. For example in my NFT contract, the owner is now an `Option` and can be set to `None` so that any `Address` can be used including the `BASE_ASSET_ID` address. Will this be a reserved address?

fundraiser/tests/harness.rs
adlerjohn: Timestamps should be used exclusively for dealing with the passing of time. That being said, given that there's no real money involved yet, block number is a drop-in replacement that serves the same purpose.

fundraiser/src/main.sw
Braqzen: It was mentioned, in some conversation under a context I do not recall, that the 0th address may eventually be a contract (emphasise on the "may"). If that occurs and people treat it as a burner nonetheless then it's a good idea to block anything coming from there.

bitzoic (dismissed):

fundraiser/src/main.sw
Braqzen: Once the base becomes a flash loan contract then I don't know what to expect exactly. I don't want any nonsense happening so it's blocked

fundraiser/src/main.sw
```diff
@@ -0,0 +1,505 @@
...
+
+    /// Direct look-up for asset data if the user wants to check via a known ID
+    asset_info: StorageMap<ContractId,
+    AssetInfo>, /// O(1) look-up to allow searching via asset_count
+    /// Map(1...asset_count => asset)
+    asset_index: StorageMap<u64,
+    ContractId>, /// The total number of unique campaigns that a user has created
+    /// This should only be incremented
+    /// Cancelling / Claiming should not affect this number
+    campaign_count: StorageMap<Identity,
```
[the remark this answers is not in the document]
Braqzen: Noted. I've had a similar thought for a moment however staring at this for so long made this another oversight

fundraiser/src/main.sw
```diff
@@ -0,0 +1,505 @@
...
+    /// Map(Campaign ID => CampaignInfo)
+    campaign_info: StorageMap<u64,
+    CampaignInfo>, /// The total number of unique campaigns that a user has pledged to
+    /// This should only be incremented.
+    /// Unpledging should not affect this number
+    pledge_count: StorageMap<Identity,
+    u64>, /// Record of how much a user has pledged to a specific campaign
+    /// Locked after the deadline
+    /// Map(Identity => Map(1...pledge_count => Pledge))
+    pledge_history: StorageMap<(Identity,
```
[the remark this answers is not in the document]
Braqzen: I don't understand. 
It uses 2 mappings. One is for a look up via a campaign ID and the other is to allow the UI to iterate over the count otherwise the UI needs to constantly have knowledge of the relationship between each campaign and all users

fundraiser/src/main.sw
Braqzen: I've already removed similar checks in other branches. Will likely do that here.

SwayStar123 (approved):
Looks good

fundraiser/src/main.sw
```diff
@@ -0,0 +1,502 @@
...
+    ///
+    /// * `id` - Unique identifier which is a number starting from 1...storage.user_campaign_count
+    ///
+    /// # Reverts
+    ///
+    /// * When the `id` is either 0 or greater than the total number of campaigns created by the author
+    /// * When an AuthError is generated
+    #[storage(read)]fn campaign(id: u64) -> Campaign {
+        // Validate the ID to ensure that the user has created the campaign
+        require(id != 0 && id <= storage.user_campaign_count.get(sender_identity()), UserError::InvalidID);
```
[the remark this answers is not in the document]
Braqzen: Looking at the code again it does look like I can repurpose the function to apply here too under a different enum error. I think I have brushed @matt-user off when a comment was made although at the time the logic may have been a little different.

adlerjohn (dismissed):

fundraiser/src/utils.sw
```diff
@@ -0,0 +1,18 @@
...
+
+use std::{assert::require, chain::auth::{AuthError, msg_sender}, identity::Identity, result::*};
+
+use errors::UserError;
+
+pub fn sender_identity() -> Identity {
+    // TODO: use msg_sender().unwrap() (and remove this function) when that becomes functional
+    // Related: https://github.com/FuelLabs/sway/pull/1958
+    let sender: Result<Identity, AuthError> = msg_sender();
+    sender.unwrap()
```
nfurfaro, line 13: Weren't you just saying that this should be able to be chained, without the temp variable `sender` ?

fundraiser/src/main.sw
```diff
@@ -0,0 +1,502 @@
...
+        let campaign_info = CampaignInfo {
+            asset, author: user,
+            beneficiary, cancelled: false,
+            claimed: false,
+            deadline, target_amount, total_pledge: 0,
+        };
+
+        // Keep track of new assets
+        let mut asset_info = storage.asset_info.get(asset);
+        if !asset_info.exists {
```
nfurfaro, line 105: Could you also just check `if !asset_info.amount == 0 {` ?

fundraiser/src/abi.sw
```diff
@@ -0,0 +1,36 @@
...
+
+    #[storage(read)]fn campaign(campaign_history_index: u64) -> Campaign;
+
+    #[storage(read)]fn pledge_count() -> u64;
+
+    #[storage(read)]fn pledged(pledge_history_index: u64) -> Pledge;
+
+    #[storage(read)]fn asset_count() -> u64;
+
+    #[storage(read)]fn asset_info_by_address(asset: ContractId) -> AssetInfo;
```
nfurfaro, line 33: If we're being picky, I would suggest renaming to `asset_info_by_contract_id` (or maybe just `asset_info_by_id`), seeing as we don't actually use the `Address` type here.

fundraiser/src/main.sw
```diff
@@ -0,0 +1,502 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `id` - Unique identifier which is a number starting from 1...storage.user_campaign_count
+    ///
+    /// # Reverts
+    ///
+    /// * When the `id` is either 0 or greater than the total number of campaigns created by the author
+    /// * When an AuthError is generated
+    #[storage(read)]fn campaign(id: u64) -> Campaign {
```
nfurfaro, line 444: I don't think we need to change this now, but in general, the approach of using `msg_sender` for this function (rather than accepting a second param like `user: Identity`) only gives a small amount of pseudo-privacy, while preventing possible future integrations (i.e: A third-party contract that interacts with the fundraiser contract on behalf of a user). Passing a second param here would enable composability in the greater ecosystem (which is currently non-existant, so could be a future improvement to make).

fundraiser/src/main.sw
```diff
@@ -0,0 +1,502 @@
...
+        // Validate the ID to ensure that the user has created the campaign
+        validate_id(id, storage.user_campaign_count.get(sender_identity()));
+        storage.campaign_history.get((sender_identity(), id))
+    }
+
+    /// Returns the number of campaigns that the user has pledged to
+    ///
+    /// # Reverts
+    ///
+    /// * When an AuthError is generated
```
nfurfaro, line 454: I'm not sure what the correct approach is here but do we need to list all possible reverts all the way down the stack, or just in the specific function we're documenting?
AutError here is from the stdlib, and probably doesn't need to be mentioned in every function which uses `msg_sender()`?

fundraiser/src/main.sw
```diff
@@ -0,0 +1,502 @@
...
+        // The users should only have the ability to pledge to campaigns that have not reached their
+        // deadline (ended naturally - not been cancelled)
+        require(height() < campaign_info.deadline, CampaignError::CampaignEnded);
+
+        // The campaign specifies an asset that it accepts therefore the user must pledge the correct
+        // asset in order to update the state of the campaign
+        require(campaign_info.asset == msg_asset_id(), UserError::IncorrectAssetSent);
+
+        // A user cannot pledge zero since it does not make sense to do so
+        require(0 < msg_amount(), UserError::AmountCannotBeZero);
```
nfurfaro, line 268: Maybe check this even before reading the campaign_info from storage. If `msg_amount() == 0` there's no point doing a storage read.

fundraiser/src/utils.sw
Braqzen: I think this may have been just before the release which fixed this

fundraiser/src/main.sw
Braqzen: I should only be documenting errors generated by the contract and not the library itself. I'll make all of these adjustments soon

fundraiser/src/main.sw
Braqzen: I think the order makes sense as it is.
When you pledge you must check that what you are pledging to is active, what you are trying to pledge is the correct type and the amount is also correct.
Trying to change the order doesn't make sense in the real world.

If and when we are able to extract the error messages then it will become more relevant.

nfurfaro: I disagree. 
Of all the checks needed in this function, this is the only one that can actually be done before reading from storage (and incurring extra gas costs).
No point in reading storage and making a number of other checks first if we already know the function is going to revert based on `msg_value()`.

In general, "Fail hard and fail early" :)

Braqzen: This is not something that I will budge on. Semantically it makes sense to perform the checks this way and if you're trying to break the system then you deserve to be punished. This only costs those who intentially pledge 0. No one who has any brain cells will actually pledge 0 because that doesn't make sense to do so therefore if you are trying to perform boundary conditions / hacks then that additional cost is on you.

That being said I do agree with your last sentence.

nfurfaro (approved):

PR #78, opened 2022-06-08
Dutch auction

### LINKED ISSUE #69: Dutch auction program

SwayStar123:
# Summary

## Type of change

- New Feature

## Changes

- Adding a new dutch auction program

# Related Issues

Closes #69


dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+storage {
+    startingPrice: u64,
+    endingPrice: u64,
+    startTime: u64,
+    endTime: u64,
+    beneficiary: Address = ~Address::from(MY_ADDRESS),
+    admin: Address = ~Address::from(MY_ADDRESS),
+    ended: bool,
+    //Enter the asset id of the currency you want the bid to be in
+    asset_id: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000,
```
Braqzen, line 35: Alphabetical order.
Document each thing so that I know what they are there for (use 3 forward slashes followed by a space instead of two forward slashes and no space).
When you document add a line break between the variables e.g.

```rust
/// This is a line of documentation
my_var: some_type,

/// This is another line
another_var: another_type,
```

No hardcoding and asset_id should be of type `ContractId`

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+
+fn win() {
+    // Do stuff on the win event
+    transfer_to_output(price(), ~ContractId::from(storage.asset_id), storage.beneficiary);
+    storage.ended = true;
+}
+
+impl MyContract for Contract {
+    fn get_price() -> u64 {
+        return price();
```
Braqzen, line 46: rename to `_get_price()` and the `return` can be implicit so remove the `return` and the `;` at the end

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+            revert(0);
+    };
+
+    let ad = if let Identity::Address(addr) = unwrapped {
+        addr
+    } else {
+        revert(0);
+    };
+    ad
+}
```
Braqzen, line 119: This can be removed and where ever you use it you can use the `match` syntax. Take a look at [this sample](https://github.com/FuelLabs/sway-applications/pull/54/files#diff-d36d108de6182b8a05f10d7d1558871f44d9166ce29aeb69ff27c43459a45188R191-R198)

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    fn get_price() -> u64;
+    fn set_beneficiary(new_beneficiary: Address);
+    fn bid();
+}
+
+//Set this to your own address
+const MY_ADDRESS: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+
+storage {
+    startingPrice: u64,
```
Braqzen, line 27:
```suggestion
    opening_price: u64,
```

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+        assert(startp > endp);
+        assert(endt > height());
+        assert(startt > height());
+        assert(endt > startt);
+
+        storage.startingPrice = startp;
+        storage.endingPrice = endp;
+        storage.startTime = startt;
+        storage.endTime = endtt;
+        storage.ended = false;
```
Braqzen, line 82: No need to set `ended` to false since it is false by default

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+fn price() -> u64 {
+    let price_difference = storage.startingPrice - storage.endingPrice;
+    let duration = storage.endTime - storage.startTime;
+    // This is the amount the price will reduce by per block
+    let price_shift = price_difference / duration;
+
+    let now = height() - storage.startTime; //Current block height - start will tell us how far we are into the auction now
```
Braqzen, line 92: `now` is not what this is doing. this is checking how much time has passed

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+        storage.endingPrice = endp;
+        storage.startTime = startt;
+        storage.endTime = endtt;
+        storage.ended = false;
+    }
+}
+
+fn price() -> u64 {
+    let price_difference = storage.startingPrice - storage.endingPrice;
+    let duration = storage.endTime - storage.startTime;
```
Braqzen, line 88:
```suggestion
    let auction_duration = storage.endTime - storage.startTime;
```

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+        storage.startingPrice = startp;
+        storage.endingPrice = endp;
+        storage.startTime = startt;
+        storage.endTime = endtt;
+        storage.ended = false;
+    }
+}
+
+fn price() -> u64 {
+    let price_difference = storage.startingPrice - storage.endingPrice;
```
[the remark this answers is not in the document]
Braqzen: Shouldn't this be checking the current highest bid rather than the reserve price?
It doesn't seem like endingPrice is being set anywhere. This is all too messy and confusing to follow. I'll need to go over this all again after adjustments have been made because I may have made a mistake in how I understand this

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    let now = height() - storage.startTime; //Current block height - start will tell us how far we are into the auction now
+    //Cap how far we are into the auction by the duration, so price doesnt go into negative or below endprice
+    let now = if now > duration {
+        duration
+    } else {
+        now
+    }
+
+    //price_shift * now tells us how much the price has reduced by now
+    return storage.startingPrice - (price_shift * now)
```
Braqzen, line 101: This won't compile... you're using a `return` keyword but you're missing `;` at the end. Remove the `return`

I'm going to assume the logic is correct

dutch-auction/tests/harness.rs
```diff
@@ -0,0 +1,25 @@
+use fuels::{prelude::*, tx::ContractId};
```
Braqzen, line 1: Hmmmmmm, submitting a draft as "good enough" with 0 tests?

Braqzen (changes requested):
This shouldn't be at the top level. Move it all to `sway-applications/auctions/dutch-acution`
Add a README.
Change "type of change" to "New Feature" instead of your custom "New Program" and add a space after "-" so that it looks like a bullet point.
Same with "Changes" for the bullet point and be more descriptive there. 
You've linked to the issue, no need to copy and paste the same content into the "additional" section. remove that entirely.
Create a logo and insert it into the readme too.

I've probably made some mistakes / not reviewed it to the same extent that I would otherwise because of how messy this is. It's difficult to keep everything in mind. I will need to make additional reviews after changes are made.

Oh, you *must* use `forc build` to check your work and also `forc fmt`.
I've forgotten to mention but also this needs to be added to the CI in `.github/workflows` and the script too

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    block::height,
+    chain::auth::msg_sender,
+    assert::assert,
+    revert::revert,
+    storage::StorageMap,
+    token::transfer_to_output,
+    context::{msg_amount, call_frames::msg_asset_id},
+    result::*,
+    identity::Identity,
+};
```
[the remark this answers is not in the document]
SwayStar123: I dont think the `transfer_to` would work here, it only transfers the native token of the current contract to the Identity, not the asset that is being bid with.

But other than that, yeah, I'll add Enum Errors and replace assert with require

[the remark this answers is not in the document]
Braqzen: Oh yeah, I remembered that and then I forgot and got mixed up with a WIP that I'm working on which replaces that function and allows you to transfer any asset instead of `contract_id()` and nothing else

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    fn set_beneficiary(new_beneficiary: Address) {
+        assert(get_sender() == storage.admin);
+
+        storage.beneficiary = new_beneficiary;
+    }
+
+    fn bid() {
+        //Since this is the dutch auction, first bid wins
+        assert(msg_amount() >= price());
+        assert(msg_asset_id() == ~ContractId::from(storage.asset_id));
```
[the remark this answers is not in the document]
Braqzen: When you purchase something it makes sense to first check that you have received the correct thing and then the amount.
For example, if you attempt to purchase something from you for an amount of "5" but I give you 5 carrots that doesn't make since in how you would pick up on that in reality. You wouldn't count the 5 carrots and say "yes, I have expected 5" and then say "wait a minute, these are carrots and not the currency that I expect".
You first notice the currency and then you count the amount

dutch-auction/src/main.sw
SwayStar123: Why the name change?

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    transfer_to_output(price(), ~ContractId::from(storage.asset_id), storage.beneficiary);
+    storage.ended = true;
+}
+
+impl MyContract for Contract {
+    fn get_price() -> u64 {
+        return price();
+    }
+
+    fn set_beneficiary(new_beneficiary: Address) {
```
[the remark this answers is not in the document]
SwayStar123: Auction can be restarted multiple times, so there's cases where different auctions would need different beneficiaries

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+        addr
+    } else {
+        revert(0);
+    };
+    ad
+}
+
+fn endauc() {
+
+}
```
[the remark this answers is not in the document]
SwayStar123: Removed it and added a public function for ending the auction

dutch-auction/src/main.sw
SwayStar123: Is that the right link?

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+
+        if msg_amount() > price() {
+            let return_amount = msg_amount() - price();
+            transfer_to_output(return_amount, ~ContractId::from(storage.asset_id), get_sender());
+        }
+
+        win();
+    }
+
+    fn setup_auction(startp: u64, endp: u64, startt: u64, endt: u64) {
```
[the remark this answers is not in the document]
SwayStar123: Not meant to be a constructor, added a constructor and made it so the setup_auction() function cant be called whilst there is an ongoing auction so this cant be abused anymore. Removed abbreviations

dutch-auction/src/main.sw
SwayStar123: If current block height is 100, start was 50, and end is 150. 
100 (height) - 50 (start) = 50 (now)
and duration is end - start so 150 -  50 = 100
we are 50 blocks into the auction, so now is correct?

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+    ended: bool,
+    //Enter the asset id of the currency you want the bid to be in
+    asset_id: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000,
+}
+
+fn win() {
+    // Do stuff on the win event
+    transfer_to_output(price(), ~ContractId::from(storage.asset_id), storage.beneficiary);
+    storage.ended = true;
+}
```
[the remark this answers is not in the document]
Braqzen: I don't understand. This is not a template, it's going to be a part of a complete application. They can change the logic in the contract itself if they need to. This is not a library that a user will be importing somewhere and overwritting the logic. Only abstract away logic if it is being repeated over and over or if it is a library (and thus may get called over and over)

dutch-auction/src/main.sw
adlerjohn: From Slack discussion: no `_` prefix. Implicit return is 👍 though

dutch-auction/src/main.sw
adlerjohn: GitHub hides the file by default unfortunately. Better link: https://github.com/FuelLabs/sway-applications/blob/174a1ae885a02e1b65ff6c4ea9eca0ae098f8373/escrow/src/main.sw#L191-L198

dutch-auction/src/main.sw
Braqzen: `price()` does not tell me anything about what the function is doing. Is it setting the price in storage, is it returning something etc. Given the change of use for `_` renaming to `get_price()` might cause a compilation error. Nevertheless, based on the logic of the function I think it should still be renamed to something like `calculate_price()` since that is what the function seems to be doing. It calculates it and returns it. The API call (`get_price()`) is fine but I would move that towards the bottom since it's just a getter

dutch-auction/src/main.sw
Braqzen: Yes, it seems like it is not being displayed correctly when opened so [here](https://github.com/FuelLabs/sway-applications/blob/174a1ae885a02e1b65ff6c4ea9eca0ae098f8373/escrow/src/main.sw#L191) is the direct line

dutch-auction/src/main.sw
Braqzen: That is a very un-intuitive way to look at this. I would have never thought of looking at it that way because of how the code is being used. I suggest you rename it accordingly.

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+
+    fn set_beneficiary(new_beneficiary: Address) {
+        assert(get_sender() == storage.admin);
+
+        storage.beneficiary = new_beneficiary;
+    }
+
+    fn bid() {
+        //Since this is the dutch auction, first bid wins
+        assert(msg_amount() >= price());
```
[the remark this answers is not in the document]
SwayStar123: Why would you not want people to send the exact amount required?

dutch-auction/src/main.sw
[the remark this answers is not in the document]
Braqzen: Refer to [this](https://github.com/FuelLabs/sway-applications/pull/54/commits/d7b156903fbbcdfed3985dc0080d2f766c8c8ea5) commit specifically. You may want to scan through the entire PR in general

dutch-auction/src/main.sw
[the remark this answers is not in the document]
Braqzen: I'm going by the assumption that when a new highest bidder is found then the variable keeping track of that is updated.
If the price returned can be equal to whatever the previous bidder has bid then they can just keep bidding the same amount to take the highest place. You'd want to stop that by making sure that the bid must be greater each time

dutch-auction/src/main.sw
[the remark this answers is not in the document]
SwayStar123: In a dutch auction the first person to bid wins. There is no current highest bid.

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+            let return_amount = msg_amount() - price();
+            transfer_to_output(return_amount, ~ContractId::from(storage.asset_id), get_sender());
+        }
+
+        win();
+    }
+
+    fn setup_auction(startp: u64, endp: u64, startt: u64, endt: u64) {
+        assert(get_sender() == admin);
+        assert(startp > endp);
```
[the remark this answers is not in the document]
SwayStar123: If the opening and reserve prices are equal then what is the point of using this dutch auction?

dutch-auction/src/main.sw
SwayStar123: Okay then it needs to be set the ended, as a winning bid will set `ended` to `true` so if you want to make a new auction the variable needs to be set back to false

dutch-auction/src/main.sw
```diff
@@ -0,0 +1,123 @@
...
+        storage.beneficiary = new_beneficiary;
+    }
+
+    fn bid() {
+        //Since this is the dutch auction, first bid wins
+        assert(msg_amount() >= price());
+        assert(msg_asset_id() == ~ContractId::from(storage.asset_id));
+        assert(!storage.ended);
+
+        storage.ended = true;
```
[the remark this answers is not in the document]
SwayStar123: price(), (now calculate_price()) Tells us how much needs to be sent, and if msg_value is greater than or equal to that price, its a valid/winning bid.

Dutch auctions end after the first bid.

dutch-auction/src/main.sw
SwayStar123: Renamed it to calculate_price

dutch-auction/src/main.sw
SwayStar123: Using unwrap and match now

dutch-auction/src/main.sw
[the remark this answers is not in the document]
SwayStar123: First bid wins in a dutch auction, there is no "new highest bidder" or "old highest bidder"

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+
+        /// If there is no auction going on currently the ended value will already be true so no need to check for that case
+        let mut auction = storage.auctions.get(auction_id);
+        auction.ended = true;
+        storage.auctions.insert(auction_id, auction);
+    }
+
+}
+
+fn win(auction_id: u64) {
```
Braqzen, line 156: add documentation stating that this is a placeholder that is meant to have additional logic added if the user wants to

Braqzen, line 156: I would rename this function. The name bothers me as it is not descriptive in what it is doing.
I'm unsure of what name to suggest but it should indicate that this is what sends the asset and possibly performs some custom logic

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+    ///  Do stuff on a winning bid, this function is called whenever a winning bid is recieved.
+
+    /// Add whatever logic you may want to execute on a win
+
+    //Currently just sends the bid amount to the beneficiary
+    let auction = storage.auctions.get(auction_id);
+    transfer_to_output(calculate_price(auction_id), storage.asset_id, auction.beneficiary);
```
Braqzen, line 163: I don't believe that you should be recalculating this. This _may_ cause a bug. Calculate the amount once in the calling function and pass it in

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+    
+    /// How much the price will go down by, throughout the auction
+    let price_delta = auction.opening_price - auction.reserve_price;
+    /// How long the auction will last
+    let auction_duration = auction.end_time - auction.start_time;
+    /// This is the amount the price will reduce by per block
+    let price_shift = price_delta / auction_duration;
+
+    /// Tells us how far we are into the auction (out of the auction_duration)
+    let now = height() - auction.start_time;
```
Braqzen, line 177:
```suggestion
    let time_passed = height() - auction.start_time; 
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+    /// How long the auction will last
+    let auction_duration = auction.end_time - auction.start_time;
+    /// This is the amount the price will reduce by per block
+    let price_shift = price_delta / auction_duration;
+
+    /// Tells us how far we are into the auction (out of the auction_duration)
+    let now = height() - auction.start_time; 
+
+    /// Cap how far we are into the auction by the auction_duration, so price doesnt go into negative or below endprice
+    let now = if now > auction_duration { auction_duration } else { now };
```
Braqzen, line 180:
```suggestion
    let time_passed = if time_passed > auction_duration { auction_duration } else { time_passed };
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+
+    /// price_shift * now tells us how much the price has reduced by now
+    return auction.opening_price - (price_shift * now);
```
Braqzen, line 183:
```suggestion
    auction.opening_price - (price_shift * time_passed)
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+fn get_sender() -> Address {
+    let a: Result<Identity, AuthError> = msg_sender();
+    let b = a.unwrap();
+
+    let addy = match b {
+        Identity::Address(addr) => addr,
+        _ => revert(0),
+    };
+    addy
+}
```
Braqzen, line 196:
```suggestion
fn unwrap_identity(sender: Result<Identity, AuthError>) -> Identity {
    sender.unwrap()
}
```

since we are dealing with Identity and not just addresses

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+    AuctionNotYetStarted: (),
+}
+
+impl DutchAuction for Contract {
+    fn constructor(admin: Address, asset: ContractId) {
+        storage.constructed = true;
+        storage.asset_id = asset;
+
+        storage.admin = admin;
+    }
```
Braqzen, line 70:
```suggestion
    fn constructor(admin: Identity) {
        require(!storage.initialized, Error::CannotReinitialize);
        storage.admin = admin;
        storage.initialized = true;
    }
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+        /// post-auction-win and so this contract template can be reused for multiple things
+        win(auction_id);
+    }
+
+    fn setup_auction(opening_price: u64, reserve_price: u64, start_time: u64, end_time: u64) -> u64 {
+        require(storage.constructed == true, Error::ContractNotConstructedYet);
+
+        require(get_sender() == storage.admin, Error::SenderNotAdmin);
+        require(opening_price > reserve_price, Error::EndPriceCannotBeLargerThanStartPrice);
+        require(end_time > height(), Error::AuctionCannotEndInThePast);
```
Braqzen, line 122:
```suggestion
        require(height() < end_time, Error::AuctionCannotEndInThePast);
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+        require(start_time >= height(), Error::AuctionCannotStartInThePast);
+        require(end_time > start_time, Error::AuctionCannotEndBeforeItStarts);
```
Braqzen, line 124:
```suggestion
        require(start_time < end_time, Error::AuctionCannotEndBeforeItStarts);
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+        win(auction_id);
+    }
+
+    fn setup_auction(opening_price: u64, reserve_price: u64, start_time: u64, end_time: u64) -> u64 {
+        require(storage.constructed == true, Error::ContractNotConstructedYet);
+
+        require(get_sender() == storage.admin, Error::SenderNotAdmin);
+        require(opening_price > reserve_price, Error::EndPriceCannotBeLargerThanStartPrice);
+        require(end_time > height(), Error::AuctionCannotEndInThePast);
+        require(start_time >= height(), Error::AuctionCannotStartInThePast);
```
Braqzen, line 123:
```suggestion
        require(height() <= start_time, Error::AuctionCannotStartInThePast);
```

Braqzen (changes requested):
Make sure to run `forc fmt` to format your files.
Add this to the `matrix` in `sway-applications/.github/workflows/ci.yml`
Add a README.

Much better than before.

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+
+    fn get_price(auction_id: u64) -> u64 {
+        calculate_price(auction_id)
+    }
+
+    fn set_beneficiary(new_beneficiary: Address, auction_id: u64) {
+        require(storage.constructed == true, Error::ContractNotConstructedYet);
+        require(get_sender() == storage.admin, Error::SenderNotAdmin);
+
+        let mut auction = storage.auctions.get(auction_id);
```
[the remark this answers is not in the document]
SwayStar123: Just removed the change beneficiary function and put it in the create_auction

auctions/dutch-auction/src/main.sw
SwayStar123: I'll use "blocks_into_auction" if thats fine instead, seems less confusing than either of those

auctions/dutch-auction/src/main.sw
SwayStar123: on_win?

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+
+        /// Disallows furthur bids
+        auction.ended = true;
+        storage.auctions.insert(auction_id, auction);
+
+        /// If someone sends more than the current price, refund the extra amount 
+        if msg_amount() > calculate_price(auction_id) {
+            let return_amount = msg_amount() - calculate_price(auction_id);
+            transfer_to_output(return_amount, storage.asset_id, get_sender());
+        }
```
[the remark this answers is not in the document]
SwayStar123: Since blocktimes could be rather fast on Fuel, I think alot of people would try to bid a little bit more than the price, hence the refunding

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    fn set_beneficiary(new_beneficiary: Address, auction_id: u64) {
+        require(storage.constructed == true, Error::ContractNotConstructedYet);
+        require(get_sender() == storage.admin, Error::SenderNotAdmin);
+
+        let mut auction = storage.auctions.get(auction_id);
+        auction.beneficiary = new_beneficiary;
+        storage.auctions.insert(auction_id, auction);
+    }
```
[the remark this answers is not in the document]
SwayStar123: Removed this function and made it so a beneficiary can only be set when creating a new auction

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,198 @@
...
+        auction.ended = true;
+        storage.auctions.insert(auction_id, auction);
+
+        /// If someone sends more than the current price, refunds the extra amount
+        if msg_amount() > price {
+            let return_amount = msg_amount() - price;
+            transfer_to_identity(return_amount, auction.asset_id, get_sender_identity());
+        }
+
+        on_win(auction_id, price);
```
Braqzen, line 82: since you already have the auction you can pass it in as an argument instead of querying storage again in the function

auctions/dutch-auction/src/abi.sw
```diff
@@ -0,0 +1,11 @@
+library abi;
+
+use std::{contract_id::ContractId, identity::Identity};
+
+abi DutchAuction {
```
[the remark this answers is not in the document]
Braqzen: I think this needs another function. I'm unsure. How will the frontend know which auction to show to users? Iterating over an infinite number of auctions in the mapping and checking that they have not ended (and currently exist) is inefficient and even then there's nothing to indicate which auction should be shown and to who.

Braqzen (changes requested):
Good work.

This needs another feature or 2, documentation (above each function using `///`, take a look at the other applications) and then I think you'll be ready to start testing!

auctions/dutch-auction/src/abi.sw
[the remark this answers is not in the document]
SwayStar123: After vectors are added we can have an "active auctions" vector in storage perhaps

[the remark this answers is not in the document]
SwayStar123: when vectors get added we can just have a vector of currently running auctions

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,198 @@
...
+    auction.opening_price - (price_shift * blocks_into_auction)
+}
+
+/// Helper function to avoid having to repeat this code
+fn get_sender_identity() -> Identity {
+    let a: Result<Identity, AuthError> = msg_sender();
+    let b = a.unwrap();
+
+    b
+}
```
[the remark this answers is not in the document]
SwayStar123: i dont think that function name is adequate enough. Maybe just `sender_identity()`?

auctions/dutch-auction/src/abi.sw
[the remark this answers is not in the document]
Braqzen: I'm doing something similar in the Escrow (using an array for now but planning on using vec) however I dislike it because the amount of juggling to get data in / out might be a pain. I'd prefer to use maps in some intelligent way to have instant (O(1)) access. We'll see what happens when vec is out

auctions/dutch-auction/src/main.sw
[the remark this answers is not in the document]
Braqzen: It's better than what I'm using in another project: `unwrap_identity()`. Nice.

auctions/dutch-auction/src/abi.sw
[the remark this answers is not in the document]
SwayStar123: You mean like move the ended auctions into a seperate storagemap maybe? I dont think anything like thats possible right now though.

Added the auction() function btw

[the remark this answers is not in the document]
Braqzen: My understanding is that you want to store the currently active auctions in a vector. This vector will increase in size as more auctions are added. You need to somehow know how to index into the vector to get the auction that you want because iterating over the vector is unacceptable. It will eventually get too large to iterate and you'll deadlock the contract. This means we need to know how to index into the vector via a mechanism similar to the storage map. Another problem is then removing the active auctions because that will reallocate a lot of memory which is something that we do not want to do therefore using a vector at all is a bad idea because even if we never remove from the vector eventually adding to it will make it reallocate so much memory that it will deadlock the contract again.
This leaves storage mappings as the only option but now you need to figure out how that can be used effectively. I have not thought that through yet...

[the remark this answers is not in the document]
SwayStar123: You could store a counter for active auctions and then just have a 0..n mapping of active auctions, and moving all the auctions down by one whenever one ends, but thats just an emulation of a vector, so sounds like a pointless endeavor. Other than that I have no idea what you could do with storage maps here that would allow you to track only the active auctions and remove the ones that end

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    /// Mapping an auction_id to its respective auction, allowing for multiple auctions to happen simultaneously
+    auctions: StorageMap<u64,
+    Auction>, /// Tracking how many auctions have been made till now
+    auction_count: u64,
+    /// Whether the constructor has been called
+    initialized: bool,
+}
+
+impl DutchAuction for Contract {
+    /// This function should be called right after being deployed, so the admin can be set
```
Braqzen, line 40: Document when every function will panic

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+        require(!storage.initialized, AuthorizationError::CannotReinitialize);
+        storage.admin = admin;
+        storage.initialized = true;
+    }
+
+    /// Returns the current price for the auction corresponding to the auction_id
+    fn price(auction_id: u64) -> u64 {
+        validate_id(auction_id);
+
+        calculate_price(auction_id)
```
Braqzen, line 51:
```suggestion
        validate_id(auction_id);
        calculate_price(auction_id)
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+                },
+                _ => false, 
+            }
+        },
+    }
+}
+
+/// Validates an auction_id to make sure it corresponds to an auction
+fn validate_id(id: u64) {
+    // If the given auction id is higher than the auction count, its an invalid auction_id
```
Braqzen, line 216:
```suggestion
    // If the id is greater than the auction count then it's invalid
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+        };
+
+        storage.auctions.insert(storage.auction_count, auction);
+
+        log(CreatedAuctionEvent {
+            id: storage.auction_count, auction, 
+        });
+    }
+
+    /// Cancels an auction so no one can bid on it. Needs to be called even after end_time of an auction if you want no one to bid anymore
```
Braqzen, line 116: Split onto 2 lines. People set different character lengths for how long a line should be and I think they are typically 80, 100 & 120 characters long.
I use 100 iirc. Set your editor to draw a line to indicate the character limit so that you do not write lines that are too long.

Braqzen: "Needs to be called even after end_time of an auction if you want no one to bid anymore" needs to be rephrased.

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    fn cancel_auction(auction_id: u64) {
+        require(storage.initialized, TechnicalError::ContractNotYetInitialized);
```
Braqzen, line 118: Remove require. Your id validation checks to see if the ID is in the valid range (i.e. also not 0)

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+        require(start_time < end_time, SetupError::AuctionCannotEndBeforeItStarts);
+
+        storage.auction_count = storage.auction_count + 1;
+
+        let auction = Auction {
+            opening_price, reserve_price, start_time, end_time, beneficiary, asset_id: asset,
+            ended: false,
+        };
+
+        storage.auctions.insert(storage.auction_count, auction);
```
Braqzen, line 109:
```suggestion
        let auction = Auction {
            opening_price, reserve_price, start_time, end_time, beneficiary, asset_id: asset,
            ended: false,
        };
        
        storage.auction_count = storage.auction_count + 1;
        storage.auctions.insert(storage.auction_count, auction);
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Returns the current price for the auction corresponding to the auction_id
+    fn price(auction_id: u64) -> u64 {
+        validate_id(auction_id);
+
+        calculate_price(auction_id)
+    }
+
+    /// Bids in the given auction_id, is a win if the amount and asset are correct
```
Braqzen, line 54: This really needs to be reworded

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    fn bid(auction_id: u64) {
+        // In a Dutch auction the first bid wins
+        require(storage.initialized, TechnicalError::ContractNotYetInitialized);
```
Braqzen, line 57: Remove require. Your id validation checks to see if the ID is in the valid range (i.e. also not 0)

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+
+        log(WinningBidEvent {
+            winner: sender_indentity(), id: auction_id, 
+        });
+    }
+
+    /// Creates a new auction
+    fn create_auction(opening_price: u64, reserve_price: u64, start_time: u64, end_time: u64, beneficiary: Identity, asset: ContractId) {
+        require(storage.initialized, TechnicalError::ContractNotYetInitialized);
+        require(eq_identity(storage.admin, sender_indentity()), AuthorizationError::SenderNotAdmin);
```
Braqzen, line 96: Remove admin check

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Cancels an auction so no one can bid on it. Needs to be called even after end_time of an auction if you want no one to bid anymore
+    fn cancel_auction(auction_id: u64) {
+        require(storage.initialized, TechnicalError::ContractNotYetInitialized);
+
+        validate_id(auction_id);
+
+        // Only the admin can end the auction (prematurely)
+        require(eq_identity(sender_indentity(), storage.admin), AuthorizationError::SenderNotAdmin);
```
Braqzen, line 123: Change to beneficiary being the only one who can cancel the auction

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+        on_win(auction, price);
+
+        log(WinningBidEvent {
+            winner: sender_indentity(), id: auction_id, 
+        });
+    }
+
+    /// Creates a new auction
+    fn create_auction(opening_price: u64, reserve_price: u64, start_time: u64, end_time: u64, beneficiary: Identity, asset: ContractId) {
+        require(storage.initialized, TechnicalError::ContractNotYetInitialized);
```
Braqzen, line 95: Remove initialized require, actually not needed anymore

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    /// Returns the auction for any given auction_id
+    fn auction(auction_id: u64) -> Auction {
+        validate_id(auction_id);
+        storage.auctions.get(auction_id)
+    }
+}
+
+/// This function is called whenever a winning bid is recieved.
+fn on_win(auction: Auction, winning_amount: u64) {
+    transfer_to_identity(winning_amount, auction.asset_id, auction.beneficiary);
```
Braqzen, line 146:
```suggestion
    // Add custom logic for winning the auction here
    transfer_to_identity(winning_amount, auction.asset_id, auction.beneficiary);
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,219 @@
...
+    initialized: bool,
+}
+
+impl DutchAuction for Contract {
+    /// This function should be called right after being deployed, so the admin can be set
+    fn constructor(admin: Identity) {
+        require(!storage.initialized, AuthorizationError::CannotReinitialize);
+        storage.admin = admin;
+        storage.initialized = true;
+    }
```
Braqzen, line 45: Constructor isn't needed anymore, nor any admin check. Delete anything to do with admin and initialization pls

auctions/dutch-auction/src/abi.sw
```diff
@@ -0,0 +1,16 @@
+library abi;
+
+dep data_structures;
+
+use std::{contract_id::ContractId, identity::Identity};
+
+use data_structures::Auction;
+
+abi DutchAuction {
```
Braqzen, line 9: can you add a function that allows only the beneficiary of the auction to edit the auction if no one has bid and the auction has not ended? This would allow them to extend the end time, change the asset to be sent and the beneficiary to be someone else. Perform any necessary error handling and any other changes that you think are valid

auctions/dutch-auction/src/abi.sw
```diff
@@ -0,0 +1,16 @@
...
+    fn price(auction_id: u64) -> u64;
+    fn bid(auction_id: u64);
+    fn create_auction(opening_price: u64, reserve_price: u64, start_time: u64, end_time: u64, beneficiary: Identity, asset: ContractId);
```
Braqzen, line 12: Despite the conversation somewhere else in the PR we still need a function that accepts an Identity as a parameter and returns their currently active auctions and completed auctions. This would be a struct with each field being an array for now -> vec in the future

auctions/dutch-auction/src/abi.sw
Braqzen: I think it will also be useful to expose the `auction_count` similar to the `price`. No error handling required

auctions/README.md
```diff
@@ -0,0 +1,7 @@
+# Overview
```
Braqzen, line 1: Can you add more information in general regarding the entire auction project? This is just 1 contract within it so an overview of the project in general is good

auctions/dutch-auction/README.md
```diff
@@ -0,0 +1,5 @@
+This is an implementation of a Dutch Auction in Sway. A dutch auction is one where price falls gradually from an initially high price, and the first person to bid wins the auction.
```
Braqzen, line 1: These are just a few lines without any structure. Add some headings or more information etc. A block of text is not going to catch the eye enough for someone to read it

Braqzen (changes requested):
Sorry about changing the requirements on you but I've thought of additional functionality for the project in general. Good work though.

auctions/dutch-auction/src/abi.sw
SwayStar123: All extending the end time would do is increase the price and slow down the price decline. Users can still bid on an auction past the end_time aslong as the auction isnt cancelled (ended). Only useful thing to change is the beneficiary and I suppose the asset_id if maybe the wrong one was set or something like that

Braqzen: Shouldn't the auction automatically end when the end time is passed and someone tries to do anything?

SwayStar123: thats a design choice which could be argued for either way. My thought process was that since the price is going down to a reserve price (the minimum the seller will sell for), the auctioneer would be fine with getting that sale even after the end_time, or else the reserve price would have been higher

SwayStar123: I just made it so people cant bid after end_time, although some people might want the sale, its still unexpected behaviour i suppose.

auctions/dutch-auction/src/abi.sw
SwayStar123: How would i do this with an array? Only vec seems plausible

Braqzen: My workaround is an array with 1 item.

```rust
struct SomeName {
    active: [u64; 1],
    completed: [u64; 1]
}
```

Then you replace the entire array when needed. It's not ideal for correct functionality so you could make it larger and add some boilerplate to handle indexing etc. but that's not really needed. As long as 1 works in the UI then it's fine imo until vec is out

SwayStar123: okay just added it with a array for now then, added some warnings saying its temporary till a vec is out

SwayStar123:
@Braqzen Should I work on testing now?

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+        calculate_price(auction_id)
+    }
+
+    /// Bids in the given auction, wins if the amount and type of asset are correct
+    /// Panics when -
+    ///     1. auction_id is 0 or higher than storage.auction_count
+    ///     2. auction has already ended
+    ///     3. current block height is lower than start_time, or higher than end_time
+    ///     4. asset_id of the msg is not the same as the asset_id set in the auction
+    ///     5. msg_amount is lower than the current going price in the auction
```
Braqzen, line 54:
```suggestion
    ///     5. The bid is less than the current price
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+impl DutchAuction for Contract {
+    /// Returns the current price for the auction corresponding to the auction_id
+    /// Panics when -
+    ///     1. auction_id is 0 or higher than storage.auction_count
+    fn price(auction_id: u64) -> u64 {
+        validate_id(auction_id);
+        calculate_price(auction_id)
+    }
+
+    /// Bids in the given auction, wins if the amount and type of asset are correct
```
Braqzen, line 48:
```suggestion
    /// Bids on the specified auction
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+
+        log(WinningBidEvent {
+            winner: sender_indentity(), id: auction_id, 
+        });
+    }
+
+    /// Creates a new auction
+    /// Panics when -
+    ///     1. reserve_price is greater than opening_price
+    ///     2. current block height is higher than end_time or start_time
```
Braqzen, line 99:
```suggestion
    ///     2. block height is greater than end_time or start_time
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+        storage.active_auctions_by_identity.insert(beneficiary, [storage.auction_count]);
+
+        log(CreatedAuctionEvent {
+            id: storage.auction_count, auction, 
+        });
+    }
+
+    /// Cancels an auction so no one can bid on it.
+    /// Panics when -
+    ///     1. auction_id is 0 or higher than storage.auction_count
```
Braqzen, line 124:
```suggestion
    ///     1. auction_id is 0 or greater than storage.auction_count
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+        storage.auctions.insert(storage.auction_count, auction);
+        /// WARNING: This needs to be changed to a push to a vec instead of just replacing the contents of the array
+        storage.active_auctions_by_identity.insert(beneficiary, [storage.auction_count]);
+
+        log(CreatedAuctionEvent {
+            id: storage.auction_count, auction, 
+        });
+    }
+
+    /// Cancels an auction so no one can bid on it.
```
Braqzen, line 122:
```suggestion
    /// Cancels an auction preventing any bids from being placed
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+            ended: false,
+        };
+
+        storage.auction_count = storage.auction_count + 1;
+        storage.auctions.insert(storage.auction_count, auction);
+        /// WARNING: This needs to be changed to a push to a vec instead of just replacing the contents of the array
+        storage.active_auctions_by_identity.insert(beneficiary, [storage.auction_count]);
+
+        log(CreatedAuctionEvent {
+            id: storage.auction_count, auction,
```
Braqzen, line 118:
```suggestion
            auction, id: storage.auction_count
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+            transfer_to_identity(return_amount, auction.asset_id, sender_indentity());
+        }
+
+        on_win(auction, price);
+
+        /// WARNING: This needs to be changed to a pop to a vec instead of just replacing the contents of the array
+        storage.active_auctions_by_identity.insert(auction.beneficiary, [0]);
+
+        log(WinningBidEvent {
+            winner: sender_indentity(), id: auction_id,
```
Braqzen, line 92:
```suggestion
            id: auction_id, winner: sender_indentity() 
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+    ///     1. auction_id is 0 or higher than storage.auction_count
+    ///     2. msg_sender is not the beneficiary of the auction
+    ///     3. auction has already ended
+    fn cancel_auction(auction_id: u64) {
+        validate_id(auction_id);
+
+        let mut auction = storage.auctions.get(auction_id);
+
+        // Only the beneficiary can end the auction (prematurely)
+        require(eq_identity(sender_indentity(), auction.beneficiary), UserError::SenderNotBeneficiary);
```
Braqzen, line 133: I think I may have said to switch this from the creator of the auction to the beneficiary. I've been unsure about that.
Having thought through a part of the user interface it makes sense to go back to the creator of the auction being the only one who can cancel it. I have some preliminary thoughts in the fundraiser spec

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+        storage.active_auctions_by_identity.insert(sender_indentity(), [0]);
+
+        log(CancelledAuctionEvent {
+            id: auction_id, 
+        });
+    }
+
+    /// Returns the auction for any given auction_id
+    /// Panics when -
+    ///     1. auction_id is 0 or higher than storage.auction_count
```
Braqzen, line 149:
```suggestion
    ///     1. auction_id is 0 or greater than storage.auction_count
```

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/// This function is called whenever a winning bid is recieved.
+fn on_win(auction: Auction, winning_amount: u64) {
+    // Add custom logic for winning the auction here
+    transfer_to_identity(winning_amount, auction.asset_id, auction.beneficiary);
+}
+
+/// Calculates the current price of a given auction
+fn calculate_price(auction_id: u64) -> u64 {
```
Braqzen, line 211: Can you move these private functions into `utils.sw` and import them?
Note that you cannot pass storage in directly into the function nor can you use it in a library so you must pass in the auction itself as an argument, likewise the auction count

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+use errors::{SetupError, TimeError, UserError};
+use events::{CancelledAuctionEvent, CreatedAuctionEvent, WinningBidEvent};
+
+storage {
+    /// Mapping an auction_id to its respective auction, allowing for multiple auctions to happen simultaneously
+    auctions: StorageMap<u64,
+    Auction>, /// Tracking how many auctions have been made till now
+    auction_count: u64,
+    /// Auction ids of the active auctions per beneficiary
+    active_auctions_by_identity: StorageMap<Identity,
```
Braqzen, line 34: I would rename this and split this into multiple mappings so that the UI can make a query for each ID. I'll show you the implementation when it hits the fundraiser draft

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,280 @@
...
+        /// WARNING: This needs to be changed to a pop to a vec instead of just replacing the contents of the array
+        storage.active_auctions_by_identity.insert(sender_indentity(), [0]);
+        /// WARNING: This needs to be changed to a push to a vec instead of just replacing the contents of the array
+        storage.active_auctions_by_identity.insert(new_beneficiary, [auction_id]);
+
+        storage.auctions.insert(auction_id, auction);
+    }
+
+    /// Returns the active auctions of which the identity is the beneficiary of
+    fn active_auctions_by_identity(identity_to_check: Identity) -> [u64;
```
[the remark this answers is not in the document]
Braqzen: we also need another function(s) (using mappings) to see which auctions a user has bid on. this should be private so only an ID should be passed in

Braqzen (changes requested):
Minor changes here and there.
There is a little bit more to do in terms of functionality so that the UI can display the info.
Aside from that you could start testing things if you want. I think most things aren't going to change at this point, only some internals might change a little in terms of queries + additional functionality

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     strategy:
       matrix:
-        project: ["escrow", "multisig-wallet"]
+        project: ["escrow", "multisig-wallet", "auctions/dutch-auction"]
```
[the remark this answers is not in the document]
Braqzen: oh, and this also needs to be added to /scripts/build.sh in alphabetical order too otherwise it won't be built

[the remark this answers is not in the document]
SwayStar123: never played with bash but copied the syntax there, i think i did it right

auctions/dutch-auction/src/main.sw
SwayStar123: for the data of a specific auction they can just use the auction function and enter the id in there?

Braqzen: Take a look at the fundraiser draft that I've mentioned

auctions/dutch-auction/src/main.sw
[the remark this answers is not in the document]
SwayStar123: This could be interpreted in two ways so i just implemented both

auctions/dutch-auction/src/main.sw
SwayStar123: I did, still dont see why someone cant just do `auction(id).author` instead

auctions/dutch-auction/src/main.sw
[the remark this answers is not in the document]
Braqzen: what?

auctions/dutch-auction/src/main.sw
Braqzen: what? don't use arrays. they are static. Look at the fundraiser again and use hashing / indexing to retrieve data

Braqzen: Have you read the fundraiser draft PR?

SwayStar123: Yes and after seeing it once again im more confused than before. Please specify which implementation exactly you are talking about.

> so that the UI can make a query for each ID.
i can only imagine you are referring to the auction id here, there is already a function to get the details of one specific auction

SwayStar123:
@Braqzen Any idea why the CI is failing? Looks like theres some kind of a version mismatch with forc maybe?

Braqzen:
> @Braqzen Any idea why the CI is failing? Looks like theres some kind of a version mismatch with forc maybe?

The error below is caused by a breaking change in `forc`.
`require` is now in the `revert` module.

SwayStar123:
@Braqzen can you review the documentation only for now? I cannot work on anything code related for around 2 days

Braqzen:
> @Braqzen can you review the documentation only for now? I cannot work on anything code related for around 2 days

What do you mean by documentation?

- Readme files
- Documentation on the functions

I don't consider the readme files to be documentation here. It's just a few sentences, not actual documentation. I can look over the code documentation if you want.

SwayStar123:
@Braqzen Yes I meant the code documentation, in the interface.sw file

auctions/dutch-auction/src/interface.sw
```diff
@@ -0,0 +1,164 @@
...
+    /// # Arguments
+    /// 
+    /// * `author` - The author of which to retrieve the auction of
+    /// * `index` - The index to retrieve
+    ///
+    /// # Reverts
+    ///
+    /// * When the index is greater than the number of auctions of the author
+    #[storage(read)]
+    fn auction_of_author(author: Identity, index: u64) -> u64;
```
Braqzen, line 64: This is too confusing. There are too many functions for `active` vs `inactive`. Just change the functions to have 1 source of truth which indicates if something is active or not rather than have 2 functions for each thing. I'll review the remaining docs after that is done.

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,220 @@
...
+use events::{CancelledAuctionEvent, ChangedAsset, CreatedAuctionEvent, WinningBidEvent};
+use utils::{calculate_price, eq_identity, sender_indentity, transfer_to_identity, validate_id, StorageMapVec};
+
+storage {
+    /// Mapping an auction_id to its respective auction, allowing for multiple auctions to happen simultaneously
+    auctions: StorageMap<u64, Auction> = StorageMap {},
+    /// Tracking how many auctions have been made till now
+    auction_count: u64 = 0,
+    /// Auction ids of the active auctions by author
+    active_auctions_of_author: StorageMapVec<Identity, u64> = StorageMapVec {},
```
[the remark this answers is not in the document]
SwayStar123: I had that convo with @simonr0204 , as maps and vecs cannot be in the same storage variable currently this project was being blocked, but this is not suitable as library as this is not something we want to maintain (especially in the standard library) Hence its in the utils.sw file

SwayStar123:
@Braqzen The getters for the auction ids/auctions has been simplified now using storagemapvec. Should be much less confusing

SwayStar123:
Okay so looks like Vecs are still not supported in the SDK so this is still blocked, cant test it

auctions/dutch-auction/src/events.sw
```diff
@@ -0,0 +1,24 @@
...
+
+use data_structures::Auction;
+
+pub struct CancelledAuctionEvent {
+    id: u64,
+}
+
+pub struct ChangedAsset {
+    new_asset: ContractId,
+    id: u64,
```
Braqzen, line 13:
```suggestion
        id: u64,
        new_asset: ContractId,
```

auctions/dutch-auction/src/interface.sw
```diff
@@ -0,0 +1,128 @@
...
+    /// # Arguments
+    ///
+    /// * `auction_id` - The id of the auction
+    ///
+    /// # Reverts
+    ///
+    /// * When the auction_id is 0 or greater than storage.auction_count
+    #[storage(read)]
+    fn auction(auction_id: u64) -> Auction;
+    /// Returns the auction ids of the auctions created by any author
```
Braqzen, line 27: Is the formatter removing the new line here?

auctions/dutch-auction/src/main.sw
```diff
@@ -0,0 +1,204 @@
...
+        end_time: u64,
+        beneficiary: Identity,
+        asset: ContractId,
+    ) {
+        require(reserve_price <= opening_price, SetupError::EndPriceCannotBeLargerThanStartPrice);
+        require(height() < end_time, SetupError::AuctionCannotEndInThePast);
+        require(height() <= start_time, SetupError::AuctionCannotStartInThePast);
+        require(start_time < end_time, SetupError::AuctionCannotEndBeforeItStarts);
+
+        let auction = Auction {
```
Braqzen, line 97: I'd consider adding a `new()` method.

auctions/dutch-auction/src/utils.sw
```diff
@@ -0,0 +1,94 @@
+library utils;
+
+dep data_structures;
+dep errors;
+
+use std::{
+    auth::{
+        AuthError,
+        msg_sender,
+    },
```
Braqzen, line 10:
```suggestion
    auth::msg_sender,
```

auctions/dutch-auction/src/utils.sw
```diff
@@ -0,0 +1,94 @@
...
+    block::height,
+    hash::sha256,
+    storage::{
+        get,
+        store,
```
Braqzen, line 15: `store` is unused

auctions/dutch-auction/src/utils.sw
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+        blocks_into_auction
+    };
+
+    // price_shift * blocks_into_auction tells us how much the price has reduced by now
+    auction.opening_price - (price_shift * blocks_into_auction)
+}
+
+/// Helper function to compare identities
+pub fn eq_identity(id_1: Identity, id_2: Identity) -> bool {
```
Braqzen, line 50: Identity implements `Eq` so this is no longer needed

auctions/dutch-auction/src/utils.sw
```diff
@@ -0,0 +1,94 @@
...
+                    contract_id_1 == contract_id_2
+                },
+                _ => false,
+            }
+        },
+    }
+}
+
+/// Helper function to avoid having to repeat this code
+pub fn sender_indentity() -> Identity {
```
Braqzen, line 72: Why not use `msg_sender().unwrap()` directly in the code? No need for additional bytecode to keep this function

auctions/dutch-auction/src/utils.sw
```diff
@@ -0,0 +1,94 @@
...
+    let sender: Result<Identity, AuthError> = msg_sender();
+    sender.unwrap()
+}
+
+/// Helper function to transfer assets to an identity
+pub fn transfer_to_identity(amount: u64, asset_id: ContractId, reciever: Identity) {
```
Braqzen, line 78: There is a `transfer` function which uses an `Identity` so this is not needed either

auctions/dutch-auction/tests/utils.rs
```diff
@@ -0,0 +1,27 @@
...
+    )
+    .await
+    .unwrap();
+
+    MyContractBuilder::new(id.to_string(), wallet.clone()).build()
+}
+
+pub async fn active_auctions_of_author(instance: &MyContract, author: Identity) -> Vec<u64> {
+    instance.active_auctions_of_author(author).call().await.unwrap().value
+}
```
Braqzen, line 27:
```suggestion
}

```

auctions/dutch-auction/tests/utils.rs
```diff
@@ -0,0 +1,27 @@
...
+        TxParameters::default(),
+        StorageConfiguration::with_storage_path(Some(
+            "./out/debug/dutch-auction-storage_slots.json"
+                .to_string(),
+        )),
+    )
+    .await
+    .unwrap();
+
+    MyContractBuilder::new(id.to_string(), wallet.clone()).build()
```
Braqzen, line 22: Even if `vec` wasn't an issue this code would need to be updated since the builder pattern is no longer exposed / exists

auctions/dutch-auction/src/main.sw
SwayStar123: I dont see a point, every field needs to be set either way

Braqzen: 1. There are 2 default values at the end that can be hardcoded and obscured
2. It makes the main file easier to read on a higher level since it's cleaner

SwayStar123:
If everything looks good ill move on to the specification and readme

.github/workflows/ci.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             "airdrop",
             "AMM",
             "auctions/english-auction",
+            "auctions/dutch-auction",
```
Braqzen, line 44:
```suggestion
            "auctions/dutch-auction",
            "auctions/english-auction",
```

auctions/README.md
```diff
@@ -0,0 +1,9 @@
...
```
Braqzen, line 1: I suggest looking at the other README files and adjusting this one to contain a higher level structure / scope / content of the auctions as this only has a few lines / bullet points. It's too barebones.

auctions/dutch-auction/project/.gitignore
Braqzen, line 1: I would check the repo to see if this is even needed or if a repo level gitignore already contains there

auctions/dutch-auction/project/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "dutch-auction"
+version = "0.0.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = { version = "0.38.1", features = ["fuel-core-lib"] }
```
Braqzen, line 9: I would bump all versions to be 0.41.0

auctions/dutch-auction/project/README.md
```diff
@@ -0,0 +1,9 @@
+# Overview
```
Braqzen, line 1: I would model the english auction docs for this file

auctions/dutch-auction/project/debug/dutch-auction-abi.json
Braqzen, line 1: The /out directory should be removed - although for some reason this is a /debug instead of /out/debug

auctions/dutch-auction/project/fuel-toolchain.toml
Braqzen, line 1: I would take a look at the recently merged bumped apps and use the same versions.
2023-04-12
0.37.0
0.17.9

auctions/dutch-auction/project/src/data_structures.sw
```diff
@@ -0,0 +1,22 @@
+library;
+
+pub struct Auction {
```
Braqzen, line 3: `impl` a `new` with default ended = false and winner is None

auctions/dutch-auction/project/src/errors.sw
```diff
@@ -0,0 +1,20 @@
...
+pub enum TimeError {
+    AuctionAlreadyEnded: (),
+    AuctionNotYetStarted: (),
+}
+
+pub enum UserError {
+    InvalidAuctionID: (),
+    BidTooLow: (),
+    WrongAssetSent: (),
+    SenderNotAuthor: (),
```
Braqzen, line 19:
```suggestion
    BidTooLow: (),
    InvalidAuctionID: (),
    SenderNotAuthor: (),
    WrongAssetSent: (),
```

auctions/dutch-auction/project/src/interface.sw
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// # Reverts
+    ///
+    /// * When auction_id is 0 or greater than storage.auction_count
+    /// * When the auction has already ended
+    /// * When the current block height is lower than start_time, or higher than end_time
+    /// * When the Incorrect asset is sent to the auction
+    /// * When the bid is less than the current price
+    #[storage(read, write)]
+    #[payable]
```
Braqzen, line 56:
```suggestion
    #[payable]
    #[storage(read, write)]
```

auctions/dutch-auction/project/src/interface.sw
Braqzen, line 1: Will need to take a closer look at the functions and docs later to see if they make sense

auctions/dutch-auction/project/src/main.sw
```diff
@@ -0,0 +1,232 @@
...
+    context::msg_amount,
+    token::transfer,
+};
+
+use ::interface::DutchAuction;
+use ::data_structures::Auction;
+use ::errors::{SetupError, TimeError, UserError};
+use ::events::{CancelledAuctionEvent, ChangedAsset, CreatedAuctionEvent, WinningBidEvent};
+use ::utils::{calculate_price, validate_id};
+use ::storagemapvec::StorageMapVec;
```
Braqzen, line 22:
```suggestion
use ::interface::DutchAuction;
use ::data_structures::Auction;
use ::errors::{SetupError, TimeError, UserError};
use ::events::{CancelledAuctionEvent, ChangedAsset, CreatedAuctionEvent, WinningBidEvent};
use ::utils::{calculate_price, validate_id};
use std::{
    auth::msg_sender,
    block::height,
    call_frames::msg_asset_id,
    context::msg_amount,
    token::transfer,
};
use ::storagemapvec::StorageMapVec;
```

auctions/dutch-auction/project/src/main.sw
```diff
@@ -0,0 +1,232 @@
...
+
+impl DutchAuction for Contract {
+    #[storage(read)]
+    fn price(auction_id: u64) -> u64 {
+        validate_id(auction_id, storage.auction_count);
+        calculate_price(storage.auctions.get(auction_id).unwrap())
+    }
+
+    #[storage(read, write)]
+    #[payable]
```
Braqzen, line 45:
```suggestion
    #[payable]
    #[storage(read, write)]
```

auctions/dutch-auction/project/src/interface.sw
```diff
@@ -0,0 +1,136 @@
+library;
+
+use ::data_structures::Auction;
+
+abi DutchAuction {
```
Braqzen, line 5: I would split this into `Info` like the other apps to separate the getters and follow the project structure in those tests

auctions/dutch-auction/project/src/utils.sw
Braqzen, line 1: I would check the imports here and in all the other files to make sure everything is used

auctions/dutch-auction/project/tests/functions/active_auctions_of_author.rs
Braqzen, line 1: mod success, mod revert

auctions/dutch-auction/project/tests/functions/active_auctions_of_author.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[tokio::test]
+    async fn can_get_active_auctions_of_author() {
+        let (instance, wallet) = get_contract_instance().await;
+
+        // Create 3 auctions and bid on the second one
+        create_auction(
+            &instance,
+            400,
+            100,
+            2, // Block height will be 1 at start, then 2 during sending of this tx
```
Braqzen, line 16: Use the provider instead of hardcoding block heights

auctions/dutch-auction/project/tests/functions/auction.rs
```diff
@@ -0,0 +1,214 @@
...
+            2, // Block height will be 1 at start, then 2 during sending of this tx
+            5,
+            Identity::Address(wallet.address().into()),
+            ContractId::zeroed(),
+        )
+        .await;
+
+        let auc = auction(&instance, 1).await;
+
+        assert_eq!(auc.asset_id, ContractId::zeroed());
```
Braqzen, line 28: You could look into the escrow which has `EscrowInfo` for events but something similar could be implemented here

auctions/dutch-auction/project/tests/functions/bid.rs
```diff
@@ -0,0 +1,262 @@
+pub mod passing {
+    use crate::utils::{auctions_won, bid, create_auction, get_contract_instance};
+    use fuels::{
+        signers::WalletUnlocked,
+        types::{AssetId, ContractId, Identity},
+    };
+
+    #[tokio::test]
+    async fn can_bid() {
```
Braqzen, line 9: Logs being emitted should be tested

auctions/dutch-auction/project/tests/functions/cancel_auction.rs
```diff
@@ -0,0 +1,128 @@
...
+            400,
+            100,
+            2, // Block height will be 1 at start, then 2 during sending of this tx
+            5,
+            Identity::Address(wallet.address().into()),
+            ContractId::zeroed(),
+        )
+        .await;
+
+        cancel_auction(&instance, 1).await;
```
Braqzen, line 22: This doesn't have assertions. How is this a test?

auctions/dutch-auction/project/tests/functions/create_auction.rs
```diff
@@ -0,0 +1,172 @@
...
+        );
+    }
+
+    #[tokio::test]
+    async fn auction_count_increases() {
+        let (instance, wallet) = get_contract_instance().await;
+
+        // retrieving active_auctions_of_author and auctions_of_author and auction_count before
+        let auction_count_before = auction_count(&instance).await;
+        let active_auctions_of_author_before =
```
Braqzen, line 37: This creation of an identity is being used repeatedly, I would wrap it into a function

auctions/dutch-auction/project/tests/functions/mod.rs
```diff
@@ -0,0 +1,11 @@
+pub mod active_auctions_of_author;
```
Braqzen, line 1: Do these need to be pub?

auctions/dutch-auction/project/tests/harness.rs
```diff
@@ -0,0 +1,3 @@
+pub mod utils;
+
+pub mod functions;
```
Braqzen, line 3:
```suggestion
pub mod functions;
pub mod utils;
```

auctions/dutch-auction/project/tests/utils.rs
```diff
@@ -0,0 +1,177 @@
+use fuels::{prelude::*, types::Identity};
```
Braqzen, line 1: I'd explicitly import

auctions/dutch-auction/project/tests/utils.rs
```diff
@@ -0,0 +1,177 @@
...
+        &wallet,
+        DeployConfiguration::default(),
+    )
+    .await
+    .unwrap();
+
+    (DutchAuction::new(id, wallet.clone()), wallet)
+}
+
+pub async fn active_auctions_of_author(instance: &DutchAuction, author: Identity) -> Vec<u64> {
```
Braqzen, line 48: I'd create a `setup.rs` for the setup and an interface dir which splits the content into core / info functionality. Take a look at the other apps for examples of how to do this

Braqzen (changes requested):

bitzoic:
Is this still being developed?

SwayStar123:
> Is this still being developed?

Not by me, someone can take over if they want

SwayStar123:
Closed as this is no longer being worked on

PR #79, opened 2022-06-08
feat: dao sc

### LINKED ISSUE #76: DAO Voting Program
### Why should this feature be added?

This program will serve as the backend for the DAO voting application #75 

### Specification

1. Owners of the program will crate and initialize a DAO voting contract
2. Users lock specified governance tokens in voting contract
3. Users can then vote on different proposals
4. Users can not vote on a proposal after a voting period is over
5. Users can get the current number of yes/no votes for a proposal
6. Users can get the result of a proposal after voting period ends (or maybe at any time)

### Additional context

_No response_

matt-user:
# Summary

## Type of change

- Implementation for Dao voting program.

## Changes

- Add Constructor
- Add deposit and balance functions
- Add tests

# Related Issues

Closes #76

# Additional context


bitzoic:
This needs to be added to CI

dao-voting/src/abi.sw
```diff
@@ -0,0 +1,21 @@
...
+dep data_structures;
+
+use std::{contract_id::ContractId, identity::Identity};
+
+use data_structures::{CallData, Proposal};
+
+abi DaoVoting {
+    #[storage(read, write)] fn constructor(gov_token: ContractId) -> bool;
+    #[storage(read, write)] fn deposit() -> bool;
+    #[storage(read)] fn get_balance() -> u64;
```
bitzoic, line 12: I believe rust convention is not to have the term 'get' in any getters.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,272 @@
...
+    u64>,
+    // The amount of votes a user has used on a proposal
+    votes: StorageMap<(Identity, u64),
+    u64>, state: u64, 
+}
+
+impl DaoVoting for Contract {
+    /// Initialize the dao with the governance token, voting parameters, and the proposal.
+    ///
+    /// # Panics
```
bitzoic, line 42: Take a look at @Braqzen's fundraiser pull request and try to match this format. The documentation should include arguments

dao-voting/src/main.sw
```diff
@@ -0,0 +1,272 @@
...
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The voting period is 0
+    /// - The approval percentage is 0
+    #[storage(read, write)]
+    fn add_proposal(voting_period: u64, approval_percentage: u64, proposal_data: CallData) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(voting_period > 0, Error::PeriodCannotBeZero);
+        require(approval_percentage > 0, Error::ApprovalPercentageCannotBeZero);
```
bitzoic, line 71: Does the contract handle cases where someone enters an `approval_percentage` greater than 100? Is this dangerous?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,272 @@
...
+        let mut proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height >= height(), Error::ProposalExpired);
+
+        if (is_yes_vote) {
+            proposal.yes_votes = proposal.yes_votes + vote_amount;
+        } else {
+            proposal.no_votes = proposal.no_votes + vote_amount;
+        };
+
+        storage.balances.insert(sender, sender_balance - vote_amount);
```
bitzoic, line 170: If there are two proposals active at the same time shouldn't the user have access to all their tokens for each proposal individually? If we remove it from their balance their vote will be decreased in the other proposal. Not sure what the convention in solidity DAO voting is

dao-voting/src/main.sw
```diff
@@ -0,0 +1,272 @@
...
+    /// - The proposal has not met the necessary approval percentage
+    #[storage(read, write)]
+    fn execute(proposal_id: u64) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
+
+        let proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height < height(), Error::ProposalActive);
+
+        let approval_percentage = proposal.yes_votes * 100 / (proposal.yes_votes + proposal.no_votes);
```
bitzoic, line 197: I feel like this hardcoded value is dangerous

dao-voting/src/main.sw
```diff
@@ -0,0 +1,272 @@
...
+    /// # Panics
+    ///
+    /// This functions will panic when:
+    /// - The constructor has not been called to initalize
+    /// - The user tries to withdraw more than their balance
+    #[storage(read, write)]
+    fn withdraw(amount: u64) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+
+        let result: Result<Identity, AuthError> = msg_sender();
```
bitzoic, line 121: While not critical, it might be nice to have a utils file with a function that returns the sender as this is repeated in a lot of places

dao-voting/src/main.sw
matt-user: good catch!  It doesn't so if the approval percentage was greater than 100 it would not be able to get approved.

dao-voting/src/main.sw
matt-user: I'm not sure I'll do more research.  My thought behind not allowing a user to use their balance for all proposals was to make them prioritize which proposals they wanted to see included.

dao-voting/src/main.sw
matt-user: hmm not sure.  I think mathematically the expression is guaranteed to be between 0 and 100.  Why would the value be dangerous?

matt-user: Yeah this could cause an overflow.  There's probably a mathematical trick with ints to avoid such scenarios.  I'll look into it.

matt-user: I couldn't find a way to avoid a potential overflow.  I think for now this will be ok, but I'll file an issue and add a comment to the code

dao-voting/src/main.sw
matt-user: I looked into how MakerDAO governance works, and it works how you describe above. (You can use the same MKR token to vote in any number of polls/proposals).  I'll create an issue for this and tackle it in a separate PR, unless anyone thinks its necessary to include it in this one.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Return the amount of governance tokens in this contract
+    #[storage(read)]fn balance() -> u64 {
+        this_balance(storage.gov_token)
+    }
+
+    /// Return the amount of governance tokens a user has in this contract
+    ///
+    /// # Parameters
+    /// user - user of which to get internal balance of governance tokens
```
Braqzen, line 265:
```suggestion
    /// # Parameters
    ///
    /// - `user` - user of which to get internal balance of governance tokens
```

I would probably reword this with a bit more careful thought

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// proposal_id - proposal of which to get votes spent by user
+    #[storage(read)]fn user_votes(user: Identity, proposal_id: u64) -> u64 {
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
+        storage.votes.get((user, proposal_id))
+    }
+
+    /// Return proposal data for a given id
+    ///
+    /// # Parameters
+    /// proposal_id - id of proposal to get
```
Braqzen, line 283:
```suggestion
    /// # Parameters
    ///
    /// - `proposal_id` - id of proposal to get
```

I would reword this too

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called ot initialize
+    /// - The given id is out of range
+    #[storage(read)]fn proposal(proposal_id: u64) -> Proposal {
+        require(storage.state == 1, Error::NotInitialized);
```
Braqzen, line 291: Why do you need to revert here if not initialized? I would remove

Braqzen: also, you've not handled the id 0 case

Braqzen: I think this should also be <= instead of <

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// user - user of which to get internal balance of governance tokens
+    #[storage(read)]fn user_balance(user: Identity) -> u64 {
+        storage.balances.get(user)
+    }
+
+    /// Return the amount of votes a user has used on a proposal
+    ///
+    /// # Parameters
+    /// user - user of which to get votes spent on a proposal
+    /// proposal_id - proposal of which to get votes spent by user
```
Braqzen, line 274:
```suggestion
    /// # Parameters
    ///
    /// - `user` - user of which to get votes spent on a proposal
    /// - `proposal_id` - proposal of which to get votes spent by user
```

again would rephrase

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    #[storage(read)]fn user_votes(user: Identity, proposal_id: u64) -> u64 {
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
```
Braqzen, line 276: Made comments on validation below

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        true
+    }
+
+    /// Vote on a given proposal
+    ///
+    /// # Parameters
+    ///
+    /// proposal_id - proposal to vote on
+    /// vote_amount - amount of votes to use on proposal
+    /// is_yes_vote - determines if you vote yes or no on the proposal
```
Braqzen, line 155:
```suggestion
    /// - `proposal_id` - proposal to vote on
    /// - `vote_amount` - amount of votes to use on proposal
    /// - `is_yes_vote` - determines if you vote yes or no on the proposal
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The proposal id is out of range
+    /// - The vote amount is 0
+    /// - The vote amount is greater than the users deposited balance
+    /// - The given proposal is expired
+    #[storage(read, write)]fn vote(proposal_id: u64, vote_amount: u64, is_yes_vote: bool) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
```
Braqzen, line 167: made comments on error handling below

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        let sender: Identity = sender_identity();
+        let sender_balance = storage.balances.get(sender);
+
+        require(sender_balance >= vote_amount, Error::NotEnoughAssets);
+
+        let mut proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height >= height(), Error::ProposalExpired);
+
+        if (is_yes_vote) {
+            proposal.yes_votes = proposal.yes_votes + vote_amount;
```
Braqzen, line 179: You can use +=

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        };
+
+        storage.balances.insert(sender, sender_balance - vote_amount);
+
+        let votes = storage.votes.get((sender, proposal_id));
+        storage.votes.insert((sender, proposal_id), votes + vote_amount);
+
+        storage.proposals.insert(proposal_id, proposal);
+
+        true
```
Braqzen, line 191: emit event

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// - The given proposal is expired
+    #[storage(read, write)]fn vote(proposal_id: u64, vote_amount: u64, is_yes_vote: bool) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
+        require(vote_amount > 0, Error::VoteAmountCannotBeZero);
+
+        let sender: Identity = sender_identity();
+        let sender_balance = storage.balances.get(sender);
+
+        require(sender_balance >= vote_amount, Error::NotEnoughAssets);
```
Braqzen, line 173:
```suggestion
        require(vote_amount <= sender_balance, Error::NotEnoughAssets);
```

So the voting is equal to the deposit? hm

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+        let mut proposal = storage.proposals.get(proposal_id);
```
Braqzen, line 175: I would change the logic to first fetch the proposal and check the deadline and then the amount for voting

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        require(proposal.end_height >= height(), Error::ProposalExpired);
```
Braqzen, line 176:
```suggestion
        require(height() <= proposal.end_height, Error::ProposalExpired);
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// # Parameters
+    /// proposal_id - proposal to turn user votes back into governance tokens
+    ///
+    /// # Panics
+    ///
+    /// This function will panic when:
+    /// - The constructor has not ben called to initialize
+    /// - The proposal id is invalid
+    /// - The proposal is still active
+    #[storage(read, write)]fn convert_votes_to_tokens(proposal_id: u64) {
```
Braqzen, line 241: This definitely needs a better name

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
```
Braqzen, line 243: error handling notes below

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        // Users can now convert their votes back into tokens
+
+        true
+    }
+
+    /// Unlock tokens used to vote on proposals to allow the user to withdraw
+    /// If the user had not voted in the given expired proposal, nothing happens
+    ///
+    /// # Parameters
+    /// proposal_id - proposal to turn user votes back into governance tokens
```
Braqzen, line 233:
```suggestion
    /// # Parameters
    ///
    /// - `proposal_id` - proposal to turn user votes back into governance tokens
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+            call rA rB rC rD;
+        }
+
+        // Users can now convert their votes back into tokens
+
+        true
+    }
+
+    /// Unlock tokens used to vote on proposals to allow the user to withdraw
+    /// If the user had not voted in the given expired proposal, nothing happens
```
Braqzen, line 230: This will need to be rephrased

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// proposal_id - proposal to turn user votes back into governance tokens
+    ///
+    /// # Panics
+    ///
+    /// This function will panic when:
+    /// - The constructor has not ben called to initialize
+    /// - The proposal id is invalid
+    /// - The proposal is still active
+    #[storage(read, write)]fn convert_votes_to_tokens(proposal_id: u64) {
+        require(storage.state == 1, Error::NotInitialized);
```
Braqzen, line 242: no need for state check here

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        let proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height < height(), Error::ProposalActive);
+
+        let sender: Identity = sender_identity();
+        let votes = storage.votes.get((sender, proposal_id));
+
+        storage.votes.insert((sender, proposal_id), 0);
+
+        let balance = storage.balances.get(sender);
+        storage.balances.insert(sender, balance + votes);
```
Braqzen, line 254: emit event

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        storage.proposals.insert(proposal_id, proposal);
+
+        true
+    }
+
+    /// Execute a given proposal
+    ///
+    /// # Parameters
+    ///
+    /// proposal_id - proposal to execute
```
Braqzen, line 198:
```suggestion
    /// - `proposal_id` - proposal to execute
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// # Panics
+    ///
+    /// This function will panic when:
+    /// - The construct has not been called to initialize
+    /// - The proposal id is out of range
+    /// - The proposal has not expired
+    /// - The proposal has not met the necessary approval percentage
+    #[storage(read, write)]fn execute(proposal_id: u64) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
```
Braqzen, line 209: error handling notes below

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+        let proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height < height(), Error::ProposalActive);
+
+        // TODO figure out how to prevent approval percentage from overflowing
+        // When close to the u64 max
+        // https://github.com/FuelLabs/sway-applications/issues/106
+        let approval_percentage = proposal.yes_votes * 100 / (proposal.yes_votes + proposal.no_votes);
+        require(approval_percentage >= proposal.approval_percentage, Error::ApprovalPercentageNotMet);
```
Braqzen, line 218:
```suggestion
        require(proposal.approval_percentage <= approval_percentage, Error::ApprovalPercentageNotMet);
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+        asm(rA: proposal.call_data.memory_address, rB: proposal.call_data.num_coins_to_forward, rC: proposal.call_data.asset_id_of_coins_to_forward, rD: proposal.call_data.amount_of_gas_to_forward) {
+            call rA rB rC rD;
+        }
+
+        // Users can now convert their votes back into tokens
```
Braqzen, line 224: SHould they be able to though?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+        true
```
Braqzen, line 226: emit event

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        require(storage.state == 1, Error::NotInitialized);
+        require(proposal_id < storage.proposal_count, Error::InvalidId);
+
+        let proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height < height(), Error::ProposalActive);
+
+        // TODO figure out how to prevent approval percentage from overflowing
+        // When close to the u64 max
+        // https://github.com/FuelLabs/sway-applications/issues/106
+        let approval_percentage = proposal.yes_votes * 100 / (proposal.yes_votes + proposal.no_votes);
```
Braqzen, line 217: yes and no votes can both be 0 here. should this be handled?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+        require(approval_percentage >= proposal.approval_percentage, Error::ApprovalPercentageNotMet);
+
+        asm(rA: proposal.call_data.memory_address, rB: proposal.call_data.num_coins_to_forward, rC: proposal.call_data.asset_id_of_coins_to_forward, rD: proposal.call_data.amount_of_gas_to_forward) {
```
Braqzen, line 220: you can actually name your registers to be meaningful instead of using rA rB etc

dao-voting/src/main.sw
Braqzen: Wait, should you be adding these or subtracting?

Braqzen, line 217: You can rearrange the calculation

(proposal.approval_percentage * (proposal.yes_votes + proposal.no_votes) ) / 100 <= proposal.yes_votes.

This doesn't solve the issue (because of the multiplication of yes + no) but I think it will make it less likely.

You should be able to come up with a different formula which uses the ratio of yes:no so that you don't even need to do the multiplication

Braqzen (changes requested):
First pass. Not looking at tests for the first iteration

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+    identity::Identity,
+    result::*,
+    revert::revert,
+    storage::StorageMap,
+    token::{transfer_to_output, force_transfer_to_contract},
+};
+
+use abi::DaoVoting;
+use data_structures::{CallData, Proposal};
+use errors::Error;
```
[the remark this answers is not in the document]
adlerjohn: To add on to this: events should be emitted for every state change. This allows off-chain clients to reconstruct the changing state of the contract from logs alone.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+    u64>,
+    // The amount of votes a user has used on a proposal
+    votes: StorageMap<(Identity, u64),
+    u64>, state: u64, 
+}
+
+impl DaoVoting for Contract {
+    /// Initialize the dao with the governance token, voting parameters, and the proposal.
+    ///
+    /// # Parameters
```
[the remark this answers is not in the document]
adlerjohn: In this case we have no reason to not follow Rust, so follow Rust it is.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+    /// The function will panic when:
+    /// - The constructor is called more than once
+    /// - The voting period is not greater than 0
+    /// - The approval percentage is not greater than 0
+    #[storage(read, write)]
+    fn constructor(gov_token: ContractId) -> bool {
+        require(storage.state == 0, Error::CannotReinitialize);
+
+        storage.gov_token = gov_token;
+        storage.proposal_count = 0;
```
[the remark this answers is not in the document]
adlerjohn: The compiler can optimize this away, so it shouldn't hurt to leave it in from a gas cost perspective. We should avoid gas optimizations at this state and trust the compiler to do everything properly eventually.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The voting period is 0
+    /// - The approval percentage is 0
+    /// - The approval percentage is above 100
+    #[storage(read, write)]fn add_proposal(voting_period: u64, approval_percentage: u64, proposal_data: CallData) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
```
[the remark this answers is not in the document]
adlerjohn: That may be true, but it's easier to not analyze that an just prevent all interaction until initialized.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+}
+
+impl DaoVoting for Contract {
+    /// Initialize the dao with the governance token, voting parameters, and the proposal.
+    ///
+    /// # Parameters
+    ///
+    /// gov_token - contract id of the token to use to vote on governance proposals
+    ///
+    /// # Panics
```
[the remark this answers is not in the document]
adlerjohn: +1 to above. Unlike Rust panics, a revert in the FuelVM doesn't crash the program irrecoverably; it's actually desirable behavior. Using a different name makes this clearer.

dao-voting/src/main.sw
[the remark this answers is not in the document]
Braqzen: Fair point for optimization however for code quality I will argue that this is a bad practice since it is not needed and it adds additional cognitive load to process the info as to why it is there.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The user deposits an asset that is not the specified governance token.
+    /// - The user does not deposit and assets
+    #[storage(read, write)]fn deposit() -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(storage.gov_token == msg_asset_id(), Error::NotGovernanceToken);
+        require(msg_amount() > 0, Error::NoAssetsSent);
+
+        let sender: Identity = sender_identity();
```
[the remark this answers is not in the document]
Braqzen: Note that if you update to the latest release you can now use `msg_sender().unwrap()` directly

dao-voting/src/main.sw
matt-user: If both yes and no votes are 0 we should probably handle it here.  Currently if both yes and no votes are both 0 this line will revert with an arithmetic error I'm guessing.  However,  if both yes and no votes are 0 there is no possible way for the approval percentage to be met (it is an error to add a proposal with 0 approval_percentage), which would revert the function anyways.  That being said do you think it is still worth it to handle this edge case?

Regarding your second comment, I believe we should be adding these.  The approval_percentage for a proposal is the number of yes votes out of the total number of votes.  It might be a good idea for me to document this line more clearly.    If this line was in a regular backend application I would use floats and just do yes_votes / (total_votes) to get the approval percentage.  Let me know if you need more clarification.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+    assert::require,
+    block::height,
+    chain::auth::{AuthError, msg_sender},
+    context::{call_frames::msg_asset_id, msg_amount, this_balance},
+    contract_id::ContractId,
+    identity::Identity,
+    result::*,
+    revert::revert,
+    storage::StorageMap,
+    token::{transfer_to_output, force_transfer_to_contract},
```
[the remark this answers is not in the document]
matt-user: Awesome!  That means I can get rid of the match statement right?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+};
+
+use abi::DaoVoting;
+use data_structures::{CallData, Proposal};
+use errors::Error;
+
+storage {
+    gov_token: ContractId,
+    voting_period: u64,
+    approval_percentage: u64,
```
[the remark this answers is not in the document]
matt-user: I agree, acceptance_threshold is definitely a better name.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,311 @@
...
+    proposals: StorageMap<u64,
+    Proposal>, proposal_count: u64,
+    // The amount of governance tokens a user has deposited
```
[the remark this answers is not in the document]
matt-user: Interesting, I did not know there was a distinction between `///` and `//`

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The voting period is 0
+    /// - The approval percentage is 0
+    /// - The approval percentage is above 100
+    #[storage(read, write)]fn add_proposal(voting_period: u64, approval_percentage: u64, proposal_data: CallData) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+        require(voting_period > 0, Error::PeriodCannotBeZero);
```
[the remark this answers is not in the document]
matt-user: Is there a particular reason to order the expression this way?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+
+    /// Update the user balance to indicate they have deposited governance tokens.
+    /// A successful deposit unlocks voting functionality.
+    ///
+    /// # Panics
+    ///
+    /// The function will panic when:
+    /// - The constructor has not been called to initialize
+    /// - The user deposits an asset that is not the specified governance token.
+    /// - The user does not deposit and assets
```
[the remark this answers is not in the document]
matt-user: Yes that's what I mean.  The "and" should be "any" I think.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,295 @@
...
+    /// This functions will panic when:
+    /// - The constructor has not been called to initalize
+    /// - The user tries to withdraw more than their balance
+    #[storage(read, write)]fn withdraw(amount: u64) -> bool {
+        require(storage.state == 1, Error::NotInitialized);
+
+        let sender: Identity = sender_identity();
+
+        let prev_balance = storage.balances.get(sender);
+        require(prev_balance >= amount, Error::NotEnoughAssets);
```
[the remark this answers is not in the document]
matt-user: Interesting, do you know which way is more popular in solidity apps?

dao-voting/src/main.sw
matt-user: What do you mean the id 0 case?  Why can't a proposal_id be 0.

matt-user: Currently I'm indexing the proposals in the storage map how I would with a vector/array, starting at the 0 index.

dao-voting/src/main.sw
matt-user: Yes, a user's voting power is directly proportional to the balance they have deposited in the contract.  There are other voting schemes, such as quadratic, but I thought for this first implementation this method would work fine.  I have also made an issue to allow voters to use one governance token to vote on any number of approvals (just not the same proposal).  This is what Maker DAO allows for their proposals.

dao-voting/src/main.sw
matt-user: Why should they not be able to?  After a proposal expires and can no longer be voted on they should be able to "unlock" their votes if they want to withdraw their tokens.

Braqzen: It could be a disincentive to prevent the spam of crappy proposals. Winning proposals are unlocked, crappy ones get burned. Design decision... I don't think it matters in the first iteration. More thought can be given to expanding the logic or creating modules which allow you to switch in other logic modules later on

dao-voting/src/main.sw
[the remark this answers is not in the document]
Braqzen: Consistency with other apps. I also find syntax like `x >= y` to be strange rather than `y <= x`. This is how I've seen math notation be taught and also coding so I am assuming that it is far more common than the alternative (which apparently people use).

dao-voting/src/main.sw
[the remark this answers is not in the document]
Braqzen: I believe people just revert if they enter the incorrect amount

dao-voting/src/main.sw
Braqzen: Expanding to use additional voting mechanisms as modules would be great in the future

dao-voting/src/main.sw
Braqzen: I guess reverting automatically when they are both 0 is fine. Whenever I see a division by 0 in a formula I wonder about the black holes that we are spawning somewhere and whether the universe will implode.

dao-voting/src/main.sw
Braqzen: If an incorrect query is made then 0 could be used as a sentinel to indicate that there is no data there. It is clearer than returning actual data and then having to notice that the data that you are expecting is not the one that you have received. 
It's fine if you want to continue to use 0 though

dao-voting/src/main.sw
matt-user: Yes there is still work to do on how a proposal gets added.  Iirc the Maker DAO system for adding proposals is basically: only specific accounts can add proposals, and that's how they prevent spam.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+    /// Unlock governance tokens which have been locked by users who have voted on a proposal
+    /// If the user had not voted on the given, expired proposal then nothing happens
+    ///
+    /// # Arguments
+    ///
+    /// - `proposal_id` - proposal to turn user votes back into governance tokens
+    ///
+    /// # Reverts
+    ///
+    /// * When the constructor has not ben called to initialize
```
Braqzen, line 251: Man this constructor is really giving you a hard time. If only the constructor wasn't mean enough to point it out each time :)

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+
+        let sender: Identity = msg_sender().unwrap();
+        let votes = storage.votes.get((sender, proposal_id));
+
+        storage.votes.insert((sender, proposal_id), 0);
+
+        storage.balances.insert(sender, storage.balances.get(sender) + votes);
+
+        log(UnlockVotesEvent {
+            id: proposal_id, vote_amount: votes,
```
Braqzen, line 268: Would add who unlocked

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+    /// # Arguments
+    ///
+    /// - `proposal_id` - Identifier used to specify a proposal (0 <= proposal_id < proposal_count)
+    ///
+    /// # Reverts
+    ///
+    /// * When the constructor has not been called ot initialize
+    /// * When the given proposal id is out of range
+    #[storage(read)]fn proposal(proposal_id: u64) -> ProposalInfo {
+        require(storage.state == State::Initialized, InitializationError::ContractNotInitialized);
```
Braqzen, line 308: The state check is unnecessary since the next check completely validates the ID

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+    #[storage(read)]fn user_votes(user: Identity, proposal_id: u64) -> u64 {
+        require(proposal_id < storage.proposal_count, UserError::InvalidId);
+        storage.votes.get((user, proposal_id))
+    }
+
+    /// Return proposal data for a given id
+    ///
+    /// # Arguments
+    ///
+    /// - `proposal_id` - Identifier used to specify a proposal (0 <= proposal_id < proposal_count)
```
Braqzen, line 301: You are using different descriptions for the same param. Pick 1 and use it everywhere consistently

dao-voting/tests/artifacts/gov_token/src/main.sw
```diff
@@ -0,0 +1,14 @@
+contract;
+
+use std::{address::Address, context::call_frames::contract_id, token::mint_to_address};
+
+abi GovToken {
+    fn mint_and_send_to_address(amount: u64, recipient: Address) -> bool;
```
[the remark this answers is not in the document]
Braqzen: So, only addresses?

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,974 @@
+#![allow(unused_variables)]
```
Braqzen, line 1: These should not be left in at the top. This hides whatever mistakes you have made. Moreover, this should be split up into modules with clean imports.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,974 @@
...
+#![allow(unused_imports)]
+
+use std::ops::Add;
```
Braqzen, line 4: You can import directly from the library too huh. That's strange considering that this is not an input or output... How is this even working?

Braqzen: Why is this needed?

dao-voting/src/abi.sw
```diff
@@ -0,0 +1,21 @@
+library abi;
+
+dep data_structures;
+
+use std::{contract_id::ContractId, identity::Identity};
+
+use data_structures::{Proposal, ProposalInfo};
+
+abi DaoVoting {
+    #[storage(read, write)] fn constructor(gov_token: ContractId);
```
Braqzen, line 10: Can these be in the same order as defined in the contract? It makes it easier to follow when looking at what comes next in the main file

Braqzen (changes requested):
More comments, later!

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+    /// Used to check the validity of a proposal id
+    proposal_count: u64,
+    /// The amount of votes a user has used on a proposal
+    votes: StorageMap<(Identity,
+    u64), u64>, /// The initilization state of the contract.  Defaults to NotInitialized.
+    state: State,
+}
+
+impl DaoVoting for Contract {
+    /// Initialize the dao with the governance token, voting parameters, and the proposal.
```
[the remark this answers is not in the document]
matt-user: Definitely a future enhancement, looking forward to abstracting voting logic as well :)

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+            revert(42);
+        }
+
+        // TODO figure out how to prevent approval percentage from overflowing
+        // When close to the u64 max
+        // https://github.com/FuelLabs/sway-applications/issues/106
+        let acceptance_percentage = proposal.yes_votes * 100 / (proposal.yes_votes + proposal.no_votes);
+        require(proposal.acceptance_percentage <= acceptance_percentage, ProposalError::ApprovalPercentageNotMet);
+
+        asm(memory_address: proposal.call_data.memory_address, num_coins_to_forward: proposal.call_data.num_coins_to_forward, asset_id_of_coins_to_forward: proposal.call_data.asset_id_of_coins_to_forward, amount_of_gas_to_forward: proposal.call_data.amount_of_gas_to_forward) {
```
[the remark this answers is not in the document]
matt-user: It does not exist in the std afaik.  I'll create an issue.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+
+storage {
+    // The amount of governance tokens a user has deposited
+    balances: StorageMap<Identity,
+    u64>, /// Contract Id of the governance token
+    gov_token: ContractId,
+    /// Information describing a proposal created via create_proposal(...)
+    proposals: StorageMap<u64,
+    ProposalInfo>, /// Number of created proposals
+    /// Used to check the validity of a proposal id
```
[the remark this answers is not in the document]
matt-user: hmm not sure that I would consider it a unique identifier.  In fact since I'm using 0 based "indexing" `proposal_count` should never refer to a `proposal_id`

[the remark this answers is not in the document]
Braqzen: You're passing in an ID into functions which is directly used to reference a proposal based on the count. That is a unique identifier

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+        })
+    }
+
+    /// Vote on a given proposal
+    ///
+    /// # Arguments
+    ///
+    /// - `proposal_id` - proposal to vote on
+    /// - `vote_amount` - amount of votes to use on proposal
+    /// - `is_yes_vote` - determines if you vote yes or no on the proposal
```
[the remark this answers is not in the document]
matt-user: What variable name would you suggest here?  IMO `is_yes_vote: bool` makes it immediately clear what the intent is from the dev.  IMO `is_yes_vote` is more clear name than something like `yes_vote`.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,312 @@
...
+    /// * When the proposal has not expired
+    /// * When the proposal has not met the necessary approval percentage
+    #[storage(read, write)]fn execute(proposal_id: u64) {
+        require(proposal_id < storage.proposal_count, UserError::InvalidId);
+
+        let proposal = storage.proposals.get(proposal_id);
+        require(proposal.end_height < height(), ProposalError::ProposalStillActive);
+
+        // Prevents divide by 0 error when calculating acceptance_percentage below
+        if (proposal.yes_votes == 0 && proposal.no_votes == 0) {
```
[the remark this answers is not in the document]
adlerjohn: Dividing by 0 is an automatic revert, so explicit checking isn't needed.

dao-voting/src/main.sw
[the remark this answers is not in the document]
adlerjohn: Please do, a raw low-level call like this is useful to have.

dao-voting/src/main.sw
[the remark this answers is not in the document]
matt-user: I assume division by 0 reverts.  If not I feel like that should be implemented in the language right?

dao-voting/src/main.sw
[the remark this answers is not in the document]
matt-user: But the ID is the unique identifier.  `proposal_count` does not identify any proposals so I don't think it should be reffered to as a unique identifier.

dao-voting/tests/harness.rs
matt-user: Do you mean create a module under a utils/ folder?  I tried that at one point and moved `setup()` `call_data()`, `Metadata` and `abigen!()` code there.  However when I did that I could no longer access `daovoting_mod` which I need for my test cases.  How did you handle this in fundraiser?

matt-user: I took a look at fundraisers test, but I don't see anywhere you use `contractname_mod`.  I'm able to move things over to the utils/mod file, but then I can no longer access things such as `daovoting_mod::Proposal`.  However, I may have done something wrong the first time I tried to split the file into modules, I'll give it another try.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+
+    mod success {
+        use super::*;
+
+        #[tokio::test]
+        async fn user_can_deposit() {
+            let (_gov_token, gov_token_id, deployer, user, asset_amount) = setup().await;
+
+            assert!(
+                deployer
```
Braqzen, line 194: This should be a helper func and it does not need an assertion

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .tx_params(tx_params)
+                .call_params(call_params)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            assert_eq!(
+                deployer.dao_voting.balance().call().await.unwrap().value,
+                asset_amount
```
Braqzen, line 242: Note that it is possible to make this test pass while having the contract not actually take a deposit because you are not making sure that the asset_amount != 0. Passing in params into functions instead of hardcoding is a good practice

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                    .call()
+                    .await
+                    .unwrap()
+                    .value,
+                asset_amount
+            );
+        }
+    }
+
+    mod revert {
```
Braqzen, line 257: I would order all reverts in the same order that they may occur in the contract. It makes it easier to follow

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+
+            deployer
+                .dao_voting
+                .constructor(gov_token_id)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            assert_eq!(deployer.dao_voting.balance().call().await.unwrap().value, 0);
```
Braqzen, line 285: No need to check balance since we are panicking anyway. If it were to pass then we would compare the balance before and after

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+            );
+        }
+    }
+
+    mod revert {
+        use super::*;
+
+        #[tokio::test]
+        #[should_panic]
+        async fn panics_with_incorrect_amount() {
```
Braqzen, line 262: I would be more specific with the name, what is an incorrect amount? zero.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .constructor(gov_token_id)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            let tx_params = TxParameters::new(None, Some(1_000_000), None, None);
+            let call_params = CallParameters::new(
+                Some(asset_amount),
+                Some(AssetId::from(*gov_token_id)),
```
Braqzen, line 370: This should be calling with the id of the "another asset"

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .call_params(call_params)
+                .call()
+                .await
+                .unwrap()
+                .value;
+        }
+    }
+}
+
+mod vote {
```
Braqzen, line 385: Can you order the modules to be in the same order as the abi functions? It makes it easier to follow when reviewing. It can easily take 2h to review something just a single time let alone multiple times. It makes it difficult to catch everything and even more so if things are out of order and I have to keep jumping around different files and within the same file

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .dao_voting
+                .constructor(gov_token_id)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            assert_eq!(deployer.dao_voting.balance().call().await.unwrap().value, 0);
+
+            assert_eq!(
```
Braqzen, line 668: You don't need to test this setup

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
```
Braqzen, line 760: Not needed

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .await
+                .unwrap()
+                .value;
+
+            assert_eq!(
+                deployer.dao_voting.balance().call().await.unwrap().value,
+                asset_amount
+            );
+
+            assert_eq!(
```
Braqzen, line 790: panic so not needed

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                asset_amount
+            );
+
+            user.dao_voting
+                .withdraw(asset_amount * 100)
+                .append_variable_outputs(1)
+                .call()
+                .await
+                .unwrap()
+                .value;
```
Braqzen, line 806: `unwrap()` will panic so no need to always call for `.value`

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+}
+
+mod vote {
+    use super::*;
+
+    mod success {
+        use super::*;
+
+        #[tokio::test]
+        async fn user_can_vote() {
```
Braqzen, line 392: need another test to prove that they can vote on multiple proposals

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+            );
+        }
+    }
+
+    mod revert {
+        use super::*;
+
+        #[tokio::test]
+        #[should_panic]
+        async fn panics_on_not_enough_votes() {
```
Braqzen, line 473: be specific, voting with more than they have.
also, no test for 0 votes

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+
+            user.dao_voting
+                .create_proposal(10, 10, call_data)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            user.dao_voting
+                .vote(true, 0, asset_amount / 4)
```
Braqzen, line 493: You don't need the `/ 4` here. That is just confusing

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .deposit()
+                .tx_params(tx_params)
+                .call_params(call_params)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            user.dao_voting
+                .vote(true, 0, asset_amount / 4)
```
Braqzen, line 549: again, magic division / numbers are confusing and not needed here. just use the full amount

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                    no_votes: asset_amount / 4,
+                    acceptance_percentage: 10,
+                    proposal_transaction: call_data,
+                    deadline: 15,
+                }
+            );
+        }
+    }
+
+    mod revert {
```
Braqzen, line 468: no test to check voting on invalid ID

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .vote(true, 0, asset_amount / 4)
+                .call()
+                .await
+                .unwrap()
+                .value;
+        }
+    }
+}
+
+mod execute_proposal {
```
Braqzen, line 558:
```suggestion
mod execute {
```

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+            user.dao_voting
+                .vote(false, 0, asset_amount / 4)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            let proposal = user.dao_voting.proposal(0).call().await.unwrap().value;
+
+            assert_eq!(
```
Braqzen, line 454: no check for user votes or user balance

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+}
+
+mod execute_proposal {
+    use super::*;
+
+    mod success {
+        use super::*;
+
+        #[tokio::test]
+        async fn user_proposal_can_execute() {
```
Braqzen, line 565: you must at least call the execute function. if the VM does not support the functionality yet then comment out the test otherwise implement it fully before opening a PR. Basically no one is going to review things as thoroughly as me so they wouldn't spot this and problems could arise in master.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .unwrap()
+                .value;
+        }
+    }
+}
+
+mod execute_proposal {
+    use super::*;
+
+    mod success {
```
Braqzen, line 561: this does not have a revert case and the function reverts in a few cases.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+            user.dao_voting
+                .vote(true, 0, asset_amount / 2)
+                .call()
+                .await
+                .unwrap()
+                .value;
+
+            let proposal = user.dao_voting.proposal(0).call().await.unwrap().value;
+
+            assert_eq!(
```
Braqzen, line 872: not needed, you should check the user votes instead (before and after to see that it has been unlocked)

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                    no_votes: 0,
+                    proposal_transaction: call_data,
+                    deadline: 6,
+                    acceptance_percentage: 1
+                }
+            );
+
+            user.dao_voting.unlock_votes(0).call().await.unwrap();
+
+            assert_eq!(
```
Braqzen, line 886: does not check votes, only balance

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                .unwrap()
+                .value;
+        }
+    }
+}
+
+mod convert_votes {
+    use super::*;
+
+    mod success {
```
Braqzen, line 814: no reverts

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,897 @@
...
+                    .user_balance(Identity::Address(user.wallet.address()))
+                    .call()
+                    .await
+                    .unwrap()
+                    .value,
+                asset_amount
+            );
+        }
+    }
+}
```
Braqzen, line 897: no explicit test cases for

1. balance
2. user_balance
3. user_votes
4. proposal

dao-voting/tests/harness.rs
Braqzen: I think it took me 3h+ to go through it this time mainly because of the tests...

Braqzen (changes requested):
https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Ftse2.mm.bing.net%2Fth%3Fid%3DOIP.Wn874XKWsrhJc6TGtC6ZjwHaE8%26pid%3DApi&f=1

Not bad for the contract. Not much more to do there.

Tests need a lot more work.
Can you also add this to the base readme for the main repo and add a placeholder readme for this specific project?  You'll be able to find that in the other repos

dao-voting/src/data_structures.sw
```diff
@@ -0,0 +1,55 @@
...
+    amount: u64,
+    /// Asset Id of the coins to forward
+    asset: ContractId,
+    /// Stores information about an arbitrary contract function call
+    call_data: CallData,
+    /// Specifies the amount of gas to forward to the arbitrary function call
+    gas: u64,
+}
+
+struct CallData {
```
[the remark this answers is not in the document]
matt-user: hmm not sure alphabetically is the best way to order these structs.  To me it makes the most sense to order the structs similarly to how one would order functions (in increasing detail).  For example I think the following ordering is better than alphabetical. 
```
fn b() {
    a();
}

fn a() {}
```
In general I think code should be in increasing order of detail.
[here] (http://websites.umich.edu/~eecs381/handouts/C++_Coding_Standards.pdf) (page 39) is where I got this reasoning from.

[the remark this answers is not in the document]
Braqzen: That may work for the dev but not for anyone else reading the code

dao-voting/src/data_structures.sw
```diff
@@ -0,0 +1,55 @@
+library data_structures;
+
+use std::{contract_id::ContractId, identity::Identity};
+use core::ops::Eq;
+
+pub enum State {
+    NotInitialized: (),
+    Initialized: (),
```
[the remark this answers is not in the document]
matt-user: I can, but how do you initialize the enum with the proper value (NotInitialized)?  When I change the order the inital value of State becomes Initialized.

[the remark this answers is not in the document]
Braqzen: Change your require conditions?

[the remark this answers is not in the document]
matt-user: Could you elaborate?
Changing this
`require(storage.state == State::NotInitialized, InitializationError::CannotReinitialize);`
to this
`require(storage.state == State::Initialized, InitializationError::CannotReinitialize);`
Doesn't make sense to me.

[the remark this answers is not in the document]
Braqzen: Fine. Leave it out of order. I'm too fatigued to think of a solution rn (renaming them may work).

dao-voting/src/events.sw
```diff
@@ -0,0 +1,61 @@
+library events;
+
+dep data_structures;
+
+use std::identity::Identity;
+use data_structures::ProposalInfo;
+
+// TODO ename once https://github.com/FuelLabs/sway/issues/2106 is fixed
+pub struct CreatePropEvent {
```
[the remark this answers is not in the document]
matt-user: I'm not entirely sure how event filtering works but does not not work like this?
`events.CreatePropEvent.filter(where author == author)`
`events.CreatePropEvent.filter(where proposal_info.author == author)`

dao-voting/tests/utils/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+            gov_token: None,
+            wallet: user_wallet,
+        };
+
+        let asset_amount: u64 = 10;
+
+        (gov_token, gov_token_id, deployer, user, asset_amount)
+    }
+
+    pub fn get_call_data(asset_id: ContractId) -> daovoting_mod::Proposal {
```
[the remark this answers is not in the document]
matt-user: No its not, for some reason I thought I needed the `daovoting_mod::` prefix to access sway structs.

dao-voting/src/events.sw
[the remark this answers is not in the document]
Braqzen: I assumed it would be the first one but I don't actually know. If I'm incorrect then let me know and I'll update my cognitive framework

dao-voting/tests/harness.rs
matt-user: hmm I don't think its possible for that to happen.  The contract's deposit function requires the amount to be greater than 0.  Therefore the function will fail if `asset_amount` is 0.

dao-voting/src/events.sw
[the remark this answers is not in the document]
matt-user: Looking at the ts and rs sdks it looks like event monitoring is unimplemented.  I'll post a question about events tomorrow.

dao-voting/tests/harness.rs
matt-user: I was unable to get the execute function to work.  I think its because I'm passing in dummy data which causes an invalid function call and a revert.  I'm going to comment out the test until I can properly encode a function selector and arguments.

Braqzen: I think it is acceptable to either skeleton out as much of it as you can so that you have less work later (and comment it out) or just create the boilerplate to indicate which functional conditions need to be tested and then comment it all out. That way it's clear what needs to be tested later. There is an alternative where you should be able to annotate the test such that it automatically skips it but you'll need to look up what the annotation is to do that

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,898 @@
...
+            create_proposal(&user, 10, 1, proposal_transaction.clone()).await;
+            vote(&user, true, 0, asset_amount / 2).await;
+
+            execute(&user, 0).await;
+
+            // TODO actually test execution of an arbitrary transaction
+        }
+    }
+
+    mod revert {
```
Braqzen, line 556: Another test to consider is whether we have a revert for a failed call (once that works)

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,898 @@
...
+}
+
+mod unlock_votes {
+    use super::*;
+
+    mod success {
+        use super::*;
+
+        #[tokio::test]
+        async fn user_can_unlock_tokens() {
```
Braqzen, line 631: I'd check to see if they can unlock from more than 1

dao-voting/tests/utils/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) {
+        user.dao_voting
+            .create_proposal(acceptance_percentage, deadline, proposal)
+            .call()
+            .await
+            .unwrap()
+            .value;
+    }
+
+    pub async fn deposit(user: &Metadata, tx_params: TxParameters, call_params: CallParameters) {
```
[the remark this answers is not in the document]
Braqzen: It seems like your Tx is always the same so I would hardcode it here

Braqzen (changes requested):
Can you also add it to the [README](https://github.com/FuelLabs/sway-applications#projects) and add a dummy [README](https://github.com/FuelLabs/sway-applications/tree/master/fundraiser#overview) to the actual DAO project itself?

dao-voting/src/main.sw
```diff
@@ -0,0 +1,317 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// - `amount` - amount of governance tokens to withdraw from the contract
+    ///
+    /// # Reverts
+    ///
+    /// * When the user tries to withdraw more than their balance
+    #[storage(read, write)]fn withdraw(amount: u64) {
+        let sender: Identity = msg_sender().unwrap();
```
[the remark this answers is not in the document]
matt-user: hmm there's no NotInit check in the withdraw function.  But I do agree a require check for an amount greater than 0 is a good idea for this function.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,317 @@
...
+    /// - `proposal_id` - Identifier used to specifiy a proposal (0 <= proposal_id < proposal_count)
+    /// - `vote_amount` - the amount of votes to cast on the proposal
+    ///
+    /// # Reverts
+    ///
+    /// * When the given proposal id is greater than or equal to proposal_count
+    /// * When the vote amount is 0
+    /// * When the proposal has passed its deadline
+    /// * When the vote amount is greater than the users deposited balance
+    #[storage(read, write)]fn vote(approve: bool, proposal_id: u64, vote_amount: u64, ) {
```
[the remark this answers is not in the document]
matt-user: Should I make an issue for this in sway?  It seems like this should be a compiler error.

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,898 @@
...
+}
+
+mod deposit {
+    use super::*;
+
+    mod success {
+        use super::*;
+
+        #[tokio::test]
+        async fn user_can_deposit() {
```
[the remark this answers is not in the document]
matt-user: The user can't deposit into multiple proposals.  They can only deposit into the contract.  However, they can vote and withdraw from multiple proposals.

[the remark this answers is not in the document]
matt-user: So I'll add test cases for those

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,898 @@
...
+                CallParameters::new(Some(0), Some(AssetId::from(*gov_token_id)), Some(100_000));
+            deposit(&user, tx_params, call_params).await;
+        }
+    }
+}
+
+mod withdraw {
+    use super::*;
+
+    mod success {
```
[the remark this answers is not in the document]
matt-user: Similar to deposit the user can only withdraw from the contract, not individual proposals.  They can unlock_votes from multiple proposals so I'll add a test for that.

dao-voting/tests/harness.rs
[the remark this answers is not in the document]
Braqzen: makes sense, brain fart from me

dao-voting/src/main.sw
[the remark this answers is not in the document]
matt-user: I don't think it was the formatter.  I probably deleted a parameter and didn't get the comma.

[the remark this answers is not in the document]
matt-user: It looks like it's being tracked [here](https://github.com/FuelLabs/sway/issues/2134)

dao-voting/tests/utils/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+);
+
+pub struct Metadata {
+    pub dao_voting: DaoVoting,
+    pub gov_token: Option<GovToken>,
+    pub wallet: LocalWallet,
+}
+
+pub mod abi_calls {
+    use fuels::contract::contract::CallResponse;
```
Braqzen, line 17: Since you're using `use super::*` below we should have access to everything so this line should not be needed.

dao-voting/src/main.sw
```diff
@@ -0,0 +1,327 @@
...
+            author: msg_sender().unwrap(), token
+        })
+    }
+
+    /// Create a new proposal
+    ///
+    /// # Arguments
+    ///
+    /// - `acceptance_percentage` - the percentage of yes votes a proposal needs to be executed
+    /// - `deadline` - the number of blocks during which a proposal can be voted on
```
Braqzen, line 83:
```suggestion
    /// - `duration` - the number of blocks during which a proposal can be voted on
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,327 @@
...
+        require(storage.state == State::Initialized, InitializationError::ContractNotInitialized);
+        require(storage.token == msg_asset_id(), UserError::IncorrectAssetSent);
+        require(0 < msg_amount(), UserError::AmountCannotBeZero);
+
+        let user = msg_sender().unwrap();
+
+        storage.balances.insert(user, msg_amount() + storage.balances.get(user));
+
+        log(DepositEvent {
+            amount: msg_amount(), user: user
```
Braqzen, line 127:
```suggestion
            amount: msg_amount(), user
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,327 @@
...
+        let prev_balance = storage.balances.get(user);
+        require(amount <= prev_balance, UserError::InsufficientBalance);
+
+        storage.balances.insert(user, prev_balance - amount);
+
+        // Transfer the asset back to the user
+        transfer(amount, storage.token, user);
+
+        log(WithdrawEvent {
+            amount, user: user,
```
Braqzen, line 154:
```suggestion
            amount, user
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,327 @@
...
+
+    /// Execute a given proposal
+    ///
+    /// # Arguments
+    ///
+    /// - `proposal_id` - Identifier used to specifiy a proposal (0 <= proposal_id < proposal_count)
+    ///
+    /// # Reverts
+    ///
+    /// * When the given proposal id is greater than or equal to proposal_count
```
Braqzen, line 210:
```suggestion
    /// * When the given proposal id is greater than or equal to proposal_count
    /// * When the proposal has already been executed
```

dao-voting/src/main.sw
```diff
@@ -0,0 +1,327 @@
...
+        // When close to the u64 max
+        // https://github.com/FuelLabs/sway-applications/issues/106
+        let acceptance_percentage = proposal.yes_votes * 100 / (proposal.yes_votes + proposal.no_votes);
+        require(proposal.acceptance_percentage <= acceptance_percentage, ProposalError::InsufficientApprovals);
+
+        asm(call_data: proposal.proposal_transaction.call_data, amount: proposal.proposal_transaction.amount, asset: proposal.proposal_transaction.asset, gas: proposal.proposal_transaction.gas) {
+            call call_data amount asset gas;
+        }
+
+        proposal.executed = true;
```
Braqzen, line 230: Given that we are calling a contract the proposal update should be above the call to prevent re-entrency. I do not recall if the VM automatically handles that. I know we have a call to check. Let's be safe and move the two lines before the `asm()`

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,1156 @@
...
+            assert_eq!(balance(&user.dao_voting).await, asset_amount);
+
+            assert_eq!(
+                user_balance(&user.dao_voting, Identity::Address(user.wallet.address())).await,
+                asset_amount
+            );
+        }
+
+        #[tokio::test]
+        async fn user_can_make_multiple_deposits() {
```
Braqzen, line 200: As you've correctly pointed out this test does not need to exist - plus you've only copy and pasted the code and gave it a different name... delete this test

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,1156 @@
...
+        async fn panics_on_vote_amount_greater_than_balance() {
+            let (_gov_token, gov_token_id, deployer, user, asset_amount) = setup().await;
+            constructor(&deployer.dao_voting, gov_token_id).await;
+
+            let proposal_transaction = proposal_transaction(gov_token_id);
+            create_proposal(&user.dao_voting, 10, 10, proposal_transaction.clone()).await;
+            vote(&user.dao_voting, true, 10, asset_amount).await;
+        }
+
+        // TODO add test for reverting on a failed proposal call
```
Braqzen, line 593: I'd move this comment down a bit so that it is near the revert of the execute suite rather than the voting suite

dao-voting/tests/harness.rs
```diff
@@ -0,0 +1,1156 @@
...
+                .await,
+                Votes {
+                    yes_votes: 0,
+                    no_votes: 0
+                }
+            );
+        }
+
+        #[tokio::test]
+        async fn user_can_unlock_tokens_from_multiple_proposals() {
```
Braqzen, line 795: This tests to see if a user can use the same tokens to vote on 1 proposal after another (which is okay) but we also need a test to see if the user can unlock tokens while there are 2 proposals active simultaneously

Braqzen (changes requested):

dao-voting/src/main.sw
matt-user: I don't understand this suggestion could you elaborate?

dao-voting/tests/harness.rs
matt-user: Whoops that was supposed to be deleted...

dao-voting/src/main.sw
Braqzen: Revert documenting the addition of

https://github.com/FuelLabs/sway-applications/blob/5a578e5b1b55d0cd43bba988aa261bfb3ae7c078/dao-voting/src/main.sw#L217

dao-voting/tests/harness.rs
matt-user: hmm currently a user can only unlock votes if the proposal is expired.  That is all votes are final and cannot be changed.  I admit this may not be the best design decision, but we can handle that in a future pr.

dao-voting/src/main.sw
matt-user: lol my b I've been looking at this code too long...

dao-voting/tests/harness.rs
Braqzen: What I mean is
- User has 100 tokens
- 2 Proposals exist
- User deposits and votes some % to each
- Both both expire
- Convert into deposit

Rather than
- User has 100 tokens
- Create 1 proposal
- Deposit and vote all
- Expires
- Convert
- Create new proposal
- ...

The point is to make sure that you can withdraw / convert / whatever while more than 1 proposal already exists in state

Braqzen (approved):
https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Ftse3.mm.bing.net%2Fth%3Fid%3DOIP.fwkkoLVN41Vlq4aN9cB5FQHaDt%26pid%3DApi&f=1

PR #86, opened 2022-06-10
Replaced compilation of forc and forc-fmt with binary fetch

### LINKED ISSUE #83: Update CI to use `fuelup`
### How does this improve a feature?

Should significantly reduce CI time if we can fetch pre-built binaries

### Specification

Replace `forc` and `forc fmt` actions to have 1 `fuelup` section

Braqzen:
# Summary

## Type of change

- Improvement (refactoring, restructoring repository, cleaning tech debt, ...)

## Changes

- Removed `forc`
- Removed `forc-fmt`
- Added `fuelup`

# Related Issues

Closes #83 

# Additional Information

- It's blazingly fast
- Do you have any idea how fast I am?


Braqzen:
**Note** This uses a PR link for now. Once `fuelup` is updated to contain these changes then another issue can be filed to update to use the correct (release) link.

.github/workflows/ci.yml
```diff
@@ -75,17 +74,10 @@ jobs:
...
-          args: forc --version ${{ env.FORC_VERSION }}
-
-      - name: Install forc-fmt
-        uses: actions-rs/cargo@v1
-        with:
-          command: install
-          args: forc-fmt --version ${{ env.FORC_VERSION }}
+      - name: Install fuelup
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/FuelLabs/fuelup/binggh/remove-tty/fuelup-init.sh | sh -s -- --no-modify-path
```
adlerjohn, line 79: This definitely should be the deployed version instead of a branch before we merge this.

adlerjohn (commented):
Hmmm. So even once the fix is deployed in `fuelup-init.sh`, my concern is around the `forc` version. Without https://github.com/FuelLabs/fuelup/issues/46, you'll run the risk of installing the latest version, which may break compatibility. That being said, it's not the end of the world since you would need to bump anyways and transient CI failures are fine at this stage of development. Not a blocker.

.github/workflows/ci.yml
Braqzen: Your recent commentary indicated to me that you wanted to be on the latest version and bump everything up every time

adlerjohn: Every time yes, but in a controlled way. I.e. the bumping is done explicitly instead of automatically in the middle of unrelated PRs.

adlerjohn: To ensure that applications build with the version of `forc` that's on `master`, we could either:

1. Manually ensure that bumping `forc` in `master` is then applied to open PRs and branches before merging them, or
2. Enforce at the repo level that all branches must be up to date with `master` before merging. This prevents branches from being merged while using a different version. This might require PRs to be re-approved after they've been updated, making merging multiple PRs in parallel more cumbersome.

Thoughts @Braqzen?

Braqzen: I don't know what you mean by `forc` on `master`. Is that in Sway or Sway-Applications?

Any new PR in this repository should aim to use the latest release in Sway.
If that release does not break any other project then nothing needs to be done.
If there are breaking changes then any project that is broken must be updated first and merged into master and subsequently into the PR.

This whole process is a pain because now every application is coupled together and when enough PRs come in all the work will consist of constantly updating your work to match another that just went in.
This is why each app should be on their own version.

adlerjohn: > I don't know what you mean by forc on master. Is that in Sway or Sway-Applications?

In this repo.

Braqzen: Okay, still... 1 and 2 sound the same to me.

1) How do you manually check? Does this mean no CI at all? I doubt you mean that and if you bump the version then that won't just affect the open PRs, it will affect all projects and then any number of them have to be updated.

2) "At the repo level" sounds like CI which is point one above. This doesn't make sense to me.

adlerjohn (approved):

bitzoic (approved):

nfurfaro (approved):
Looks good.

PR #92, opened 2022-06-13
English Auction

### LINKED ISSUE #70: English auction program
An English auction is one where the auctioneer sets an opening price and a reserve price that must be met in order to sell the item.
Bidders continue to place bids in order to be the highest bidder and the highest bidder will be the one the item is sold to if conditions are met.

bitzoic:
# Summary

Implementation of an English Auction Contract. 

The contract shall be initialized to accept an asset in return for the sold asset. A selling asset must be provided on initialization. An initial price must be set and a reserve price may be set. A time for the auction must be set.

While the auction is active:
1. A bid can be made at or higher than the initial price if no bid has been made 
2. A bid can be made higher than the current winning bid
3. The reserve can be met and the assets will be bought outright

When the auction is over the buyer, seller, and failed bids can be withdrawn. The auction will end when the reserve is met, time runs out, or is canceled.

Both tokens and NFTs can be listed for auction. Both tokens and NFTs can be provided as bids.

## Type of change

New contract application

## Changes

- Added main contract
- Added interface library
- Added errors library
- Added data structures library
- Added events library
- Added utils library
- Added tests
- Added README.md
- Added SPECIFICATION.md
- Updated APPLICATION_PROGRESS.md

# Related Issues

Closes #70

# Feature(s) Required

### `StorageVec` in structs

The use of a `StorageVec` in structs will be needed to auction off and bid multiple NFTs. This can be tracked [here](https://github.com/FuelLabs/sway/issues/2465)

### Formatting Bug

Match arms with 2 arguments over a certain length result in a [formatter bug](https://github.com/FuelLabs/sway/issues/3131)

### Ordering of Dependencies

There are dependencies that are not ordered alphabetically due to a [dependency ordering problem](https://github.com/FuelLabs/sway/issues/3131)


auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub struct TokenAsset {
+    /// The amount of the native asset that the struct is representing.
+    amount: u64,
+    /// The `ContractId` of the native asset that the struct is representing.
+    contract_id: ContractId,
+}
+
+pub trait AssetTraits {
```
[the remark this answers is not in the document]
Braqzen: The contract id doesn't make sense to me as a trait.
The amount could be a trait depending on the token and what it exposes e.g keep an internal balance and show a user the balance from some given calculation.
Does that apply here though?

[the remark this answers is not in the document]
Braqzen: Yeah, I really don't know about all of these traits. Will need to look through the code to see if this structure makes sense

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,114 @@
+library utils;
+
+dep abi;
+dep data_structures;
+dep errors;
+
+use abi::NFT;
+use data_structures::*;
```
Braqzen, line 8: Explicit imports should work

auctions/english-auction/tests/artifacts/asset/src/main.sw
```diff
@@ -0,0 +1,14 @@
...
+
+use std::{address::Address, token::mint_to_address};
+
+abi Asset {
+    fn mint_and_send_to_address(amount: u64, recipient: Address) -> bool;
+}
+
+impl Asset for Contract {
+    fn mint_and_send_to_address(amount: u64, recipient: Address) -> bool {
+        mint_to_address(amount, recipient);
```
Braqzen, line 11: No minting to contract via the general transfer function?

README.md
```diff
@@ -50,6 +50,7 @@ sway-applications/
 - [Escrow](./escrow) is a third party that keeps an asset on behalf of multiple parties
 - [Fundraiser](./fundraiser/) is a program allowing users to pledge towards a goal
 - [Multi-Signature Wallet](./multisig-wallet) is a wallet that requires multiple signatures to execute a transaction
+- [English Auction](./auctions/english-auction/) is a English Auction template where user bid up the price of an asset until time is up or a reserve has been met.
```
Braqzen, line 53: Move to top, check grammar for sentence and nest under

- Auctions
    - Auction 1
    - 2
    - 3
    - ...

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,377 @@
+contract;
+
+dep abi;
+dep data_structures;
+dep errors;
+dep events;
+dep utils;
+
+use abi::{EnglishAuction, NFT};
+use data_structures::*;
```
Braqzen, line 10: Explicit imports should work

auctions/english-auction/tests/utils/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        let token = NFTAsset {
+            contract_id,
+            token_ids,
+        };
+
+        Asset::NFTAsset(token)
+    }
+}
+
+pub mod abi_calls {
```
Braqzen, line 131: can you change the names of the helpers to be the same as the function names?

auctions/english-auction/tests/utils/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+                reserve_price,
+                time,
+            )
+            .call()
+            .await
+            .unwrap()
+            .value
+    }
+
+    pub async fn bid_tokens(
```
Braqzen, line 200: they are not abi functions. these are helpers similar to the setup(). these should be generic to "bid". likewise all the other functions

auctions/english-auction/tests/utils/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+            .auction
+            .total_auctions()
+            .call()
+            .await
+            .unwrap()
+            .value
+    }
+}
+
+pub async fn deploy_funds(deploy_wallet: &Metadata, wallet: &LocalWallet, asset_amount: u64) {
```
Braqzen, line 338: This should be in test helpers too

Braqzen (changes requested):
I'll review more tomorrow

auctions/english-auction/src/main.sw
bitzoic: Last I checked it didn't work with my trait but I'll try again.

Braqzen: Update toolchain, it has been fixed in 0.16.2 iirc

auctions/english-auction/src/abi.sw
```diff
@@ -0,0 +1,23 @@
...
+    // #[storage(read)]fn auction_info(auction_id: u64) -> Option<Auction>;
+    #[storage(read, write)]fn bid(auction_id: u64, asset: Asset);
+    #[storage(read, write)]fn cancel_auction(auction_id: u64);
+    #[storage(read, write)]fn constructor(seller: Identity, sell_asset: Asset, buy_asset: Asset, inital_price: u64, reserve_price: u64, time: u64) -> u64;
+    // #[storage(read)]fn deposits(identity: Identity, auction_id: u64) -> Option<Asset>;
+    #[storage(read, write)]fn withdraw(auction_id: u64);
+    #[storage(read)]fn total_auctions() -> u64;
+}
+
+abi NFT {
```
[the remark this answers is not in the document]
bitzoic: I cannot handle native tokens and NFTs the same. For native tokens I can check the message amount, but for NFTs I have to interact with the NFT contract and transfer it to the auction contract. It cannot be included in the transaction. For that, I need to know the abi of  the NFT and the functions I would like to call. Is there a way for me to import without defining in the abi.sw and then using `dep abi` in my utils.sw file?

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,185 @@
+library data_structures;
+
+use std:: {
+    assert::assert, contract_id::ContractId, identity::Identity, option::Option, storage::StorageMap
+};
+
+pub enum Asset {
```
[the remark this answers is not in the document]
bitzoic: In the main.sw I am trying to minimize the use of match statements absolutely everywhere to differentiate between NFTs and native tokens. 99% of their functionality is the same. To do this I have two structs, one for each type, and then an enum to handle them identically. I then implement traits such as Eq and greater than to easily handle both cases.

auctions/english-auction/src/abi.sw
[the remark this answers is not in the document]
Braqzen: There's talk of being able to call contracts without knowing much about them. I do not recall the specifics though. I think using the "call" opcode, similar to what @matt-user has recently done in the DAO might be useful. I have mentioned that his code should be added to the std so perhaps once that is done then you can use it?

I haven't looked into it so matt can comment more on this.

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,377 @@
...
+
+        // Make sure the sender still has something to withdraw
+        require(sender_deposit.is_some(), UserError::UserHasAlreadyWithdrawn);
+        storage.deposits.insert((sender, auction_id), Option::None());
+        let mut withdrawnAsset = sender_deposit.unwrap();
+
+        // Go ahead and withdraw
+        if (bidder.is_some() && sender == bidder.unwrap()) {
+            // The buyer is withdrawing
+            match auction.sell_asset {
```
bitzoic, line 324: These will need to be put in an external utils file. A lot of repetitive code here.

Braqzen: I saw the control flow and thought "not today... I'll deal with that tomorrow"

bitzoic: Yeah it was temporary lol. Already updated

auctions/english-auction/src/abi.sw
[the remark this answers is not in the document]
matt-user: I don't know if there is a way to import without defining the abi.sw and using dep abi.  If you want to call a function on a contract not known at compile time you can use the call opcode specified [here](https://github.com/FuelLabs/fuel-specs/blob/1be31f70c757d8390f74b9e1b3beb096620553eb/specs/vm/instruction_set.md#call-call-contract).  However, currently I do not know of a way to encode the function selector and function arguments with the sdk, which is necessary to give the contract what it needs to use the call opcode.  However, it seems like you know your abi at compile time so it might be easiest just to use the NFT abi.

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+        match(self, other) {
+            (Asset::NFTAsset(nft_asset1), Asset::NFTAsset(nft_asset2)) => {
+                // TODO: Compare NFT token ID Vec length
+                nft_asset1.contract_id == nft_asset2.contract_id
+            },
+            (Asset::TokenAsset(token_asset1), Asset::TokenAsset(token_asset2)) => {
+                token_asset1.contract_id == token_asset2.contract_id && token_asset1.amount > token_asset2.amount
+            },
+            _ => {
+                false
```
Braqzen, line 155: revert too

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+        match(self, other) {
+            (Asset::NFTAsset(nft_asset1), Asset::NFTAsset(nft_asset2)) => {
+                // TODO: Compare NFT token ID Vec length
+                nft_asset1.contract_id == nft_asset2.contract_id
+            },
+            (Asset::TokenAsset(token_asset1), Asset::TokenAsset(token_asset2)) => {
+                token_asset1.contract_id == token_asset2.contract_id && token_asset1.amount < token_asset2.amount
+            },
+            _ => {
+                false
```
Braqzen, line 170: revert too

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+            (Asset::TokenAsset(token_asset1), Asset::TokenAsset(token_asset2)) => {
+                require(token_asset1.contract_id == token_asset2.contract_id, AssetError::AssetsAreNotTheSame);
+                let total_amount = token_asset1.amount - token_asset2.amount;
+                let token = TokenAsset {
+                    amount: total_amount, contract_id: token_asset1.contract_id
+                };
+                Asset::TokenAsset(token)
+            },
+            _ => {
+                self
```
Braqzen, line 193: revert too

auctions/english-auction/src/events.sw
```diff
@@ -0,0 +1,28 @@
...
+
+use data_structures::{Asset, Auction};
+use std::identity::Identity;
+
+pub struct AuctionCancelEvent {
+    auction: Auction,
+    auction_id: u64,
+}
+
+pub struct AuctionStartEvent {
```
Braqzen, line 13: I think "created" makes more sense than start since you don't create it and then wait, correct? I assume you create it and start it instantly

auctions/english-auction/src/events.sw
```diff
@@ -0,0 +1,28 @@
+library events;
+
+dep data_structures;
+
+use data_structures::{Asset, Auction};
+use std::identity::Identity;
+
+pub struct AuctionCancelEvent {
```
Braqzen, line 8:
```suggestion
pub struct AuctionCancelledEvent {
```

auctions/english-auction/src/events.sw
```diff
@@ -0,0 +1,28 @@
...
+    auction: Auction,
```
Braqzen, line 9: You probably don't need to emit the auction itself since the front end can query by the id

auctions/english-auction/src/events.sw
```diff
@@ -0,0 +1,28 @@
...
+    auction_id: u64,
+}
+
+pub struct AuctionStartEvent {
+    auction: Auction,
```
Braqzen, line 14: @matt-user has recently asked a similar question however for now I think it's better to include the creator directly in the event and if it turns out that we can filter inside the Auction then that can be removed later

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,125 @@
...
+    assert::require,
+    chain::auth::{AuthError, msg_sender},
+    context::{call_frames::{contract_id, msg_asset_id}, msg_amount},
+    identity::Identity,
+    option::Option,
+    result::Result,
+    token::transfer,
+};
+
+/// This function returns true if the `to` `Identity` is approved to transfer the token.
```
Braqzen, line 20: I'm not going to say to remove the docs, yet, but idk if we should be documenting the functions, like this, that are not directly on the ABI. we probably should though... 

can you order the params and the docs in the same order?

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,125 @@
...
+    let token_id = asset.token_ids;
+
+    let nft_abi = abi(NFT, nft_contract.value);
+    // TODO: This will need to loop over a Vec of token_ids
+    let token_owner: Option<Identity> = nft_abi.owner_of(token_id);
+    token_owner.is_some() && owner == token_owner.unwrap()
+}
+
+/// This function will return the identity of the sender.
+pub fn sender_identity() -> Identity {
```
Braqzen, line 59: this is no longer needed in the latest release

Braqzen: you can unwrap the sender directly without assigning to anything

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,125 @@
...
+    let sender: Result<Identity, AuthError> = msg_sender();
+    sender.unwrap()
+}
+
+/// This function will transfer the `Asset` given to the specified `Identity`.
+pub fn transfer_asset(to: Identity, asset: Asset) {
+    match asset {
+        Asset::NFTAsset(asset) => {
+            transfer_nft(Identity::ContractId(contract_id()), to, asset)
```
Braqzen, line 68: I think this might not work as you expect it because it is outside a contract. 

https://github.com/FuelLabs/sway/blob/c3869c960969b51eb0fc9b9b26f75260bcf287ba/sway-lib-std/src/context/call_frames.sw#L20-L27

Braqzen (changes requested):
I'll make more comments later.

The more I look at this the more I dislike it. It's just messy. That's not a comment about your code it's a comment about mixing the functionality of the NFT and a regular token. The more I look at this the more I want to see two separate auctions. 1 for NFTs and 1 for Tokens (although I don't immediately understand why that needs to exist). I think it will be much cleaner for anyone reading the code to see one contract or the other. they probably won't want the functionality of both so it's better to separate them into two.

auctions/english-auction/src/abi.sw
```diff
@@ -0,0 +1,23 @@
+library abi;
+
+dep data_structures;
+
+use data_structures::{Asset, Auction};
+use std::{contract_id::ContractId, identity::Identity, option::Option};
+
+abi EnglishAuction {
```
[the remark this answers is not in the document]
bitzoic: I'll rename `constructor` to `create` as this makes more sense.

auctions/english-auction/src/abi.sw
```diff
@@ -0,0 +1,23 @@
...
+    #[storage(read, write)]fn bid(auction_id: u64, asset: Asset);
+    #[storage(read, write)]fn cancel_auction(auction_id: u64);
+    #[storage(read, write)]fn constructor(seller: Identity, sell_asset: Asset, buy_asset: Asset, inital_price: u64, reserve_price: u64, time: u64) -> u64;
+    // #[storage(read)]fn deposits(identity: Identity, auction_id: u64) -> Option<Asset>;
+    #[storage(read, write)]fn withdraw(auction_id: u64);
+    #[storage(read)]fn total_auctions() -> u64;
+}
+
+abi NFT {
+    fn get_approved(token_id: u64) -> Option<Identity>;
```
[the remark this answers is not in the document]
bitzoic: I believe this was updated in my NFT contract but has not been updated here

auctions/english-auction/src/abi.sw
```diff
@@ -0,0 +1,23 @@
...
+    fn is_approved_for_all(owner: Identity, operator: Identity) -> bool;
+    fn owner_of(token_id: u64) -> Option<Identity>;
+    fn transfer_from(from: Identity, to: Identity, token_id: u64);
```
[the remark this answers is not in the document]
bitzoic: Yes this matches

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+    /// On initalization, the amount will be set to 0 and the `contract_id` will be set to the
+    /// `ContractId` of the asset in return.
+    buy_asset: Asset,
+    /// The current highest bidder of the auction. When the auction is over, this is the winner.
+    /// If no one bid on the auction or the auction is canceled, this will be `None`.
+    highest_bidder: Option<Identity>,
+    /// The block at which the auction should end
+    end_block: u64,
+    /// The starting price for the auction to start. This can be 0.
+    inital_price: u64,
```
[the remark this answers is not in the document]
bitzoic: Clearly grammar and spelling are not my strong suit. This is what happens when you speak more than one language

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+            },
+            (Asset::TokenAsset(token_asset1), Asset::TokenAsset(token_asset2)) => {
+                require(token_asset1.contract_id == token_asset2.contract_id, AssetError::AssetsAreNotTheSame);
+                let total_amount = token_asset1.amount + token_asset2.amount;
+                let token = TokenAsset {
+                    amount: total_amount, contract_id: token_asset1.contract_id
+                };
+                Asset::TokenAsset(token)
+            },
+            _ => {
```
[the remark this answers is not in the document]
bitzoic: Thank you for catching this!

auctions/english-auction/src/events.sw
bitzoic: I have all my events in the present tense. It doesn't make sense to me to have a past tense event to log something that is happening in that moment. The same applies to AuctionStartEvent below. I also wouldn't have a BiddedEvent.

auctions/english-auction/src/abi.sw
```diff
@@ -0,0 +1,23 @@
+library abi;
+
+dep data_structures;
+
+use data_structures::{Asset, Auction};
+use std::{contract_id::ContractId, identity::Identity, option::Option};
+
+abi EnglishAuction {
+    // #[storage(read)]fn auction_info(auction_id: u64) -> Option<Auction>;
```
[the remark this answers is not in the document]
Braqzen: I wasn't but I think it makes sense that you should be able to query an existing auction or error out if it does not exist

auctions/english-auction/src/events.sw
bitzoic: I agree

auctions/english-auction/src/data_structures.sw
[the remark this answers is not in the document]
Braqzen: It's just a typo. Everyone makes small mistakes when they stare at a screen for a while, especially if it's the same work you just stop paying attention. This is why I make multiple passes from top to bottom looking at each line until I think it is perfect. Things still slip past me at times

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,197 @@
...
+    NFTAsset: NFTAsset,
+    TokenAsset: TokenAsset,
+}
+
+pub enum State {
+    Closed: (),
+    Open: (),
+}
+
+pub struct Auction {
```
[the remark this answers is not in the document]
Braqzen: Consistency is key. I've mentioned this recently to @matt-user . It makes it easier to review because I don't have to figure out how you structure your code. I just need to know the alphabet. I think it makes sense to keep the trait impl's below the declaration too

auctions/english-auction/src/events.sw
Braqzen: `CreateAuctionEvent`?

auctions/english-auction/src/events.sw
bitzoic: Yes that will work

auctions/english-auction/src/utils.sw
bitzoic: There can never be enough documentation imo. 

But following what we see in Solidity, any libraries imported have their own documentation as well.

bitzoic:
> I'll make more comments later.
> 
> The more I look at this the more I dislike it. It's just messy. That's not a comment about your code it's a comment about mixing the functionality of the NFT and a regular token. The more I look at this the more I want to see two separate auctions. 1 for NFTs and 1 for Tokens (although I don't immediately understand why that needs to exist). I think it will be much cleaner for anyone reading the code to see one contract or the other. they probably won't want the functionality of both so it's better to separate them into two.

I understand your sentiment. A lot of my approach here is from an object-oriented perspective, wrapping up both assets and treating them as one. It certainly adds some complexity but also generality. I don't believe this is what we traditionally see in Ethereum. 

The primary use case today will absolutely be to auction off NFTs for tokens. But the ability to go the other way around or bid NFTs, if they are to be more than monkey pictures, can also be valuable. Perhaps making the base contract more modular would work better, but I'd need to give some thought as to how to do that. Regardless, I'd definitely appreciate any feedback I can get.

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,48 @@
...
+///
+/// * `asset` - The struct which contains the NFT data.
+/// * `from` - The owner of the NFT.
+/// * `to` - The user which the NFTs should be transfered to.
+///
+/// # Reverts
+///
+/// * The NFT transfer failed.
+pub fn transfer_nft(asset: NFTAsset, from: Identity, to: Identity) {
+    let nft_contract = asset.contract_id;
```
Braqzen, line 41: Why allocate for a new variable instead of using directly?

auctions/english-auction/src/utils.sw
```diff
@@ -0,0 +1,48 @@
...
+    let nft_abi = abi(NFT, nft_contract.value);
+
+    nft_abi.transfer_from(from, to, asset.token_id);
+
+    let owner = nft_abi.owner_of(asset.token_id);
+    require(owner == to, AccessError::NFTTransferNotApproved);
```
Braqzen, line 47: Why check at the end instead of earlier?

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,265 @@
...
+                // than the current winnning bid
+                // TODO: Move this outside the match statement once StorageVec is supported in structs
+                require(token_asset.amount >= auction.initial_price, InputError::InitialPriceNotMet);
+                require(token_asset.amount > auction.bid_asset.amount(), InputError::IncorrectAmountProvided);
+            }
+        }
+
+        // Check if reserve has been met if there is one set
+        let reserve: Option<u64> = auction.reserve_price;
+        match reserve {
```
Braqzen, line 93: `if auction.reserve_price.is_some()` would make this much cleaner than the match

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,265 @@
...
+        // Update and store the auction's information
+        auction.highest_bidder = Option::None();
+        auction.state = State::Closed;
+        storage.auctions.insert(auction_id, Option::Some(auction));
+
+        log(CancelAuctionEvent { auction_id });
+    }
+
+    #[storage(read, write)]
+    fn create(
```
Braqzen, line 137: create and return? I prefer to not mix computation with getters

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,265 @@
...
+
+    #[storage(read)]
+    fn deposit(auction_id: u64, identity: Identity) -> Option<Asset> {
+        storage.deposits.get((identity, auction_id))
+    }
+
+    #[storage(read, write)]
+    fn withdraw(auction_id: u64) {
+        // Make sure this auction exists
+        let auction: Option<Auction> = storage.auctions.get(auction_id);
```
Braqzen, line 212: Why are all the option types annotated? It should work without them

Braqzen (commented):
Partial review. Re-request after comments add considered.

auctions/english-auction/README.md
```diff
@@ -0,0 +1,71 @@
+<p align="center">
+    <picture>
+        <source media="(prefers-color-scheme: dark)" srcset=".docs/english-auction_dark.png">
+        <img alt="light theme" src=".docs/english-auction_light.png">
+    </picture>
+</p>
+
+## Overview
+
+An english auction is an auction in the most traditional sense and the most well known of auction types; a seller places an asset for auction with a initial price and a reserve price, and then bidders will begin to bid until the bidding period has ended or the reserve has been met. The English Auction application implements this idea in a decentralized manner without the need for a 3rd party and with strong settlement assurances.
```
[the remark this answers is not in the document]
simonr0204: I agree on dropping "traditional",  but it's definitely the most well known - I don't think we need to qualify that.

auctions/english-auction/SPECIFICATION.md
```diff
@@ -0,0 +1,80 @@
...
+
+### English Auction State Checks
+
+#### `auction_info()`
+
+1. Returns information on a specific auction
+
+#### `deposit()`
+
+1. Returns the total deposits owed to a user for a specific auction
```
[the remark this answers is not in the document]
bitzoic: I'm using the "owed" terminology as depending on the outcome of the auction they will receive different assets i.e. the seller could receive either their original deposit or the winning bidder's deposit. 

If I just used "deposit", it would sound as if they were receiving what the user deposited rather than their "owed" or balance. I'm not 100% happy with it but I'm unable to find a better word, if you have a suggestion for better terminology I am all ears.

auctions/english-auction/SPECIFICATION.md
```diff
@@ -0,0 +1,80 @@
...
+1. Allows a user to create a new auction if
+    1. They have provided an asset to sell
+    2. They have provided an asset type to accept as bids
+    3. They have provided an initial price
+    4. They have provided a duration for the auction
+    5. They have specified whether they would like to set a reserve price
+
+#### `withdraw()`
+
+1. Allows a user to withdraw their owed assests for an auction if:
```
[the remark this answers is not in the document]
bitzoic: There are 4 different outcomes when calling this function.

1. You are the seller, no one bids, you receive your original deposit
2. You are the seller, there were bids, you receive the winning bid
3. You are the buyer, you won the auction, you receive the selling deposit
4. You are the buyer, you didn't win the auction, you receive your original deposit.

I've added an "or" as well as changed "deposit" to "balance".

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,170 @@
...
+pub enum Asset {
+    NFTAsset: NFTAsset,
+    TokenAsset: TokenAsset,
+}
+
+impl Asset {
+    pub fn amount(self) -> u64 {
+        match self {
+            Asset::NFTAsset(nft_asset) => {
+                // TODO: This should return the number of tokens when StorageVec in structs is supported
```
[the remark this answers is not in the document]
bitzoic: I would like to change the `NFTAsset` struct from this:

```
pub struct NFTAsset {
    contract_id: ContractId,
    token_id: u64,
}
```
to this:
```
pub struct NFTAsset {
    contract_id: ContractId,
    token_ids: StorageVec<u64>,
}
```
Then on `amount()` simple return the `len()` of the `StorageVec`.

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+impl core::ops::Add for Asset {
+    fn add(self, other: Self) -> Self {
+        match (self, other) {
+            (
+
+                Asset::TokenAsset(token_asset1),
+                Asset::TokenAsset(token_asset2),
```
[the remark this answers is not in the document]
bitzoic: Yes, yes it is. I'll see if I can make a minimal repro

auctions/english-auction/src/utils.sw
bitzoic: This used to create conflicts and fail to build. Looks like that has been resolved.

auctions/english-auction/src/utils.sw
bitzoic: This ensures that the NFT has been transferred correctly. If the transfer itself fails (the `from` doesn't own the NFT, the `to` is not approved, or the `to` is not an operator) then the `to` identity wouldn't own the NFT and revert.

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,170 @@
...
+            ) => {
+                require(token_asset1.contract_id == token_asset2.contract_id, AssetError::AssetsAreNotTheSame);
+                let token = TokenAsset {
+                    amount: token_asset1.amount + token_asset2.amount,
+                    contract_id: token_asset1.contract_id,
+                };
+                Asset::TokenAsset(token)
```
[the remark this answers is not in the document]
bitzoic: I'm not sure what you're referring to here

auctions/english-auction/src/data_structures.sw
```diff
@@ -0,0 +1,170 @@
...
+                token_asset1.amount < token_asset2.amount
+            },
+            _ => {
+                revert(0);
+            },
+        }
+    }
+}
+
+pub struct Auction {
```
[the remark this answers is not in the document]
bitzoic: How are we handling the combination of structs and `impl`? 

It is currently ordered alphabetically where structs are alphabetical with their implementations directly below that are themselves in alphabetical order.

auctions/english-auction/SPECIFICATION.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. Returns the total deposits owed to a user for a specific auction
+
+#### `total_auctions()`
+
+1. Returns the total number of auctions that have ever been created with the contract
+
+## Sequence Diagram
+
+![English Auction Sequence Diagram](.docs/english-auction-sequence-diagram.png)
```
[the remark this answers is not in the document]
Braqzen: The documentation is not limited to a single descriptive word, more words can be used or additional documentation can be written somewhere with a semantic reference to that section (i.e. instead of saying "more info over there" you'd structure the sentence to refer to that context).

auctions/english-auction/src/data_structures.sw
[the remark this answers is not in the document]
Braqzen: Instead of assigning the struct to a variable and passing that variable into the enum the struct instantiation can be done directly inside the enum which alleviates the need to allocate space for a variable.

auctions/english-auction/src/data_structures.sw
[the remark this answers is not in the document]
Braqzen: I was referring to the fields inside of the struct, for example the order of `end_block` and `highest_bidder`. Something useful to consider is converting this to a directory and having each type be a separate module inside the `data_structures` directory.

auctions/english-auction/tests/functions/auction_info.rs
```diff
@@ -0,0 +1,165 @@
...
+        )
+        .await;
+
+        let total_duration2 = provider.latest_block_height().await.unwrap() + duration;
+        let auction1 = auction_info(auction1_id, &seller.auction).await;
+        let auction2 = auction_info(auction2_id, &seller.auction).await;
+        assert!(auction1.is_some());
+        assert!(auction2.is_some());
+
+        let auction1 = auction1.unwrap();
```
matt-user, line 138: IMO it would be better to construct a second auction object and then just have one assert.  
1.) I think it looks cleaner
2.) If a new member/field is added to your auction object the test case will break

matt-user (approved):
lgtm, you might want to wait for another approval

auctions/english-auction/src/data_structures/auction.sw
```diff
@@ -0,0 +1,48 @@
...
+    ) -> Self {
+        Auction {
+            bid_asset,
+            end_block,
+            highest_bidder: Option::None(),
+            initial_price,
+            reserve_price,
+            sell_asset,
+            seller,
+            state: State::Open,
```
Braqzen, line 45: Do we want to automatically start a new auction or create one and at some point flip a switch to start it? This could be a future improvement.

auctions/english-auction/src/data_structures/nft_asset.sw
```diff
@@ -0,0 +1,61 @@
+library nft_asset;
+
+dep traits;
+
+use ::errors::AssetError;
+use traits::Asset;
+
+pub struct NFTAsset {
+    /// The `ContractId` of the NFT that the struct is representing.
+    contract_id: ContractId,
```
Braqzen, line 10: `contract_id` is a module so we need to be mindful of the prelude and possible namespace collisions / shadowing

auctions/english-auction/src/data_structures/nft_asset.sw
```diff
@@ -0,0 +1,61 @@
...
+    /// The token id of the NFT that the struct is representing.
+    token_id: u64,
```
Braqzen, line 12: `asset_id`?

auctions/english-auction/src/events.sw
```diff
@@ -0,0 +1,29 @@
...
+pub struct CreateAuctionEvent {
+    auction_id: u64,
+    bid_asset: AuctionAsset,
+    sell_asset: AuctionAsset,
+}
+
+pub struct BidEvent {
+    amount: u64,
+    auction_id: u64,
+    identity: Identity,
```
Braqzen, line 22: `user: Identity`?

Same below?

auctions/english-auction/src/interface.sw
```diff
@@ -0,0 +1,123 @@
...
+    /// * When the `auction_id` does not map to an existing auction.
+    /// * When the auction is no longer open.
+    /// * When the sender is not the seller of the auction.
+    #[storage(read, write)]
+    fn cancel(auction_id: u64);
+
+    /// Starts an auction with a seller, selling asset, accepted bid asset, initial price, a
+    /// possible reserve price, and duration of the auction.
+    ///
+    /// This function will return the the newly created auction's ID number which must be used to
```
Braqzen, line 58:
```suggestion
    /// This function will return the newly created auction's ID number which is used to
```

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,253 @@
+contract;
+
+dep data_structures/state;
+dep errors;
+dep events;
+dep interface;
+dep utils;
+
+// TODO: Move these into alphabetical order once https://github.com/FuelLabs/sway/issues/409 is resolved
```
Braqzen, line 9: aaay, the link looks correct here.

auctions/english-auction/src/main.sw
```diff
@@ -0,0 +1,253 @@
...
+use auction_asset::AuctionAsset;
+use auction::Auction;
+use errors::{AccessError, InitError, InputError, UserError};
+use events::{BidEvent, CancelAuctionEvent, CreateAuctionEvent, WithdrawEvent};
+use interface::{EnglishAuction, NFT};
+use state::State;
+use std::{
+    block::height,
+    chain::auth::{
+        AuthError,
```
Braqzen, line 22: Doesn't seem to be used, is this needed? I may have asked this already - or in another PR.

auctions/english-auction/tests/utils/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+            TxParameters::default(),
+            StorageConfiguration::with_storage_path(Some(
+                "./out/debug/english-auction-storage_slots.json".to_string(),
+            )),
+        )
+        .await
+        .unwrap();
+
+        let sell_asset_id = Contract::deploy(
+            "./tests/artifacts/asset/out/debug/asset.bin",
```
Braqzen, line 348: It may be beneficial to factor out these hardcoded paths to global variables (or some local module) and use them as variables in case things change in the future. These could be defined somewhere towards the top of the file.

auctions/english-auction/tests/utils/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+        };
+
+        let seller = Metadata {
+            asset: MyAsset::new(sell_asset_id.to_string(), wallet2.clone()),
+            auction: EnglishAuction::new(auction_id.to_string(), wallet2.clone()),
+            nft: Nft::new(sell_nft_id.to_string(), wallet2.clone()),
+            wallet: wallet2.clone(),
+        };
+
+        let rng = &mut StdRng::seed_from_u64(2322u64);
```
Braqzen, line 381: I was wondering why `rand` is added as a dependency. In this case a seed is being used to generate a salt so the salt should always be the same. Is there a reason for the magic number or is it purely magic? Can that be added as comments?

I generally advise against the usage of randomness in tests because of the boundary conditions and unexpected failures that may happen. I'd prefer a hardcoded value however the seed will constrain the randomness so it comes down to whether we want that additional dependency and compilation time.

auctions/english-auction/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "english-auction"
+version = "0.0.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = { version = "0.25", features = ["fuel-core-lib"] }
+rand = {version = "0.8" }
```
Braqzen, line 10:
```suggestion
rand = { version = "0.8" }
```

Comments on rand somewhere below

Braqzen (commented):
1. I have not validated the logic / control flow / coverage of tests however things look reasonable
2. I am convinced that this contract will need to be split into 4 _in the future_. There is a lot going on here and I think it will be easier to maintain each sub-auction and it will be clearer to the reader when looking at the individual auctions that are available. The 4 auctions would be where the seller gets to specify an asset (NFT or otherwise) and for each of those two types they can either bid with an NFT or some asset. If we have clear examples then users may take the concept and apply whatever changes they need in their production version.

auctions/english-auction/tests/utils/mod.rs
matt-user: check @supiket AMM pr for an example

matt-user: https://github.com/FuelLabs/sway-applications/pull/267/files#diff-c93260d3df375e9c19479a07a4acaf77697d238c68391f46cd23b3a1348d8901

auctions/english-auction/src/data_structures/auction.sw
bitzoic: Considering that the user must currently provide an asset to create an auction, requiring only one call to create and start rather than 2 saves gas. 

But this may be something to consider during the implementation of the front end.

auctions/english-auction/tests/utils/mod.rs
bitzoic: I'm following the [Rust SDK Book](https://fuellabs.github.io/fuels-rs/v0.26.0/getting-started/contracts.html) here.

auctions/english-auction/SPECIFICATION.md
```diff
@@ -0,0 +1,79 @@
...
+- [Use Cases](#use-cases)
+    - [Actions that users are able to perform](#actions-that-users-are-able-to-perform)
+        - [English Auction Core Functionality](#english-auction-core-functionality)
+            - [`bid()`](#bid)
+            - [`cancel()`](#cancel)
+            - [`create()`](#create)
+            - [`withdraw()`](#withdraw)
+        - [English Auction State Checks](#english-auction-state-checks)
+            - [`auction_info()`](#auction_info)
+            - [`deposit()`](#deposit)
```
Braqzen, line 12:
```suggestion
            - [`deposit_balance()`](#deposit-balance)
```

Braqzen (approved):

PR #104, opened 2022-06-22
Full test coverage for the basic section of the Multisig functionality

### LINKED ISSUE #7: Full test coverage
All of the features in [The Basics](https://github.com/FuelLabs/sway-applications/issues/1) should be fully tested.

- [x] #2 
- [x] #3 
- [x] #4 
- [x] #5 
- [x] #6

Braqzen:
## Type of change

- New feature

## Changes

- TODO

## Notes

- Awaiting Vec support in SDK in order to test

## Related Issues

Closes #7


multisig-wallet/src/main.sw
```diff
@@ -46,28 +44,39 @@ storage {
...
-        while user_index < 25 {
-            require(~Address::from(BASE_ASSET_ID) != users[user_index].identity, InitError::AddressCannotBeZero);
-            require(users[user_index].weight != 0, InitError::WeightingCannotBeZero);
-            storage.weighting.insert(users[user_index].identity, users[user_index].weight);
+        while user_index < users.len() {
+            // Workaround
+            let user: Option<User> = users.get(user_index);
+            let user = user.unwrap();
+
+            require(BASE_ASSET_ID != user.identity.value, InitError::AddressCannotBeZero);
```
nfurfaro, line 74: Super minor nit here, but seeing BASE_ASSET_ID first in the comparison puts the emphasis in it, rather than on the `.value` field of the user. Reversing these to be
 `require(user.identity.value != BASE_ASSET_ID, InitError::AddressCannotBeZero);`
reads a bit better, but maybe it's just me.

Also, I get why you used BASE_ASSET_ID here, but creating a local const with a better name _might_ be a bit clearer, i.e:
`const ZERO_ADDRESS = BASE_ASSET_ID`

nfurfaro: Yeah, that should probably be added. QQ: If I add an address to `storage`, ie:
```
storage {
    user: Address,
}
```
Does `user` start as and `Address` type with it's inner value `0x000...000`?

nfurfaro: I know we've had this discussion around preferring to use the `Option` type over the ZERO address, where `None` would replace the usage of ZERO_ADDRESS in comparisons.

Braqzen: From memory I cannot confirm since I do not recall if I have ever explicitly tested that however all `b256` values are initialized to 0x00... therefore that _should_ naturally happen to the address inner value.

K1-R1:
This PR can be closed once #272 is merged, as it includes the addition of full test coverage for the Multisig-wallet.

simonr0204:
If the other PR includes the changes, then this one will never be merged, right? We should close it, if that's the case.

PR #125, opened 2022-07-01
Bumped fundraiser and multisig to v0.17

Braqzen:
## Type of change

## Changes

- Fundraiser & Multisig
    - Renamed abi library to contract_abi
    - Adjusted usage accordingly to new name
- Multisig
    - Renamed contract to address
    - Changed import from BASE_ASSET_ID (which is now a ContractId type) to ZERO_B256

## Notes

- ABI is now a forbidden keyword
- contract is now a forbidden keyword

### DIFF
--- a/multisig-wallet/src/main.sw
+++ b/multisig-wallet/src/main.sw
@@ -5,7 +5,7 @@ contract;
 //      - change the "data" in the Tx hashing from b256 to vec
 
 // Our library dependencies
-dep abi;
+dep contract_abi;
 dep data_structures;
 dep errors;
 dep events;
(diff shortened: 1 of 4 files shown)


multisig-wallet/src/main.sw
```diff
@@ -124,7 +124,7 @@ impl MultiSignatureWallet for Contract {
         storage.nonce = storage.nonce + 1;
 
         match to {
-            Identity::Address(address) => transfer_to_output(value, asset_id, address), Identity::ContractId(contract) => force_transfer_to_contract(value, asset_id, contract), 
+            Identity::Address(address) => transfer_to_output(value, asset_id, address), Identity::ContractId(address) => force_transfer_to_contract(value, asset_id, address),
```
matt-user, line 127: You can flatten this with the generic transfer function from the std, but that can be handled in another pr.

matt-user (approved):

multisig-wallet/src/main.sw
Braqzen: Bumping versions should not be equated to functional upgrades or code quality. It's just a way to make the other projects work as expected.

PR #140, opened 2022-07-05
Refine CONTRIBUTING.md to include a clear section for issues and pull requests. Add a code quality section

### LINKED ISSUE #130: Refine CONTRIBUTING.md to include a clear section for issues and pull requests. Add a code quality section
#### Description

Extend the documentation in [CONTRIBUTING.md](https://github.com/FuelLabs/sway-applications/blob/master/CONTRIBUTING.md) to highlight code quality practices.

#### Motivation

Everyone has a unique coding style however that is a bit too archaic when trying to maintain code.
Standardizing the codying style sets expectations when reviewing code and makes it a lot easier to catch mistakes.
Moreover, it makes it easier to decide on improvements to the standard when they are explicitly written out.

The documentation should include the following

- Always order everything in alphabetical order - imports, contract declarations, data structures and fields inside them
    - There may be two exceptions:
        - When naming cannot be altered to be reasonable e.g. the first field in an enum is the default and forcing a name may not be a good idea for readability purposes just so that the order is alphabeticized
        - It makes sense to group functions in a contract in a certain way e.g. pure / read only getters at the bottom while other functions may be closer to the top. That may need to be considered as its own standard / copied from Ethereum in the near future
- Split up the code into its own modules e.g.
    - A file just for abi declarations (`contract_abi`)
    - A file just for structs that are treated as events (`events.sw`)
    - A file just for custom data structures defined by the user (`data_structures.sw`)
    - A file just for private utility functions that the contract uses to keep the code DRY (`utils.sw`)
    - A file just for enum errors (`errors.sw`)
- Document your functions!
    - Documentation uses `///` while `//` is for comments. Reference [this guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html)
    - Make sure to document the arguments and when your functions revert
    - Make sure that the order of your reverts matches the order in which you document your reverts e.g. if you have 2 revert statements in your function your documentation should first mention the first revert that can happen and then the last revert in that order
    - Similarly, your arguments should be in alphabetical order and mentioned in the same order
- Testing
    - When testing split up your code into modules so that it is easy to see which cases are meant to succeed and which are meant to fail. [Take a look at the existing projects to see what I mean]
    - Each module should be called the same thing as the ABI function so that they are easy to find and they should be defined in the file in the same order as in your ABI.
    - Create a `utils` module which wraps your abi calls so that you do not repeat the same long line hundreds of times. It's easier to declare a wrapper in 1 place and pass in data
- Note that you should not leave in any imports, functions, tests etc that are commented out / do not work. It might be the case that Sway does not fully support your feature in which case leaving a TODO comment at the top of the file to explain why something is there / needs to be done is fine however if you fails to write some code then just delete the entire feature since it doesn't work anyway

Braqzen:
## Type of change

- Documentation

## Changes

- Changed the `issues` and `pull requests` sections to contain a lot more info
- Added a `code quality` section which contains multiple subsections

## Notes

- Will definitely rewrite testing in a future PR since it's a mess but it's enough for an initial section

## Related Issues

Closes #130

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -297,9 +297,37 @@ As you can see, a single question spawns more questions which can spawn more tec
 
 ### Testing
 
-TODO
+There's a lot that can be said about testing so it's difficult to sum it all up in a few paragraphs.
 
-What is your test coverage like?
+Here are some pointers:
+
+- Order your modules alphabetically
+- Split your code up into two modules
+  - The tests themselves
+    - It's easier to navigate a file for tests when it only consists of the tests
+  - All of the utility (helper) functions used for testing and keeping your tests DRY (do not repeat yourself)
+    - Split the utilities into two modules
+      - 1 module for your custom code such as creating wallets, generating setup parameters etc.
+      - 1 module which wraps your abi calls
+        - If you have an abi function `create_game(player_two: Player, player_one: Player) -> Game` then you should have a function which takes appropriate parameters and makes a call to `contract.create_game(player_two, player_one).call().await.unwrap()`
+        - This allows you to call the same abi wrapper function by passing in the arguments instead of repeating any additional setup that you have in order to make the call
+        - Note that this should only call your contract and return data rather than have complex code such as a `setup()` might have
+- You should have a test module inside your test file for each abi call that can be made
+  - Each test module should be split into 2
+    - 1 `success` case where all of your tests work as expected
+    - 1 `revert` case where you intentionally test your functions in order to make them break
+- Ensure that your code coverage covers all possible cases
+  - Check boundary conditions
+  - Assertions
+  - Calling things out of order
+- When writing a test make sure to perform a "before" and "after" check to ensure that your code works as expected
+  - Most people only check the "after" assertions however this may lead to your test passing falsely because
+    - There is a bug in your test
+    - There is a bug in your implementation
+    - Your test and implementation may be correct however there is a bug in the code you are using
+- The tests do not need to be in alphabetical order and instead they should be sequentially ordered in the way the code operates from start to finish
+  - If you have 3 assertions to test then the first assertion should be at the top, the next test should be the next assertion and the last test should be the last assertion
+- As long as you structure your code appropriately and name your tests well (reference the earlier sections) then you do not need to document your tests at all but comments are always nice when they are useful
 
 ---
 ## Pull Requests
@@ -355,3 +383,5 @@ You can
   - GitHub has some special keywords referenced [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) which should be used in addition to the #123 syntax
 
 The reason is that this allows the reviewer to quickly find the issue and check to see what work needed to be done and compare that against the pull request.
+
+If your work has been accepted and you have an option to "squash" your commits then delete all the commit messages and summarize your work into a single commit message.


CONTRIBUTING.md
```diff
@@ -1,23 +1,239 @@
...
+
+To be explicit
+
+- All dependencies must be declared in alphabetical order just below the declaration of the program type e.g. below `contract;`
+- All imports must be in alphabetical order and any modules within those imports must be in alphabetical order
+- All functions, structs, enums, fields inside structs and enums, parameters of a function, parameters in documentation etc. must be in alphabetical order in all files (this does not apply to variables inside a function)
+
+Example:
+
+```rust
```
bitzoic, line 80: I think it would be smart to have this in an external file and included in CI, much like how examples are written in the sway book

CONTRIBUTING.md
```diff
@@ -1,23 +1,239 @@
...
+
+There are two ways to document your code
+
+- Using `///` for official documentation
+  - This allows tools to be used and parse out your documentation automatically
+- Using `//` for comments
+  - These are helper comments within the code to guide the developer who is reading your code
+  - They do not get parsed out via automatic doc generation
+
+Each function that is written should be documented (the `///`). The only exception applies to the functions inside the `impl` of your contract (notice that the example above does not have an documentation, that would be on the `abi` that is imported). The documentation for those functions should be on the `abi` because a developer will have access to your `abi` and not necessarily your implementation.
```
bitzoic, line 138:
```suggestion
Each function that is written should be documented (the `///`). The only exception applies to the functions inside the `impl` of your contract (notice that the example above does not have any documentation, that would be on the `abi` that is imported). The documentation for those functions should be on the `abi` because a developer will have access to your `abi` and not necessarily your implementation.
```

CONTRIBUTING.md
```diff
@@ -1,23 +1,239 @@
...
+
+When writing documentation make sure that your arguments are in alphabetical order and any assertion that you have in your function is also documented in the same order. It makes it easier to read the code when you see everything in sequential order rather than searching for where an assertion is placed in the function.
```
bitzoic, line 140:
```suggestion
When writing documentation make sure that your arguments are in alphabetical order and any assertions that you have in your function are also documented in the same order. It makes it easier to read the code when you see everything in sequential order rather than searching for where an assertion is placed in the function.
```

CONTRIBUTING.md
```diff
@@ -1,23 +1,239 @@
...
+pub struct Item {
+   /// Unique identifier used to retrieve the item from a vector of items held in storage
+   id: u64,
+
+   /// The number of remaining items left in production
+   quantity: u64,
+}
+```
+
+Documenting your code does not stop at writing documentation. It is also important to name your variables, functions and data structures appropriately. This is a difficult task and it can be argued to be a skill and an art because if your name is too verbose then it will be difficult to read (because of how much screenspace it takes up) and annoying to use while an abbreviated name requires insider knowledge to be able to infer what the variable is (you could document it extensively however that still means the reader must remember what the documentation says).
```
bitzoic, line 204:
```suggestion
Documenting your code does not stop at writing documentation. It is also important to name your variables, functions and data structures appropriately. This is a difficult task and it can be argued to be a skill and an art. If your name is too verbose, then it will be difficult to read (because of how much screenspace it takes up) and annoying to use. On the other hand, an abbreviated name requires insider knowledge to be able to infer what the variable is (you could document it extensively however that still means the reader must remember what the documentation says).
```

CONTRIBUTING.md
```diff
@@ -1,23 +1,239 @@
...
+
+There are two general rules to follow
```
bitzoic, line 206:
```suggestion
There are two general rules to follow:
```

CONTRIBUTING.md
Braqzen: :) Probably, I did write this off-hand without checking if it even compiles

adlerjohn: Unfortunately, plain markdown doesn't support that kind of include. We could consider making a Sway Applications Book though!

Braqzen: This warrants a separate issue. Will be done in #141

adlerjohn (approved):

Issue #120, opened 2022-06-29
Flash Loan Application 


bitzoic:
#### Description

A Flash Loan is a special kind of loan that does not require any collateral. Traditionally in Defi, loans are over-collateralized to ensure that the borrower will return the funds. A Flash Loan on the other hand takes place in a single transaction. 

When a user requests for a Flash Loan, the funds are sent to the smart contract that requested those funds. Once the funds have been sent, the loan provider called a function by the borrower to execute the transaction. Once the borrower has completed their execution the funds are returned to the lender with a small fee. If at any point the execution fails, the transaction is reverted and the funds are returned. 

### Features

- [x] #115 
- [x] #119

PR #142, opened 2022-07-06
OTC swap predicate

### LINKED ISSUE #170: OTC Swap predicate
### Motivation

The swap mechanism requires a predicate which constrains the spending transaction's outputs to satisfy the conditions of the swap.

simonr0204:
## Type of change

- New feature / new application

## Changes

- Adds a new application showcasing predicates in Fuel - an OTC swap.

## Notes

- The predicate itself is really the entire "application". But I think it would be good to get a simple example of using predicates in here as many (including smart contract developers coming from Ethereum) are unfamiliar with the kinds of new designs they open up.

Closes #170


simonr0204:
Currently specifying a github url for std dependency while waiting for next `forc` release

otc-swap-predicate/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+fuel-core = { version = "0.9", default-features = false }
+fuel-crypto = "0.5"
+
+fuel-gql-client = { version = "0.9", default-features = false }
+fuels = { version = "0.16", features = ["fuel-core-lib"] }
+fuels-abigen-macro = "0.16"
+fuels-contract = { version = "0.16" }
+fuels-signers = { version = "0.16" }
```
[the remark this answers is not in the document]
simonr0204: I think `fuels` re-exports _parts_ of `fuels-contract` but not all of it - e.g. `script::Script`. I was able to remove the others as I'm not using them anyway.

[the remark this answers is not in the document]
simonr0204: ah never mind, found it !

`use fuels::contract::script::Script;`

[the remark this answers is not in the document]
simonr0204: Resolved [ca97658](https://github.com/FuelLabs/sway-applications/pull/142/commits/ca9765844a3b264426c8f7d5b99dfea695b92c7d)

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,66 @@
...
+    contract_id::ContractId,
+    option::Option,
+    result::Result,
+    revert::revert,
+    tx::{b256_from_pointer_offset, tx_input_owner, tx_output_amount, tx_output_pointer}
+};
+
+/// Order / OTC swap Predicate
+///
+/// This predicate root serves as an "order" that anyone can fill
```
adlerjohn, line 16:
```suggestion
/// This predicate serves as an "order" that anyone can fill.
```

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,66 @@
...
+/// The coin sent to the predicate root can be unlocked by any transaction which has an output which meets the conditions of the order
+/// The order maker can "cancel" the order by spending the predicate's coins in any transaction containing an input they have signed
```
adlerjohn, line 18:
```suggestion
/// The order maker can "cancel" the order by spending the predicate's coins in any transaction containing only coin inputs they own
```

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,66 @@
...
+///
+/// Limitations:
+///    - An order can not be partially filled - the taker must pay the entire ask amount
+///    - There is no on-chain matching engine, so an order placed 'offside' would not be matched with an existing order with a better price (on the contrary, it would be vulnerable to arbitrage)
```
adlerjohn, line 22:
```suggestion
///    - There is no on-chain matching engine, so an order placed "offside" would not be matched with an existing order with a better price (on the contrary, it would be vulnerable to arbitrage)
```

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,66 @@
...
+/// - `output_index` - The index of the Coin output which pays the order maker.
+///
+fn main(input_index: u8, output_index: u8) -> bool {
+    // Order conditions: This must be hardcoded here.
+    // The spending transaction must have an output that sends `ask_amount` of `ask_token` to `maker`
+    let maker = ~Address::from(0x0303030303030303030303030303030303030303030303030303030303030303);
+    let ask_amount = 42;
+    let ask_token = BASE_ASSET_ID;
+
+    // First, check if the transaction contains an input coin from the maker, to cancel their own order:
```
adlerjohn, line 38: I think it might be dangerous to only use a single owner instead of `msg_sender()`. Especially with https://github.com/FuelLabs/fuel-specs/issues/40, it would allow anyone to attach such a predicate to any partially-signed tx and potentially burn the funds if there's no change output for the given asset.

```suggestion
    // First, check if the transaction contains only input coins from the maker, to cancel their own order:
```

otc-swap-predicate/tests/harness.rs
```diff
@@ -0,0 +1,145 @@
...
+    wallet.sign_transaction(&mut tx).await.unwrap();
+    let script = Script::new(tx);
+    let _receipts = script.call(&client).await.unwrap();
+
+    let predicate_balance = utils::get_balance(&provider, predicate_root, base_asset).await;
+    let wallet_balance = utils::get_balance(&provider, wallet.address(), base_asset).await;
+    let receiver_balance = utils::get_balance(&provider, receiver_address, base_asset).await;
+
+    // The predicate root's coin has been spent
+    assert_eq!(predicate_balance, 0);
```
adlerjohn, line 135: It would be nice to see a negative test or two, that the coin isn't spendable if the conditions are violated.

otc-swap-predicate/src/main.sw
simonr0204: I don't think we can use `msg_sender()` because of :
`^^^^^^^^ The GM (get-metadata) opcode, when called from an external context, will cause the VM to panic.`

otc-swap-predicate/tests/harness.rs
simonr0204: hmm is there currently a way to test this? Since the transaction isn't reverting, but rather invalid :

`ContractCallError("Response errors; Invalid transaction: The transaction contains a predicate which failed to validate`

Braqzen (changes requested):
1. hm, I don't understand why this built without being added to [build.sh](https://github.com/FuelLabs/sway-applications/blob/master/.github/scripts/build.sh). That script has conditional checks for projects in the matrix otherwise it should do nothing. When I was writing the script it did not build a project unless it matched one of those conditionals but now it does. Any idea why this builds @adlerjohn ?
2. This repo is meant to be maintained rather than have releases therefore in all the projects we set the "version" to be "0.0.0". Could you update your version?
3. Are the dependencies in `Cargo.toml` for `core` & `client` required i.e. `fuels` does not import them?
4. The purpose of the `utils` module is to move as many imports and utility functions in there so that the `harness.rs` file only contains tests. Following on from the suggestion by @adlerjohn could you split the `harness.rs` into modules where there is a distinction between the successful and reverting cases while factoring out the duplicated setup into the `utils` module? More information can be found [here](https://github.com/FuelLabs/sway-applications/blob/master/CONTRIBUTING.md#testing).
5. Could you also replicate the README.md similar to e.g. the [fundraiser](https://raw.githubusercontent.com/FuelLabs/sway-applications/master/fundraiser/README.md)? It currently doesn't have much info since it's not a proper app yet however it provides a template to fill in later. It also has a logo which changes based on the theme the user has set (light/dark)
6. Could you also replicate an "application" issue similar to e.g. the [flashloan](https://github.com/FuelLabs/sway-applications/issues/120) and link the contract subtask to this issue? It's helpful to have some context in the issue rather than buried as documentation in the contract itself

otc-swap-predicate/src/main.sw
simonr0204: Resolved [20f843d](https://github.com/FuelLabs/sway-applications/pull/142/commits/20f843d164f786d2f7976dd1d64f08252a31ff46)

simonr0204:
"build" is just the name of the CI run - it goes through each of the `steps`. So here it is executing the script but just stepping over my app as it doesn't have a match. I'll add it there.

Is there a template for the readme logo or do I have to get my MS-Paint skillz out ?

Braqzen:
> "build" is just the name of the CI run - it goes through each of the `steps` for each project in the matrix. So here it is executing the script but just stepping over my app as it doesn't have a match. I'll add it there.
> 
> Is there a template for the readme logo or do I have to get my MS-Paint skillz out ?

I was under the impression that the call to `build.sh` with the matrix arg made the project compile and I thought that I saw it successfully run the test which implies it did build it. Perhaps I was seeing things, I cannot handle hot weather.

My approach to the logo is to crop the text from any of the apps and add a text box with either black / white text on a transparent background. I use [gimp](https://www.gimp.org/) for image manipulation. If you do not want to change the logo itself then renaming the files and using the [sway logos](https://github.com/FuelLabs/sway-applications/tree/master/.docs) (or any other app logo) is fine. I'll adjust the logos later.

simonr0204:
> it successfully run the test which implies it did build it.

Yeah this is because `forc test` _also_ builds, so that's how the test was able to run. Maybe it's a bit misleading for the whole CI process (carried out for each project) to be called "build" though, as it also formats, tests, etc.

Braqzen:
> > it successfully run the test which implies it did build it.
> 
> Yeah this is because `forc test` _also_ builds, so that's how the test was able to run. Maybe it's a bit misleading for the whole CI process (carried out for each project) to be called "build" though, since whereas it does do that (via the script) it also formats, tests, etc.

That makes sense. So this wasn't built until the test and unlike some other projects, which have a different structure and require manual building of the token, this continued happily because there wasn't any strange path that wasn't built.

otc-swap-predicate/src/main.sw
simonr0204: What do you think @adlerjohn ?

simonr0204: ping @adlerjohn

simonr0204: See [ff3591f](https://github.com/FuelLabs/sway-applications/pull/142/commits/ff3591f055672715c4a415c26f2e726f69da40e4)

simonr0204:
converting to draft as SDK is now ready with custom assets so can implement more realistic tests

simonr0204:
update: just waiting for SDK 0.21 to mark for final review

Braqzen:
Nudging since SDK is on version 0.21. A newer release should be coming out soon too.

1. Could you update the readme to follow the updated structure in the projects that are sitting in master?
2. Could you separate the tests into 2 modules: success and revert? It should be in the same file. It's easier to distinguish at a glance where the successful cases are vs the reverts.
3. Could you add some form of a spec similar to the escrow / DAO (in master)? Since it's just 1 function I'm unsure of what should be added but I think some sort of non-technical flow of information can be presented for example 1 success case and multiple branches indicating each assertion failure

simonr0204:
Just realized this is still unmerged!

@Braqzen :
I've bumped to latest versions, and split tests into passing and reverting as suggested. I've also updated the readme to reflect the same format as the other apps.

For the specification, I think the readme really contains everything there is to know about the predicate. Whereas other apps have various entry points / different categories of user, the predicate is very simple. 

For the logos, despite my best efforts I can't get the "dark" one to render with white text on github. On my local the text is white. Not sure what to do about that.

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+
+/// Order / OTC swap Predicate
+fn main() -> bool {
+    // Order conditions: These are set in Forc.toml
+    // The spending transaction must have an output that sends `ask_amount` of `ask_token` to `receiver`
+    // Conversion to ContractId and Address types will be unnecessary once
+    // https://github.com/FuelLabs/sway/issues/2647 is fixed
+    let ask_token: ContractId = ContractId {
+        value: ask_token_config,
+    };
```
Braqzen, line 28:
```suggestion
    let ask_token = ContractId {
        value: ask_token_config,
    };
```

No need to be explicit here since the name of the struct is right there

otc-swap-predicate/src/main.sw
```diff
@@ -0,0 +1,63 @@
...
+    if (input_count() == 2u8) {
+        let owner = match input_owner(0) {
+            Option::Some(owner) => owner,
+            _ => input_owner(1).unwrap(),
+        };
+
+        if (owner == receiver) {
+            return true;
+        };
+    };
```
Braqzen, line 42:
```suggestion
    if input_count() == 2u8 {
        let owner = match input_owner(0) {
            Option::Some(owner) => owner,
            _ => input_owner(1).unwrap(),
        };

        if owner == receiver {
            return true;
        };
    };
```

Not much going on here so I don't believe that wrapping with parenthesis helps readability.

Noting that the `2u8` is padded to `u64` so I'm unsure what actually happens here. Hopefully  the tests are sufficient and document this in case this gets fixed in the VM in the future

Braqzen (commented):
1. Could you change the lower case config constants in the manifest file to be all capitalized?
2. I see that there are a few "TO<space>DO" comments in the contract, could you change them to be
    - "TODO" (I think most people use this so using a global search would make this come up)
    - Document any relevant work in the application progress doc?
3. logo is probably fine, github cashes things and iirc I also had issues in the PR itself so I assume it will work on master if the color of the text and the name of the file match
4. Does it make sense to have tests that attempt to reuse the predicate i.e. if the owner recovers tokens once but then they try again - is that possible; should that be tested etc?
5. Noting that the utils `mod` helpers are kinda large. They have a lot going on. I'm hoping that I'm never going to be the one having to debug them
6. Perhaps "Predicates in Fuel" needs to be removed in the future and added to some official documentation for users rather than in a specific app
7. Some more points about the README
    - There is an "Overview" which is empty. Can this be filled in or removed if you think the following sections cover the content sufficiently?
    - I'm wondering if the "Order / OTC" section should use additional heading tags to indicate when "taking", "cancelling" etc. applies. It would be a little clearer with a heading and it would break up the "large" section of text. There isn't much text there and if headings are added then having a sentence or two may or may not be that helpful so it may even need a little more text. Unsure, thoughts?
    - Any particular reason for the change to the project structure? The other apps copy and paste a block for visual purposes
    - Could you tell the reader what the "root" of the project is? If they are to run the commands then it would be beneficial to explicitly provide "commands" as in a declaration of what must be done and also a literal command to show them how to execute that. Moreover, the commands to build and test should use code-blocks since they are not used as part of a sentence and instead they are used as code examples

otc-swap-predicate/src/main.sw
simonr0204: resolved in [c0a5fc1](https://github.com/FuelLabs/sway-applications/pull/142/commits/c0a5fc1c00ce18d3fafca269cb4e0fac78b2837d)

otc-swap-predicate/src/main.sw
simonr0204: resolved in [c0a5fc1](https://github.com/FuelLabs/sway-applications/pull/142/commits/c0a5fc1c00ce18d3fafca269cb4e0fac78b2837d)

`input_count()` returns a `u8` so this annotation avoids the "lose precision" compiler warning

simonr0204:
Addressed the last batch of comments, except:

4. Probably not? Testing against a double spend is out of scope for applications
5. I think it's not so bad - it's well commented. Unfortunately spending custom predicates is a bit messy. I could "artificially" factor it into functions for each step but to me, that's the same thing with added jumps for the reader.
6. Possibly. I think since for now this is the only example of using a predicate, having that reminder front and center is not a bad idea.
7. The `# Overview` section contains `## Predicates in Fuel` and `## Order / OTC swap Predicate` subsections

APPLICATION_PROGRESS.md
```diff
@@ -422,16 +422,17 @@ the sdk
 <details>
 <summary>Over the counter (OTC) Swap Predicate</summary>
 
-<h3>Contracts</h3>
+<h3>Contracts✅</h3>
 
-<h3>User Interface</h3>
+<h3>User Interface❌</h3>
```
Braqzen, line 427:
```suggestion
<h3>User Interface</h3>
```
If something has yet to be added then I wouldn't label it with a cross. I would only do that if the feature exists but there are significant enough issues that it should be marked as "not there yet" as opposed to "good enough for now / complete"

APPLICATION_PROGRESS.md
```diff
@@ -422,16 +422,17 @@ the sdk
 <details>
 <summary>Over the counter (OTC) Swap Predicate</summary>
 
-<h3>Contracts</h3>
+<h3>Contracts✅</h3>
```
Braqzen, line 425: Given the comment below I would mention the todo in config time constants but probably leave this as a check mark since that seems to be more code quality than functionality

APPLICATION_PROGRESS.md
simonr0204: Yeah I just pushed a new commit. Added ticks where feature is unimplemented but there is no intention for it to be.

APPLICATION_PROGRESS.md
```diff
@@ -422,17 +422,25 @@ the sdk
...
 
-<h3>User Interface</h3>
+<h3>User Interface✅</h3>
 
+- Will not exist for this predicate. A larger predicate based exchange could implement a UI.
```
Braqzen, line 429:
```suggestion
- Will not exist for this predicate. A larger predicate based exchange could implement a UI.

```

Approving after this

Braqzen (approved):

PR #160, opened 2022-07-11
Airdrop Smart Contract

### LINKED ISSUE #121: Airdrop Smart Contract
#### Description

An airdrop gives the ability for projects to distribute tokens. The airdrop contract should be a template for those who wish to create an airdrop with their own native assets. 

#### Motivation

Users should be able to deploy airdrops of their own Native Asset tokens.

#### Specification

- A Merkle Proof should be created to verify distribution 
- Users should be able to claim an airdrop for native assets

### LINKED ISSUE #122: Airdrop Documentation
#### Description
The Airdrop Application will need documentation for the user.

This includes

- README.md
     - What the application is, how to run it etc.
- SPECIFICATION.md
     - General overview of the application use cases, contract design, UI design etc.

Any additional informative files.

bitzoic:
## Type of change

- New feature
- Documentation

## Changes

- Add Airdrop Distributor contract
- Add Airdrop Distributor interface
- Add Airdrop Distributor events
- Add Airdrop Distributor errors
- Add Simple Token example contract
- Add Simple Token example interface
- Add Simple Token example errors

## Notes

- This is a basic implementation of an airdrop with merkle proof verification. 
- Includes a simple token example that has the corresponding `mint_to` function implemented. 
- Implements the Binary Merkle Proof library in Sway-Libs 

## Related Issues

Closes #121 
Closes #122


airdrop/airdrop-distributor/src/utils.sw
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// This function will verify the merkle leaf and root given against the proof.
+///
+/// # Arguments
+///
+/// `merkle_leaf` - A hash of a (user, amount) tuple representing the allotted tokens to a user.
+/// `merkle_root` - The computed merkle root given at the initialization of the airdrop contract.
+/// `proof` - The merkle proof that will be used to verify the leaf and root hashes.
+pub fn verify_merkle_proof(merkle_leaf: b256, merkle_root: b256, proof: Vec<b256>) -> bool {
```
adlerjohn, line 34: I feel like this function is an example of a thing that would be better served in e.g. a `sway-libs` repo of libraries rather than applications, for libraries that don't belong in the stdlib. Do you think now would be a good time to start splitting up library code that's independently useful from their applications?

airdrop/airdrop-distributor/src/utils.sw
```diff
@@ -0,0 +1,60 @@
...
+    let mut computed_hash = merkle_leaf;
+    let mut index = 0;
+    let proof_length = proof.len();
+
+    // Itterate over proof
+    while index < proof_length {
+        // Get the current element in the proof
+        let proof_element: Option<b256> = proof.get(index);
+        let proof_element = match proof_element {
+            Option::Some(b256) => proof_element.unwrap(), Option::None(b256) => revert(0),
```
adlerjohn, line 44: Until fmt-v2 is out, you can avoid the messy formatting here by enclosing the bodies of the match arms with `{ }`, e.g.

```rust
Option::Some(b256) => { proof_element.unwrap() }, Option::None(b256) => { revert(0) },
```

fmt-v1 will at least format that okay-ish.

airdrop/airdrop-distributor/src/utils.sw
```diff
@@ -0,0 +1,60 @@
...
+            Option::Some(b256) => proof_element.unwrap(), Option::None(b256) => revert(0), 
+        };
+
+        if computed_hash <= proof_element {
+            // Hash(current computed hash + current element of the proof)
+            computed_hash = sha256((computed_hash, proof_element));
+        } else {
+            // Hash(current element of the proof + current computed hash)
+            computed_hash = sha256((proof_element, computed_hash));
+        }
```
adlerjohn, line 53: Thinking out loud, but would it be better to use a `match` for control flow here? With `if` and `else` and `else if`, it's possible to be non-exhaustive and the compiler won't catch it.

bitzoic:
CI will fail until #209 is merged.

airdrop/simple-token/tests/functions/mint_to.rs
```diff
@@ -0,0 +1,145 @@
...
+        assert_eq!(
+            deployer
+                .wallet
+                .get_asset_balance(&AssetId::new(*deployer.asset_id))
+                .await
+                .unwrap(),
+            10
+        );
+
+        let identity2 = Identity::Address(wallet2.wallet.address().into());
```
Braqzen, line 65: I would try to keep all the declarations in 1 spot at the top so that the reader does not have to browse through the test to see where something is declared

airdrop/simple-token/tests/functions/mint_to.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    #[tokio::test]
+    #[should_panic(expected = "Revert(42)")]
+    async fn panics_when_mint_more_than_supply() {
+        let (deployer, _) = setup().await;
+
+        let identity = Identity::Address(deployer.wallet.address().into());
+        constructor(identity.clone(), &deployer.simple_token, 100).await;
+
+        mint_to(101, &deployer.simple_token, identity.clone()).await;
```
Braqzen, line 143: I would also refactor the total supply to a variable, possibly in the `setup`, so that you can change it in one place and do things like `mint(total_supply + 1, ...)`. This makes the tests less brittle

airdrop/airdrop-distributor/tests/functions/constructor.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        airdrop_constructor(
+            10,
+            &deploy_wallet.airdrop_distributor,
+            [1u8; 32],
+            asset.asset_id,
+        )
+        .await;
+
+        // TODO: Get block height and add 10
```
Braqzen, line 28: [Block manipulation](https://fuellabs.github.io/fuels-rs/v0.22.0/testing/chains.html) is now a thing, could you try it?

airdrop/airdrop-distributor/tests/functions/constructor.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        airdrop_constructor(
+            10,
+            &deploy_wallet.airdrop_distributor,
+            [1u8; 32],
+            asset.asset_id,
+        )
+        .await;
+
+        assert_eq!(
```
Braqzen, line 55: Assertions are unnecessary when a test is meant to revert. In the future we should be able to parse out the log which made it fail to prove that it failed where we expect it

Braqzen (changes requested):
1. Once [this doc](https://github.com/FuelLabs/sway-applications/blob/72c70cda427b8cc2e2226f5dd7cc19a373b39df6/APPLICATION_PROGRESS.md#decentralized-apps) goes in, could you add some content to the airdrop?
    - Looking at the `README` for this project it contains a "current state of app" section. This can be moved into this new doc once it gets some more eyes
    - The "releases" `TODO` in the distributor manifest should be mentioned as well
    - Any `TODO`s in the tests should be added here too
3. It looks like the build script, on line 7, is doing the same thing as on line 9 in the separate conditional. Is that correct? Should they be merged?
    - I'm also wondering if we should start joining projects i.e. only list "airdrop" in the matrix of the CI file and then the build script can do all the jumping that it needs. This groups the content into 1 section per project
4. In the project `README` the spec section at the bottom, at the end of the line, mentions a `fundraiser`. This bolognese should be updated to mention the correct application
5. The `SPEC` contains info to be presented to users, under the use cases, but I'm unable to find where this functionality exists. Does it exist?
    - It also contains requirements from the first iteration of the spec in some other project. I was planning on updating that at some point and removing it so that it's closer to the escrow and dao (since adding a spec where one does not exist takes priority over re-writing an existing one). The spec in this PR looks visually appealing but I'm unsure if it should be removed since doc parsing will eventually, hopefully, be a feature and thus we will be able to parse out the docs from the contracts automatically. The README also mentions a non-technical spec while this is technical imo. Perhaps we should have a separate doc for a technical spec? Thoughts?
    - Could you add a diagram to show the flow of info similar to the escrow and dao?
    - I think the UI placeholder at the bottom of the spec can be removed until a UI is adde
6. `distributor`
    - I wonder if the constructor needs the `require` for `claim != 0`. Height cannot be zero (genesis is block 1, correct?) and if they set the claim to be 0 then I'm unsure of what the side effect would be. Is this assertion needed or is it just semantic?
7. I think the `Signer` import for the tests is not required. Can you remove it where ever it does not error?
8. I think the test modules do not need to be marked as `pub`, can you check?
9. Can you add tests for `end_block` and `merkle_root`?

airdrop/SPECIFICATION.md
```diff
@@ -0,0 +1,233 @@
...
+2. A Merkle root constructed from a Merkle Tree where each leaf is the resulting `sha-256` hash of a `(user address, amount)` tuple
+3. A length of time in which a claim is valid
+
+### Claim Airdrop
+
+A user should be able to claim their airdrop
+
+1. Before the claim period is over
+2. If they has not yet claimed
+3. If the amount they are claiming is verifiable
```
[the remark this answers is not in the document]
bitzoic: I've added this suggestion however, excluded the "within". This would suggest the user can claim a smaller portion of tokens than allottedwhich is not true. They are only able to claim the exact amount of tokens they have been allotted.

bitzoic:
> 1. Once [this doc](https://github.com/FuelLabs/sway-applications/blob/72c70cda427b8cc2e2226f5dd7cc19a373b39df6/APPLICATION_PROGRESS.md#decentralized-apps) goes in, could you add some content to the airdrop?

I've included this in the README to point to this file. Of course, right now it points to something that doesn't exist. I'll also make an issue to update this file.

> 2. It looks like the build script, on line 7, is doing the same thing as on line 9 in the separate conditional. Is that correct? Should they be merged?

Yes this is correct, these are two different projects in CI and run tests in parallel. The airdrop-distributor is dependent on the simple-token binaries to run and therefore needs to be compiled alongside the airdrop-distributor. 

>    * I'm also wondering if we should start joining projects i.e. only list "airdrop" in the matrix of the CI file and then the build script can do all the jumping that it needs. This groups the content into 1 section per project

This would also be useful in the Sway-Libs repo. I believe right now it only acts as a single project due to all the subdirectories and causes CI to take 10 minutes to complete. 

> 4. The `SPEC` contains info to be presented to users, under the use cases, but I'm unable to find where this functionality exists. Does it exist?

Nope, this functionality does not yet exist. The fundraiser spec has this as well and others do not. I was unsure whether the others were incomplete or updated, they should all be the same format. Maybe we need a SPECIFICATION template? Don't know where in the repo this would go though...

>    * The spec in this PR looks visually appealing but I'm unsure if it should be removed since doc parsing will eventually, hopefully, be a feature and thus we will be able to parse out the docs from the contracts automatically. 

I don't believe so. I'm unfamiliar with the current state of doc parsing and believe it would be unwise to remove it with the anticipation of that feature. Any dev that checks out the repo right now should have some information available. I can have the spec conform to the dao and escrow spec however.

> The README also mentions a non-technical spec while this is technical imo. Perhaps we should have a separate doc for a technical spec? Thoughts?

It is definitely technical, perhaps I struggle to write less technical documentation. Is there really a need for a separation in technical and non-technical specification or should we just get rid of that statement all together? 

>    * I wonder if the constructor needs the `require` for `claim != 0`. Height cannot be zero (genesis is block 1, correct?) and if they set the claim to be 0 then I'm unsure of what the side effect would be. Is this assertion needed or is it just semantic?

If claim is set to zero, the airdrop is never active. In terms of ensuring that the constructor doesn't get called again, yes it does not need to be needed. Perhaps it could be removed to let the users shoot themselves in the foot.

> 7. I think the test modules do not need to be marked as `pub`, can you check?

Yes they need this.

matt-user (changes requested):

airdrop/airdrop-distributor/src/main.sw
```diff
@@ -0,0 +1,126 @@
...
+
+    #[storage(read, write)]
+    fn constructor(
+        claim_time: u64,
+        merkle_root: b256,
+        token_contract: ContractId,
+    ) {
+        // If `end_block` is set to a value other than 0, we know that the contructor has already
+        // been called.
+        require(storage.end_block == 0, InitError::AlreadyInitialized);
```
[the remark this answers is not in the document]
bitzoic: This was recently discussed with @Braqzen. It was determined that this is something that should be implemented in the front end and was removed to save gas.

airdrop/simple-token/src/interface.sw
```diff
@@ -0,0 +1,35 @@
...
+    /// 
+    /// * `minter` - The Address or Contract which will be permissioned to mint tokens.
+    /// * `token_supply` - The total number of tokens that may ever be minted.
+    /// 
+    /// # Reverts
+    /// 
+    /// * When the constructor has already been called.
+    /// * When the provided `token_supply` is zero.
+    #[storage(read, write)]
+    fn constructor(minter: Identity, token_supply: u64);
```
Braqzen, line 34:
```suggestion
        /// An example constructor which implements an airdrop distributor contract.
    /// 
    /// # Arguments
    /// 
    /// * `minter` - The Address or Contract which will be permissioned to mint tokens.
    /// * `token_supply` - The total number of tokens that may ever be minted.
    /// 
    /// # Reverts
    /// 
    /// * When the constructor has already been called.
    /// * When the provided `token_supply` is zero.
    #[storage(read, write)]
    fn constructor(minter: Identity, token_supply: u64);

    /// An example function that is to be called by the airdrop distributor contract.
    /// 
    /// The mint function is authorized to be called only by the airdrop contract.
    /// 
    /// # Arguments
    /// 
    /// * `amount` - The quantity of tokens that is to be minted.
    /// * `to` - The user which should recieve the minted tokens.
    /// 
    /// # Reverts
    /// 
    /// * When the sender is not the airdrop contract.
    /// * When the amount of tokens to be minted is greater than the total supply.
    #[storage(read, write)]
    fn mint_to(amount: u64, to: Identity);
```

Hopefully this pasted correctly

airdrop/airdrop-distributor/tests/functions/merkle_root.rs
```diff
@@ -0,0 +1,27 @@
...
+    use super::*;
+
+    #[tokio::test]
+    async fn returns_root() {
+        let (deploy_wallet, _, _, _, asset, claim_time) = setup().await;
+
+        airdrop_constructor(
+            claim_time,
+            &deploy_wallet.airdrop_distributor,
+            [2u8; 32],
```
Braqzen, line 17: I would consider factoring out this unnamed array into a variable (presumably "merkle root") so that a reader has an easier time understanding this magic structure. I would do this in all files. Having multiple roots to choose from would be fine too. If that is a valid idea then a struct containing them could be returned from the setup

airdrop/airdrop-distributor/src/main.sw
```diff
@@ -0,0 +1,100 @@
...
+use utils::mint_to;
+
+storage {
+    /// Stores the ClaimData struct of users that have interacted with the Airdrop Distrubutor contract.
+    /// Maps (user => claim)
+    claims: StorageMap<Identity, ClaimData> = StorageMap {},
+    /// The block at which the claiming period will end.
+    end_block: u64 = 0,
+    /// The computed merkle root which is to be verified against.
+    merkle_root: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000,
```
Braqzen, line 33: This should be an option in the future when the SDK supports it.
I thought I posted it somewhere else too but I cannot find it now... I think any default value should eventually be set to None from Option. This will prevent any potential weird behaviour if the value turns out to be valid in some condition

Braqzen (changes requested):
1. I was under the impression that a user can only claim once, the full amount, however the diagram has a looping line on the claim. Can you elaborate / remove that loop?
2. The table of contents has weird formatting for the functions. They are all on 1 line. Could you look into this?
3. I've made some suggestions in regards to using the term `asset` instead of `token` but I have not done this in all places so that it doesn't spam too much. I think it's better to avoid `tokens` and refer to `assets` so that people start to learn more about the UTXO model over the Ethereum account model. It encourages the readers to learn a little terminology
4. I would consider adding a little note in any test that is ignored so that a reader knows why the test would fail if it were to be included
5. Why is `AirdropIdentity` used rather than referring to the original `Identity`?
    - I would consider factoring this into a structure where you index since this seems to be repeated multiple times

airdrop/airdrop-distributor/tests/functions/claim.rs
```diff
@@ -235,17 +243,19 @@ mod revert {
...
-            (*wallet2.wallet.address().hash(), 2),
-            (*wallet3.wallet.address().hash(), 3),
-        ];
+        let identity_a = AirdropIdentity::Address(wallet1.wallet.address().into());
+        let identity_b = AirdropIdentity::Address(wallet2.wallet.address().into());
+        let identity_c = AirdropIdentity::Address(wallet3.wallet.address().into());
+        let minter = TokenIdentity::ContractId(deploy_wallet.contract_id);
         let key = 0;
         let num_leaves = 3;
+        let airdrop_leaves: [(AirdropIdentity, u64); 3] = [
```
matt-user, line 252: It might be useful to add a helper function to get the airdrop leaves and identities as this is repeated across tests.

matt-user (commented):
Minor suggestion.  Feel free to take it or leave it.

bitzoic:
> 1. I was under the impression that a user can only claim once, the full amount, however the diagram has a looping line on the claim. Can you elaborate / remove that loop?

The looping line signifies other users claiming their airdrop, but in terms of the user that creates the airdrop this is not a possible action and has been removed.

> 5. Why is `AirdropIdentity` used rather than referring to the original `Identity`?

There is a collision between the Airdrop-Distributor contract and the Simple-Asset contract's Identity. This has been moved into its own function in utils.

airdrop/simple-asset/src/main.sw
```diff
@@ -0,0 +1,50 @@
...
+    token::mint_to,
+};
+
+storage {
+    /// The Address or Contract that has permission to mint.
+    minter: Option<Identity> = Option::None(),
+    /// The maximum quantity of the asset ever to be minted.
+    asset_supply: u64 = 0,
+    /// The current quantity of the asset minted.
+    asset_minted: u64 = 0,
```
Braqzen, line 27:
```suggestion
    /// The current quantity of the asset minted.
    asset_minted: u64 = 0,
    /// The maximum quantity of the asset ever to be minted.
    asset_supply: u64 = 0,
    /// The Address or Contract that has permission to mint.
    minter: Option<Identity> = Option::None(),
```

airdrop/airdrop-distributor/tests/functions/end_block.rs
```diff
@@ -0,0 +1,31 @@
...
+            asset.asset_id,
+            claim_time,
+            &deploy_wallet.airdrop_distributor,
+            root,
+        )
+        .await;
+
+        assert_eq!(
+            end_block(&deploy_wallet.airdrop_distributor).await,
+            provider.latest_block_height().await.unwrap() + claim_time - 1
```
Braqzen, line 28: Is there a reason for the `-1`?

airdrop/airdrop-distributor/tests/functions/claim.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+mod revert {
+
+    use super::*;
+
+    #[tokio::test]
+    #[should_panic(expected = "Revert(42)")]
+    async fn panics_after_claim_period() {
```
Braqzen, line 141: Not a big deal but I would personally not prepend `panics_` so that the running tests say `revers::after_claim_period` as opposed to `reverts::panics_after_claim_period`

Braqzen (commented):
I'm almost ready to approve. Mostly some questions and a few suggestions. That table of contents is bothering me though. I use `Markdown all in one` in VsCode for automatic formatting.

airdrop/airdrop-distributor/src/data_structures.sw
```diff
@@ -0,0 +1,15 @@
...
+    claimed: bool,
+}
+impl ClaimData {
+    fn new(amount: u64, claimed: bool) -> Self {
+        Self {
+            amount,
+            claimed,
+        }
+    }
+}
```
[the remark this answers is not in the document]
bitzoic: I believe it did as I did not write it like this. Running the formatter again it looks like it doesn't modify anything if I add the lines back or leave it as is.

airdrop/airdrop-distributor/tests/functions/claim.rs
bitzoic: I agree this reads better. I believe there are a few applications that follow this format so this should be updated.

airdrop/airdrop-distributor/tests/functions/end_block.rs
bitzoic: I've enabled manual blocks [here](https://github.com/FuelLabs/sway-applications/blob/6cc956d75d19e57e23bbe719a1ffd328296f3840/airdrop/airdrop-distributor/tests/utils/mod.rs#L266). Every time a call is made the block height increases. So calling the `airdrop_constructor` would assign the block height at say 1. Then `end_block` would be 2. Getting the `latest_block_height()` would then need to be 1 less as we have moved forward 1 block in time from when we assigned the claim time in the constructor. 

I've gone ahead and flipped these so that we do not need to subtract by one.

airdrop/SPECIFICATION.md
```diff
@@ -0,0 +1,75 @@
+Table of Contents
+- [Overview](#overview)
+- [Use Cases](#use-cases)
+    - [Actions that users are able to perform](#actions-that-users-are-able-to-perform)
+        - [Airdrop Core Functionality](#airdrop-core-functionality)
+    - [`constructor()`](#constructor)
+        - [`claim()`](#claim)
```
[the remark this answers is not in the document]
bitzoic: I have 0 extensions in VsCode. I'll see what's going on here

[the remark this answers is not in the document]
bitzoic: Should be fine now

Braqzen (dismissed):
May want to wait for more eyes on this

airdrop/SPECIFICATION.md
```diff
@@ -62,13 +62,13 @@ This sub-section details what a user is able to do e.g. click a button and "x, y
...
 
-### Simple Token Core Functionality
+### Simple Asset Core Functionality
 
 #### `constructor()`
 
-1. Allows minting of the token to begin
-    1. Requires a token supply be given
-    2. Requires a user that will be authorized to mint
+1. Allows thw owner to begin minting the asset and requires
```
matt-user, line 69:
```suggestion
1. Allows the owner to begin minting the asset and requires

```

matt-user (changes requested):
I hate to be this guy.  But after this should be good.

airdrop/SPECIFICATION.md
bitzoic: Thanks for catching this

matt-user (approved):

airdrop/SPECIFICATION.md
Braqzen: I like "this guy".

Braqzen (approved):

PR #168, opened 2022-07-15
StakingRewards.sw

### LINKED ISSUE #161: Staking Smart Contract
## Summary

The contract should replicate the functionality from [Synthetixio](https://github.com/Synthetixio/synthetix/blob/be8ef365d206065f5afe8844866d969249a3b476/contracts/StakingRewards.sol)

simonr0204:
## Type of change

- New feature

## Changes

Implements a Sway port of Synthetix's [StakingRewards.sol](https://github.com/Synthetixio/synthetix/blob/be8ef365d206065f5afe8844866d969249a3b476/contracts/StakingRewards.sol)

Closes #161


Braqzen (changes requested):
1. Could you rename the files to follow the convention in [CONTRIBUTING.md](https://github.com/FuelLabs/sway-applications/blob/master/CONTRIBUTING.md#project-structure)?
2. `main.sw`
    - Could you merge all the imports into 1 import (context)?
    - Also the `*` imports should be changed to be explicit and alphabetical
    - There is a comment about precision and a constant that is commented out. For some reason there is a "ONE" constant too which is not particularly descriptive. Can this be cleaned up?
    - Variables in `storage` should be alphabetical and documented
    - All the functions in the contract should be alphabetical - this means that there is no need for extraneous comments such as "getters here"
    - The error message in each require should really be an Enum depiciting the specific error rather than some string
    - I don't understand what a "test" timestamp is and how that is semantically relevant. I think this should be renamed to timestamp
    - In the function `stake` I think line 150 should be moved directly into the require since it is being used once. If you want to keep the amount as a variable on line 147 then I suggest renaming line 153 to `user` and then replacing the logging of the struct to be `StakedEvent { amount, user }`. The `amount: amount` is not needed since it's named exactly the same and the compiler can infer that.
    - Events should have the postix `Event`, ref the example in the point above
    - All of the functions outside of the contract, i.e. at the bottom, should be moved into whatever library files they need to be in e.g. utils.sw etc.
        - There has been discussion a while back and it's been decided that functions should not be prefixed with an underscore (to denote privacy)
        - Functions that touch storage cannot be simply moved into libraries since storage cannot be accessed from libraries - at least currently in version 0.19.2 of forc. This means that those functions would need to be re-written or the code should be moved directly into the ABI calls
        - `fn recover_tokens()` has a param `token_address` of `ContractId`. That is semantically incorrect and may be confusing since we have an `Address` type. This should be renamed to something else e.g. `asset`
        - Logs should be moved out of helper functions and directly into the ABI calls since it makes it easier to follow the data that is being logged and provides stricter control of what is logged. This means that we do not spam multiple logs and instead we can construct 1 meta log per ABI call or so at the end of the function (or depending on the control flow in multiple places).
3. `interface.sw`
    - Join the imports into 1 rather than multiple lines and alphabetically order the functions.
    - The interface is what is likely to be used by a user therefore each function should be documented. There are extensive examples in every project in the master branch of the repo. Again, it's all in the contributing doc which should be reviewed before the next PR request.
4. Similar import and alphabetical comments in other files
5. Events should also be documented
6. The tests should be split up into functions, 1 module per abi function with relevant success and revert cases. This can be seen in other projects too
7. There's only a handful of tests. Every function should be tested even if it's just a getter. Those would be tested with the default value and then a value after it has been changed
8. The utils should be clearly separated into abi and helper functions
9. This should be added to the main repo README and it should have its own README and a basic SPEC

simonr0204:
`test_timestamp` is a stand-in for the actual timestamp (c.f. `block.timestamp` in Solidity) once implemented. In order to test the contract we need this dummy timestamp.  @SwayStar123 maybe just add comments explaining this? Or should we put this PR on hold until that feature is available ?

simonr0204:
Ok. Let's finish this review cycle, address remarks, and then set to Draft status with and tag with Awaiting Sway Features.

Braqzen:
> Ok. Let's finish this review cycle, address remarks, and then set to Draft status and tag with Awaiting Sway Features.

Note that there has already been some work to implement timestamps however the discussion has been to not use them for anything that is critical to the point of measuring a period of less than 30 minutes.
cc @adlerjohn

simonr0204:
It appears that since I created the PR I can't actually review it, even though i've reassigned it.

Apart from style/quality as identified by @Braqzen , biggest remark is it needs more tests. We need to test all the permissioned functions (passing and not passing, e.g. due to lack of permissions), check that after setting new rewards that this is reflected in accruals after the change, etc.

SwayStar123:
So the private functions here are doing alot of reads. For example the `_reward_per_token` function has 5 different storage reads. This would be quite messy to just replace with function arguments.

Similarly, `_withdraw` and `_get_reward` call the `_update_reward` function, which is entirely just storage writes.

I think these private functions should just stay in the main file, it would be quite ugly to try to place them in another file
@Braqzen

SwayStar123:
i think the CI is failing because the formatter is not updated? I updated my forc fmt to the latest and its not changing anything more

Braqzen (changes requested):
I recommend carefully re-reading my previous review since almost all of my comments are the same and yet my review has been "dismissed as stale".

1. Latest `fuels` is version `0.22`
2. What does it mean to have a value of a `b256` as a string for an `Identity` type in the toml?
    - An identity has either an Address or ContractId and they have a value but Identity does not have a value
    - Can these be documented somewhere if they are not already so that they aren't some magic numbers?
3. The very first comment in my previous review was to rename the files to follow the contributing doc. Please read that again.
    - Similarly, `token_address`
    - Again... imports
    - ... Alphabetically order things, I can see the events are not ordered. I assume other things are not as well
    - ... I've also mentioned documenting the events. This means the fields should be documented contextually
    - Test separation
        - This includes `mod.rs`
    - ... Hasn't been added to the repo README
        - Hasn't got its own README or SPEC
    - Not all events have not been cleaned up e.g. `reward: reward` should be just `reward`
    - Requires still use strings
    - Precision comment is still there
    - Events should be at the end of the ABI functions rather than hidden inside util functions
5. Any function that logs should return a [CallResponse](https://github.com/FuelLabs/sway-applications/blob/master/fundraiser/tests/utils/mod.rs#L39) so that a log can be verified once that feature is implemented in the SDK
    - If a function logs and returns a value then it should return that type wrapped in the CallResponse
6. Now that [APPLICATION_PROGRESS](https://github.com/FuelLabs/sway-applications/blob/master/APPLICATION_PROGRESS.md) exists this should be added in there too

simonr0204:
should be unblocked by https://github.com/FuelLabs/sway/pull/2752

SwayStar123:
the tests currently still assume im using the "test_timestamp" thing, Im not sure how to manipulate timestamps in the blocks tho, so not sure what I can do to ensure the tests run the same way every time

Braqzen:
> the tests currently still assume im using the "test_timestamp" thing, Im not sure how to manipulate timestamps in the blocks tho, so not sure what I can do to ensure the tests run the same way every time

I believe the SDK does not currently support that feature.

SwayStar123:
@simonr0204 I dont know what the precision stuff is, can you tell me what comments to add/explain it?

simonr0204:
> @simonr0204 I dont know what the precision stuff is, can you tell me what comments to add/explain it?

Since all numbers are integers, when performing arithmetic one must decide on the precision to use for the operands. In other words, if the smallest indivisible balance is `1`, then how much should that be scaled by to make "one" unit of the token? 

This is necessary to define when dealing with things like rates, since for example "10 tokens per day" requires a consistent definition of "one" token. This number should be reasonably large to allow for a reasonable precision in the calculations (think integer division) , but small enough to fit in a `u64`

For example, the ERC20 standard on Ethereum is `10**18`, and you'll see lots of scaling by the number everywhere in defining protocols. So far we've been thinking `10**9` is a good round number for applications on Fuel.

Braqzen:
> > * Events should be at the end of the ABI functions rather than hidden inside util functions
> 
> @Braqzen What about for `_get_reward` where the log is under an if statement?

IMO contracts that use many DRY functions are brittle and can probably be redesigned. Taking a look at that function it's only actually used in 2 places and it doesn't have many lines or any complex functionality. I would delete it and inline the code.

simonr0204:
@SwayStar123 Can this one be finished off now that you can read the timestamps from the transaction receipts? You can calculate the time elapsed from deposit to claim, and from there check the yield is as expected for that interval.

bitzoic:
Closing due to inactivity

PR #182, opened 2022-08-10
Break Escrow `harness.rs` into individual modules

### LINKED ISSUE #181: Break Escrow `harness.rs` into individual modules
### Motivation

The file is around 4,000 lines long and that's even after abstracting calls into `utils` to keep it DRY.

The aim is to try a new structure for testing namely to break each function into its own test file.

The structure would be

```bash
tests/
|
---- utils
|        |
|        ---- mod.rs
|
---- functions
|        |
|        ---- ABI func 1 (abc.rs)
|        ---- ABI func 2 (def.rs)
|
---- harness.rs
```

Braqzen:
## Type of change

## Changes

- Split `harness.rs` into modules where each module is a function from the ABI

## Related Issues

Closes #181


bitzoic (approved):
Makes things easier to find, limits unnecessary imports, and I don't have to endlessly scroll. LGTM!

PR #198, opened 2022-08-16
Mattauer/oracle sc

### LINKED ISSUE #164: Oracle Smart Contract
## Summary

The first iteration of the smart contract should be simple therefore it should have

- 1 setter function for its price feed (only the node can update)
- 1 getter function for contracts to retrieve the data

Any price feed should be fine as long as the data can be freely queried / via the use of an API key.
One suggestion is the BTC / USD price.

## Rough Contract Design

- Constructor to set the address that is the owner (node)
- 1 setter which can only be called by the owner to update the price and it emits an event for the latest change in data
- 1 getter for the price feed

## Future work

A future update would be to allow multiple oracles to aggregate data and submit votes with the majority being the chosen value

matt-user:
## Type of change

- New feature

## Changes

- Created a very basic oracle with a getter and a restricted setter for the price of an asset

## Related Issues

Closes #164


oracle/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "oracle"
+version = "0.1.0"
+authors = ["Matt Auer"]
```
bitzoic, line 4:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

oracle/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["Matt Auer"]
```
bitzoic, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

oracle/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+};
+
+use interface::Oracle;
+use data_structures::State;
+use errors::{AccessError, InitializationError};
+use events::PriceUpdateEvent;
+
+storage {
+    /// The Identity that can control the oracle (node)
+    owner: Identity = Identity::ContractId(BASE_ASSET_ID),
```
bitzoic, line 24: In order to avoid setting the owner to the `BASE_ASSET_ID`, perhaps set this as an `Option`. There is a section on [Contract Ownership](https://fuellabs.github.io/sway/v0.20.2/blockchain-development/access_control.html#contract-ownership) in the Sway Book.

oracle/tests/functions/constructor.rs
```diff
@@ -0,0 +1,29 @@
...
+    use super::*;
+
+    #[tokio::test]
+    #[should_panic]
+    async fn panics_when_reinitialized() {
+        let user = setup().await;
+        constructor(&user.oracle, Identity::Address(Address::from(user.wallet.address()))).await;
+        constructor(&user.oracle, Identity::Address(Address::from(user.wallet.address()))).await;
+    }
+}
```
bitzoic, line 29: It looks like your test files are missing new lines

oracle/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+            price: new_price
+        });
+    }
+
+    #[storage(read)] fn owner() -> Identity {
+        storage.owner
+    }
+
+    #[storage(read)] fn price() -> u64 {
+        storage.price
```
bitzoic, line 56: I'm not sure if this is done in Solidity or not, but should you be able to query the price if the oracle hasn't been initialized yet?

Braqzen (changes requested):
Could you look over the code for quality?

1. Cargo version should be 0 instead of 1
2. New lines at the end of each file
3. I don't know if it matters but the spec has a capitalized extension
4. The price in the spec is missing `()`
5. `set_price` should use a var `price` so that it can be inferred in the struct directly
6. Alphabetical ordering of everything - this includes var names in each test function
7. Since the sender is being used once in `set_price` I would not assign it to a var and instead use it directly in the `require`
8. Could you use the `owner()` in the `construct()` test as another layer of validation?
9. Could you also merge the two imports into 1 for `fuels` in `constructor.rs` similar to how the `owner.rs` does it?
10. Could you add 1 more test for the owner when the constructor is not called first?
11. Similarly a check for the default price being 0?
12. In the project structure of the readme can you add the spec under `README.md`?
13. The readme also points to the escrow logo which does not exist in this project nor should it. Could you amend this?
14. I would also add the `owner()` to the spec
15.

oracle/src/main.sw
matt-user: I was debating whether or not to require the oracle to be initialized.  I decided not to just because 1.) It's unlikely any consumer application will interact with the oracle before it is initialized.  2.) IMO it should be the responsibility of a consumer contract to ensure they are interacting with an initialized contract

matt-user: It may be hard to compare to solidity because solidity constructors are run during contract deployment and not afterwards

oracle/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+    /// The Identity that can control the oracle (node)
+    owner: Identity = Identity::ContractId(BASE_ASSET_ID),
+    /// Current price of tracked asset
+    price: u64 = 0,
+    /// The initialization state of the contract.
+    state: State = State::NotInitialized,
+}
+
+impl Oracle for Contract {
+    #[storage(read, write)] fn constructor(owner: Identity) {
```
simonr0204, line 32: Maybe instead of a constructor here, consider a configuration-time constant for the owner? This is a brand new feature and not documented yet but you can find some details in the PR :

https://github.com/FuelLabs/sway/pull/2549

oracle/src/main.sw
simonr0204: perhaps `price()` could return an `Option<u64>` ? I don't know if that's an unacceptable gas overhead for the edge case of non-initialized price. But maybe could also be useful if the node encounters a error- then you'd get a `None` instead of a stale price.

oracle/Cargo.toml
simonr0204, line 4:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

oracle/src/main.sw
matt-user: Options are not yet supported in the SDK so I think the best course of action is to just use a u64 and then change it to and Option in the future.

oracle/Forc.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "oracle"
+
+[constants]
+owner = { type = "b256", value = "0x09c0b2d1a486c439a87bcba6b46a7a1a23f3897cc83a94521a96da5c23bc58db" }
```
bitzoic, line 8: Where did this hardcoded value come from?

matt-user: It's the first wallet address generated deterministically by the SDK

matt-user: The deterministic wallets is a new feature not included in the latest version of fuels.  If you check out my Cargo.toml you'll see I'm getting my fuels from github.

bitzoic: Do these changes also include the ".DS_Store" binary file?

matt-user: Yeah it looks like it in files changed.  However, I don't see the file locally...  I could add it to the .gitignore but honestly I don't really know what it is

matt-user: Weird I see it with ls -a I guess I can just delete it?  I assume we don't want this file

bitzoic: Any file that does not have a name that starts with `.` for example `.gitignore` will not show using ls unless you give the all command. Are you using a Mac?

bitzoic: A quick google search is telling me it is a macOS file.

oracle/src/interface.sw
```diff
@@ -0,0 +1,27 @@
...
+
+use std::{identity::Identity, option::Option};
+
+abi Oracle {
+    /// Initialize the oracle with the owner (node)
+    ///
+    /// # Arguments
+    ///
+    /// - `owner` - Identity of node that controls the oracle
+    // #[storage(read, write)] fn constructor(owner: Identity);
```
bitzoic, line 11: I think this should be removed if you don't have the function implemented.

oracle/src/main.sw
```diff
@@ -0,0 +1,48 @@
...
+
+dep interface;
+dep data_structures;
+dep errors;
+dep events;
+
+use std::{
+    address::Address,
+    chain::auth::msg_sender,
+    constants::BASE_ASSET_ID,
```
bitzoic, line 11: I don't believe this import is used anywhere. Same goes for option

oracle/tests/functions/constructor.rs
```diff
@@ -0,0 +1,42 @@
...
+        assert!(false);
+        constructor(
+            &user.oracle,
+            Identity::Address(Address::from(user.wallet.address())),
+        )
+        .await;
+        let owner = owner(&user.oracle).await;
+        assert_eq!(
+            owner,
+            Option::Some(Identity::Address(Address::from(user.wallet.address())))
```
bitzoic, line 19: Small nitpick but using a variable to hold the `Identity` would make this a lot more readable and can be used on line 13 as well. Same applies for other files.

matt-user: Whoops that file shouldn't even be included

oracle/src/main.sw
```diff
@@ -0,0 +1,46 @@
+contract;
+
+dep interface;
+dep data_structures;
+dep errors;
+dep events;
```
Braqzen, line 6:
```suggestion
dep data_structures;
dep errors;
dep events;
dep interface;
```

oracle/src/main.sw
```diff
@@ -0,0 +1,46 @@
...
+    identity::Identity,
+    logging::log,
+    result::Result,
+    revert::require,
+};
+
+use interface::Oracle;
+use data_structures::State;
+use errors::AccessError;
+use events::PriceUpdateEvent;
```
Braqzen, line 20:
```suggestion
use data_structures::State;
use errors::AccessError;
use events::PriceUpdateEvent;
use interface::Oracle;
```

oracle/tests/utils/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub mod abi_calls {
+    use super::*;
+
+    pub async fn price(contract: &Oracle) -> u64 {
+        contract.price().call().await.unwrap().value
+    }
+
+    pub async fn owner(contract: &Oracle) -> Identity {
+        contract.owner().call().await.unwrap().value
+    }
```
Braqzen, line 19:
```suggestion
    pub async fn owner(contract: &Oracle) -> Identity {
        contract.owner().call().await.unwrap().value
    }
    
    pub async fn price(contract: &Oracle) -> u64 {
        contract.price().call().await.unwrap().value
    }
```

Braqzen (commented):
Some comments, after this I think it should be fine to go in.

1. As you've mentioned the Cargo dependency is currently pointing at git however that will eventually break things. Could you create an issue so that this is documented so that we can get around to this ASAP after a new version of the SDK is released?
2. Could you inform the reader in the spec under the owner section that the hardcoded owner constant is the first deterministically created address? Having a magic value that is unexplained is difficult to reason about especially given new features that are not released let alone documented in the relevant repo.
    - I think it might also be a good idea to add a single line comment in the `owner()` of the contract too because if I read the code then I have no idea where that variable is coming from. You can point the reader to the spec or copy and paste the same line here or write something else
    - Currently the owner is always going to return the address variant of the owner however we'd like to support contracts too. If it's possible to match on the owner variable then I think we should match and return either a contract id or the address variant of the identity. Could the owner const be changed to be an Address etc.? I saw there was a convo internally with some problem however I did not read it.
3. Could you do either of the following
    - Increase the size of the "oracle" text in both the logos by a couple font sizes or so. It looks a little small imo
    - Create an issue to document this and either assign it to me or yourself. Once this goes into master it can be updated later
4. In the readme the project structure has had the UI node changed to `node`. The oracle will eventually have a basic UI too. Could you add the UI back in so that a user knows that a UI will exist as well?
5. :) Could you change the extension (the `.MD` to `.md`) of the spec to be lowercase  and adjust any links that point to the spec?
6. `events.sw` imports an `Option` but that is unused. Could you delete that import?
7. I'm confused by the `set_price()` using a `read`. From what I can see it only writes but does not read. Has the compiler been updated to force a read whenever writing or was this always a thing and I've simply forgotten?
8. The `set_price` module has a test which panics and it uses a new syntax that I haven't seen. It's `contract._with_wallet`. It looks like you can init a contract from 1 wallet and then make calls on it from other wallets so now you do not have to keep track of other instances of contracts. 1 per wallet. Is that correct? Can that still be used or is this the new format and the previous format is no longer supported?

matt-user:
@Braqzen
I definitely agree that we should match on the owner const and return either a contract or address identity.  I'm not sure if it is possible to make the owner const an address or contract id (inside of the toml file).  I was unable to get it to work with anything other than b256.  @sezna implemented this feature so maybe he can shed some light on if it is possible.

Regarding 8.) Yes that is exactly what the code does (calls a contract initialized with one wallet from another separate wallet).  I see nothing in the docs that says we can no longer have multiple contract instances for different wallets.  The [docs](https://fuellabs.github.io/fuels-rs/v0.20.0/calling-contracts/calls-with-different-wallets.html) just indicate this is a shorthand way to do it.

simonr0204:
just noting that the most recent commit :

> fix: ci 

has CI failing 😅  .

`Variable "owner" does not exist in this scope.`   - that's a variable thats defined in `Forc.toml` - is something broken there ?

Braqzen:
> just noting that the most recent commit :
> 
> > fix: ci
> 
> has CI failing sweat_smile .
> 
> `Variable "owner" does not exist in this scope.` - that's a variable thats defined in `Forc.toml` - is something broken there ?

What I suspect is happening is that the locked version of forc (0.19.2) in CI does not have that feature and thus it fails.

The reason for a slightly lagging version is two fold, it takes a while to update projects to a new version so unless it's needed (i.e. there are breaking changes which make CI fail in the previously rolling release of the forc github action / other projects rely on newer features) then the version is not bumped and secondly there are a bunch of PRs / issues which currently rely on the locked version however they are blocked but the recent breaking changes in various places.

There's nothing inherently stopping them from receiving further _version_ updates however that does not resolve the current problem for them. It only unlocks new functionality. The only hassle there is that the issues would need to be updated to be broader to bump the version and the tests BUT if we bump a version for 1 project then they all need to be bumped so now all previous issues must be scrapped and 1 massive bumping issue must be created to do everything. That work won't be able to go in anyway since other repos have not caught up so there's a deadlock there.

Ultimately this means that those PRs must be ignored, all projects must be independently bumped while having outdated SDK versions and hope that an older core, SDK and a newer forc work. After this those other PRs would need to merge that work in.

Basically, it's spaghetti and I'm waiting for changes to be made so that I can bump the tests and then separately bump everything to be on the latest forc so that other PRs, like this one, can continue to go in without issues.

matt-user:
> just noting that the most recent commit :
> 
> > fix: ci
> 
> has CI failing 😅 .
> 
> `Variable "owner" does not exist in this scope.` - that's a variable thats defined in `Forc.toml` - is something broken there ?

Haha yeah I saw.  It works locally so it could be that the ci does not have the latest version of forc/sway that includes the config time constants.  I've been sick so I haven't looked into it yet.  I will investigate when I'm back in "office"

oracle/tests/functions/owner.rs
```diff
@@ -0,0 +1,16 @@
+use crate::utils::{abi_calls::owner, test_helpers::setup, Identity};
+use fuels::{prelude::*, tx::Address};
```
Braqzen, line 2: There is a warning that apparently this prelude is unused. Could you remove it to check?

oracle/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "oracle"
+version = "0.0.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels = { git = "https://github.com/FuelLabs/fuels-rs.git", features = ["fuel-core-lib"] }
```
Braqzen, line 9: This should be changed to some version rather than a repo

Braqzen (commented):
Could you also mention this in the [APP-SPEC](https://github.com/FuelLabs/sway-applications/blob/master/APPLICATION_PROGRESS.md)?

APPLICATION_PROGRESS.md
```diff
@@ -347,16 +347,34 @@ Will move from Apps repo to Libs repo soon
...
-<h3>User Interface</h3>
+- Needs a sway fix for using non-primitive types as configuration time constants to represent the owner variable appropriately.
+
+<h3>User Interface ❌</h3>
+
+- Not implemented
+
+<h3>Oracle Node ❌</h3>
+
+- Not implemented
```
Braqzen, line 360:
```suggestion
<h3>User Interface</h3>

<h3>Oracle Node</h3>
```

APPLICATION_PROGRESS.md
```diff
@@ -347,16 +347,34 @@ Will move from Apps repo to Libs repo soon
...
 
 <h3>Tests</h3>
 
-- <h3>Rust</h3>
-- <h3>Typescript</h3>
+- <h3>Rust ✅</h3>
+- <h3>Typescript ❌</h3>
+  
+  - Not implemented
```
Braqzen, line 367:
```suggestion
- <h3>Typescript</h3>
```

APPLICATION_PROGRESS.md
```diff
@@ -347,16 +347,34 @@ Will move from Apps repo to Libs repo soon
...
 
-<h3>Documentation</h3>
+<h3>Documentation ❌</h3>
+
+- Readme ❌
+  
+  - needs to include information about unimplemented features such as the User Interface and Oracle Node.
```
Braqzen, line 373:
```suggestion
- Readme ✅
```

This should go into the spec

APPLICATION_PROGRESS.md
```diff
@@ -347,16 +347,34 @@ Will move from Apps repo to Libs repo soon
...
+
+- Specification ❌
+  
+  - needs to include information about the Oracle Node
```
Braqzen, line 377:
```suggestion
- Specification ✅
```

Doesn't exist yet so it's fine

Braqzen (changes requested):

Braqzen (approved):

PR #217, opened 2022-09-10
fix: change order of parameters

### LINKED ISSUE #180: Change the order of the parameters in the `mint()` wrapper of the Escrow tests
The escrow has a wrapper function which mints an asset.
The order of the parameters is not alphabetically ordered as seen below

https://github.com/FuelLabs/sway-applications/blob/7b8bd69cca2091abc4194f230425a219a492cbbd/escrow/tests/utils/mod.rs#L204-L211

Line 204 should be `(address: &Bech32Address, amount: u64, contract: &MyAsset)`.
The tests in the `functions` directory should be updated to continue to pass.

Niravprajapati1:
## Type of change

## Changes

- Change the order of the parameters in the mint() wrapper of the Escrow tests

## Notes

- Note 1

## Related Issues

Closes #180

### DIFF
--- a/escrow/tests/functions/accept_arbiter.rs
+++ b/escrow/tests/functions/accept_arbiter.rs
@@ -129,9 +129,9 @@ mod revert {
         let asset = create_asset(defaults.asset_amount, defaults.asset_id).await;
 
         mint(
-            &defaults.asset,
             seller.wallet.address(),
             defaults.asset_amount * 2,
+            &defaults.asset,
         )
         .await;
         mint(
--- a/escrow/tests/functions/accept_arbiter.rs
+++ b/escrow/tests/functions/accept_arbiter.rs
@@ -19,9 +19,9 @@ mod success {
         let asset = create_asset(defaults.asset_amount, defaults.asset_id).await;
 
         mint(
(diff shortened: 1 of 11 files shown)


Braqzen (changes requested):
Thanks for contributing again but in order for this change to work there are some adjustments needed to be made in `escrow/tests/functions/`. This `mint()` function is used in various tests therefore if the function signature is changed then anywhere this call is made also needs to have its arguments re-ordered.

To find out where these changes need to be made you need to perform two steps

1. When inside `/escrow/` run `forc build --path tests/artifacts/asset` to build the test asset
2. Then run the tests using `forc test`

This should be done using `forc 0.24.1`. You can check your version using `forc --version` and you can use [these instructions](https://fuellabs.github.io/fuelup/master/concepts/toolchains.html) to set up a valid toolchain.

Another way is to check the CI below to see where there is an `X`. After clicking on `details` you may scroll down to where the error is and it will tell you what the problem is.

Niravprajapati1:
Thanks for help 🙂

Braqzen (approved):
Looks good to me!

PR #225, opened 2022-09-16
feat: allow execution of dao proposal

### LINKED ISSUE #105: Implement function execution for the DAO application
### Motivation

Currently we are passing a dummy function selector and function data to the dao voting contract.  We would like to make sure the low level assembly in the function actually works as intended.

matt-user:
## Type of change

## Changes

- The arbitrary dao execution code has been commented out
- In its place an abi and corresponding implementation have been added for execution of an abi known at CT

Closes #105


matt-user:
I'm going to close this as we should be able to execute an arbitrary call after [this](https://github.com/FuelLabs/sway/pull/3626) is merged.

PR #238, opened 2022-09-29
Name Registry contract

### LINKED ISSUE #137: Name Registry Smart Contract

SwayStar123:
## Type of change

- New feature

## Changes

Implements a very simple name registry which allows users to have a shorter name which can resolve to their Identity (Address/ContractId)

Closes #137


SwayStar123:
@Braqzen How should the pricing work in this?

Braqzen:
> @Braqzen How should the pricing work in this?

Probably a better question for @simonr0204 .

I would decide on some arbitrary value for an initial purchase and then if it expires then it costs the same amount. I'm unsure if you can transfer a name while you own it, you probably can, but I probably make that a new issue which would be an improvement unless it's so easy to implement that it might as well be a part of this issue.

SwayStar123:
> > @Braqzen How should the pricing work in this?
> 
> Probably a better question for @simonr0204 .
> 
> I would decide on some arbitrary value for an initial purchase and then if it expires then it costs the same amount. I'm unsure if you can transfer a name while you own it, you probably can, but I probably make that a new issue which would be an improvement unless it's so easy to implement that it might as well be a part of this issue.

I already have made a set_owner function, its pretty much the same as setting an address/identity

Braqzen (changes requested):
https://fuellabs.github.io/sway-applications/book/index.html

SwayStar123:
@Braqzen Sorry I think I was using the request review functionality wrong, bad communication on my part. I wanted to ask if the way im making this name-registry is fine.

Also looks like the CI is using an older version of forc so the build is failing. How do I update it?

Braqzen:
> @Braqzen Sorry I think I was using the request review functionality wrong, bad communication on my part. I wanted to ask if the way im making this name-registry is fine.
> 
> Also looks like the CI is using an older version of forc so the build is failing. How do I update it?

If you're referring to the logic then I'd need to properly review it but I'm unsure how much is completed which doesn't make it obvious about how I should approach the review.
Just by glancing it's not obvious why there is an identity vs an owner.
A specification is helpful to convey the flow of information and it helps to reveal potential flaws.

We do not bump to the latest release every time one is made because that's too much of a time sink. At the bottom of the repo readme there is a note stating the versions used.
If there is a critical feature that this requires that has been implemented between `0.24.4` and `0.24.5` then I can bump to the latest otherwise use the current release set for the repo.

SwayStar123:
@Braqzen any idea why the airdrop-distributor is suddenly not building? I didnt touch it

Braqzen:
> @Braqzen any idea why the airdrop-distributor is suddenly not building? I didnt touch it

@bitzoic I don't know how this has trickled in since the version in the distributor is locked and a new release has not been made but it looks like your recent change to the merkle lib has broken code here.

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+dep interface;
+
+use data_structures::Record;
+use errors::Errors;
+use events::{IdentityChanged, NameRegistered, OwnerChanged, RegistrationExtended};
+use interface::NameRegistry;
+use std::{
+    block::timestamp,
+    chain::auth::msg_sender,
+    constants::{
```
bitzoic, line 15: Since there is only one field, this can be collapsed without the braces

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+            },
+            None => {
+                log(Errors::NameNotRegistered);
+                revert(0)
+            }
+        }
+    }
+
+    #[storage(read, write)]
+    fn extend(name: str[8], duration: u64) {
```
bitzoic, line 53: Seems like anyone can extend the duration. Do we want this to be just the owner?

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+
+        log(RegistrationExtended {
+            duration,
+            name,
+            new_expiry: new_record.expiry,
+        });
+    }
+
+    #[storage(read)]
+    fn identity(name: str[8]) -> Identity {
```
bitzoic, line 75: I think this should check if the identity is expired and revert if it is. Otherwise the registration expires but still resolves to the same `Identity` even after it's expired, essentially enabling a lifelong registration.

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+            },
+            None => {
+                log(Errors::NameNotRegistered);
+                revert(0)
+            }
+        }
+    }
+
+    #[storage(read)]
+    fn owner(name: str[8]) -> Identity {
```
bitzoic, line 88: Same comment in regards to checking for expiry as with `identity()`.

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    #[storage(read, write)]
+    fn register(name: str[8], duration: u64) {
+        if storage.names.get(name).is_some() {
+            let record = storage.names.get(name).unwrap();
+            require(timestamp() > record.expiry, Errors::NameNotExpired);
+        }
+
+        require(duration / 100 * PRICE_PER_HUNDRED <= msg_amount(), Errors::InsufficientPayment);
```
bitzoic, line 107: Can you add parenthesis here so we know in what order the multiplication and division should happen?

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+        require(msg_asset_id() == BASE_ASSET_ID, Errors::WrongAssetSent);
+
+        let record = Record {
+            expiry: timestamp() + duration,
+            identity: msg_sender().unwrap(),
```
bitzoic, line 112: Just a thought but do you want to include the ability to resolve to a different `identity` than the one who is registering it? This would be passed into the function itself in that case and the front end would handle it.

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+        };
+
+        storage.names.insert(name, Option::Some(record));
+
+        log(NameRegistered {
+            expiry: record.expiry,
+            name,
+            owner: record.owner,
+        });
+    }
```
bitzoic, line 123:
```suggestion
    }
    
```

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+    #[storage(read, write)]
+    fn set_identity(name: str[8], identity: Identity) {
```
bitzoic, line 125: This should also check if the name is expired. I don't think users should be allowed to change their resolved `Identity` if the registration has expired.

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+
+        log(IdentityChanged {
+            name,
+            new_identity: new_record.identity,
+            old_identity: old_record.identity,
+        });
+    }
+
+    #[storage(read, write)]
+    fn set_owner(name: str[8], new_owner: Identity) {
```
bitzoic, line 146: Same comment about checking for expired

name-registry/src/errors.sw
```diff
@@ -0,0 +1,9 @@
+library errors;
+
+pub enum Errors {
```
bitzoic, line 3: Can you split this up into different categories?

name-registry/tests/utils/mod.rs
```diff
@@ -0,0 +1,114 @@
+use fuels::{prelude::*, tx::ContractId};
+
+// Load abi from json
+abigen!(MyContract, "out/debug/name-registry-abi.json");
+
+pub async fn get_contract_instance() -> (MyContract, ContractId, WalletUnlocked) {
```
bitzoic, line 6: Are you able to split this into a test utils and an abi calls section?

name-registry/tests/functions/expiry.rs
```diff
@@ -0,0 +1,34 @@
...
+
+        assert_eq!(old_expiry + 5000, new_expiry);
+    }
+}
+
+mod failing {
+    use crate::utils::*;
+
+    #[tokio::test]
+    #[should_panic]
```
bitzoic, line 26:
```suggestion
    #[should_panic(expected = "Revert(42)")]
```

Could you add this to all tests to make sure they are panicking in the way you intend? We've caught some bugs by retroactively adding this before.

bitzoic (changes requested):
Could you update the status of the project in the [APPLICATION_PROGRESS.md](https://github.com/FuelLabs/sway-applications/blob/master/APPLICATION_PROGRESS.md) file?

Very cool app fyi

name-registry/src/main.sw
SwayStar123: I dont see why, a person may want to use a different address to pay the fees, and even if a complete stranger pays the fees it does not harm the owner in any way

name-registry/src/main.sw
SwayStar123: Well my thought process was that the expiry is really just a way for someone else to grab the name if they want it, if its expired and yet no one wants it then let the old owner keep it. But I can just change it to not have it be like that

name-registry/README.md
```diff
@@ -0,0 +1,48 @@
+<p align="center">
+    <picture>
+        <source media="(prefers-color-scheme: dark)" srcset=".docs/fundraiser-logo-dark-theme.png">
+        <img alt="SwayApps Fundraiser Logo" width="400px" src=".docs/fundraiser-logo-light-theme.png">
+    </picture>
+</p>
+
+## Overview
+
+This is a simple name registry app built in Sway. Anyone can register a name for the price of 1 gwei per 100 seconds. The name will resolve to any Address/Identity that the owner sets.
```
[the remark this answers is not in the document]
SwayStar123: Fuel uses a precision of 9. I asked john and he said to use gwei

name-registry/src/main.sw
```diff
@@ -0,0 +1,165 @@
...
+            identity: msg_sender().unwrap(),
+            owner: msg_sender().unwrap(),
+        };
+
+        storage.names.insert(name, Option::Some(record));
+
+        log(NameRegistered {
+            expiry: record.expiry,
+            name,
+            owner: record.owner,
```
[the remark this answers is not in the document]
SwayStar123: Currently the owner is set as the identity when a new name is registered so it would always result in a duplicate identity for both owner and identity, so i omitted it

name-registry/src/main.sw
SwayStar123: Can you elaborate? I dont get what you meant

bitzoic: Your function definition would be `fn register(duration: u64, resolver: Identity, name: str[8]);` and you set `resolver` to the `identity` field of your `Record` struct. This would eliminate the user's need to make an additional call to set which `Identity` the registry should resolve to and save them some gas. I've done this personally in the past with ENS where I purchased an ENS from one wallet and set it to point to another

name-registry/tests/functions/expiry.rs
SwayStar123: Thank you, that found some flaws in my tests

name-registry/README.md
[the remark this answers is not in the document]
Braqzen: I think we should avoid price comparisons and generally equating one chain to another. That is for some high level overview of the general protocol rather than building 1-to-1 comparisons. I suggest removing any reference to any concept of Ethereum, such as the price here, and coming up with your own abstraction e.g. create some asset and decide how much of that asset is required. The base asset could be such an asset.

name-registry/src/interface.sw
```diff
@@ -0,0 +1,93 @@
+library interface;
+
+use std::identity::Identity;
+
+abi NameRegistry {
```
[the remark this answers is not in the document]
Braqzen: The reason that this link is misformatted is because there is a space after the `[`.

name-registry/src/main.sw
SwayStar123: The user would never have to make 2 transactions, the two abi calls would be made in the same script. But yeah sure I can change it to allow you to set the identity in the same call

name-registry/src/errors.sw
SwayStar123: What categories do you suggest?

name-registry/src/main.sw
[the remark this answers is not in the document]
SwayStar123: only twice, i dont think its necessary

name-registry/src/data_structures.sw
```diff
@@ -0,0 +1,12 @@
+library data_structures;
+
+use std::identity::Identity;
+
+pub struct Record {
+    /// The timestamp at which the name expires, and someone else can re-register the same name
+    expiry: u64,
+    /// The identity to which the name resolves to
+    identity: Identity,
```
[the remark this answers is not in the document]
SwayStar123: record.identity sounds better than record.resolvee to me

name-registry/tests/utils/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+    )
+    .await
+    .unwrap();
+
+    let instance = MyContractBuilder::new(id.to_string(), wallet.clone()).build();
+
+    (instance, id.into(), wallet, wallet2)
+}
+
+pub mod abi;
```
Braqzen, line 36: Move mod to the top.

Is there a reason for having a separate `abi` module rather than including it here? It's a matter of preference but I'm curious since the other apps follow the other pattern.

name-registry/tests/utils/abi.rs
```diff
@@ -0,0 +1,90 @@
+use crate::utils::*;
+
+pub async fn extend(instance: &MyContract, name: &String, duration: u64) {
```
Braqzen, line 3: These tests should be returning a callresponse so that logs can be checked. I think that is now testable in the latest SDK which the repo has been bumped to

name-registry/README.md
```diff
@@ -0,0 +1,49 @@
...
+        <source media="(prefers-color-scheme: dark)" srcset=".docs/fundraiser-logo-dark-theme.png">
+        <img alt="SwayApps Fundraiser Logo" width="400px" src=".docs/fundraiser-logo-light-theme.png">
+    </picture>
+</p>
+
+## Overview
+
+This is a simple name registry app built in Sway. Anyone can register a name for the price of 1 unit of the base asset per 100 seconds. The name will resolve to any Address/Identity that the owner sets.
+
+In this case the price paid is in the native asset on the Fuel network. More information can be found in the [specification](./SPECIFICATION.md).
```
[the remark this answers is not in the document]
simonr0204: Suggest:

> In this implementation the price is paid in the base asset.

name-registry/SPECIFICATION.md
```diff
@@ -0,0 +1,66 @@
...
+1. If the payment is sufficient for the duration
+2. If the payment is in the correct asset
+
+### `set_identity()`
+
+The owner of the name should be able to change the identity to which the name resolves
+
+### `set_owner()`
+
+The owner of the name should be able to change the owner of the name
```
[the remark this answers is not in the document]
SwayStar123: I used the fundraiser as a template, the specification in the fundraiser uses "should" everywhere so I thought thats what we were going for?

name-registry/tests/utils/mod.rs
SwayStar123: Someone requested it as a change

Braqzen: Correct, they did. That's because you've had all of them free-floating in the same file rather than like what the other apps do which is to have a mod inside the file for abi calls rather than move everything out of the file into a new module.

name-registry/src/interface.sw
```diff
@@ -0,0 +1,91 @@
...
+    /// Reverts if the name is not registered, if the payment is not sufficient for the duration, or if the wrong asset is sent
+    #[storage(read, write)]
+    fn extend(name: str[8], duration: u64);
+
+    /// Registers the name and assigns the sender as the owner and identity to resolve to
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name to register
+    /// * `duration` - The duration to register for
```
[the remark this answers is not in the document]
SwayStar123: This is the only function I can see which did not document all the parameters, which other functions are you referring to?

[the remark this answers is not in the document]
Braqzen: Oh, my bad. I must have scrolled or been distracted as I was writing this. I must have seen the args for this function and the parameters and thought that there were more.

name-registry/src/events.sw
```diff
@@ -0,0 +1,43 @@
+library events;
+
+/// The event for when a name has its resolving identity changed
+pub struct IdentityChanged {
```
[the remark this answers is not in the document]
SwayStar123: I dont know what other name to use, "resolvee" sounds too weird and "identity to which name resolves to" is too long. Open to suggestions but I think with the given documentation its fine

name-registry/src/main.sw
```diff
@@ -0,0 +1,154 @@
...
+    names: StorageMap<str[8], Option<Record>> = StorageMap {},
+}
+
+// TODO: Change the static 8 length str with a dynamic string when possible
+impl NameRegistry for Contract {
+    #[storage(read, write)]
+    fn extend(name: str[8], duration: u64) {
+        require(storage.names.get(name).is_some(), ValidityErrors::NameNotRegistered);
+        require((duration / 100) * price_per_hundred <= msg_amount(), AssetErrors::InsufficientPayment);
+        require(msg_asset_id() == BASE_ASSET_ID, AssetErrors::IncorrectAssetSent);
```
Braqzen, line 38: Ah, I see. I suggest taking the base asset and putting it inside the config as a config time constant to allow a user to configure their registry without having to change the code.

In the future, this could be expanded to use a registry of available assets and the user gets to select which asset they want to pay in and as long as the asset is specified in that registry and the calculation for the price is met then they can use any asset they like from the curated list.

Braqzen: I would also switch the order of the price and base asset checks for semantic reasons.

Do you go to a store and first count the amount and then check that the correct currency is used or do you first notice the currency and then tally up?

Just to drive the point home, you wouldn't count 100 carrots and then at the end say "sorry, these are carrots", you'd automatically look at me like I'm crazy and reject the carrots without counting them.

name-registry/src/main.sw
```diff
@@ -0,0 +1,154 @@
...
+
+        let previous_record = storage.names.get(name).unwrap();
```
Braqzen, line 40: I think you should be able to use struct destructuring like `let Record { expiry, identity, owner } = storage...` and then reference the variables directly below without having to do `previous_record.`

name-registry/src/main.sw
Braqzen: I'm wondering if we should restrict this to only allow the owner to extend the name. I'm indifferent either way but I'd like to hear your thoughts.

name-registry/src/main.sw
```diff
@@ -0,0 +1,154 @@
...
+
+        log(IdentityChangedEvent {
+            name,
+            new_identity: new_record.identity,
+            previous_identity: previous_record.identity,
+        });
+    }
+
+    #[storage(read, write)]
+    fn set_owner(name: str[8], new_owner: Identity) {
```
Braqzen, line 110:
```suggestion
    fn set_owner(name: str[8], owner: Identity) {
```

name-registry/src/main.sw
```diff
@@ -0,0 +1,154 @@
...
+        require(storage.names.get(name).is_some(), ValidityErrors::NameNotRegistered);
+        let previous_record = storage.names.get(name).unwrap();
+        require(timestamp() < previous_record.expiry, ValidityErrors::NameExpired);
+        require(previous_record.owner == msg_sender().unwrap(), AuthorisationErrors::SenderNotOwner);
+
+        let new_record = Record {
+            expiry: previous_record.expiry,
+            identity: previous_record.identity,
+            owner: new_owner,
```
Braqzen, line 119:
```suggestion
            owner
```

name-registry/src/main.sw
```diff
@@ -0,0 +1,154 @@
...
+            new_owner: new_record.owner,
+            previous_owner: previous_record.owner,
+        });
+    }
+
+    #[storage(read)]
+    fn expiry(name: str[8]) -> u64 {
+        require(storage.names.get(name).is_some(), ValidityErrors::NameNotRegistered);
+        let record = storage.names.get(name).unwrap();
+        require(timestamp() < record.expiry, ValidityErrors::NameExpired);
```
Braqzen, line 135: I think the getter should not be reverting here when a record has expired. I think it should return 0. I think it should continue to revert in the two getters below though.

Alternatively, the 3 getters can return an option which is None if the record does not exist / is expired otherwise it returns whatever it needs to. I think that's probably a better approach for modularity / integration with contracts. This will allow them to handle logic rather than crash.
This also aligns with your spec which does not mention any revert if the name has expired.

name-registry/SPECIFICATION.md
```diff
@@ -0,0 +1,66 @@
...
+1. If the payment is sufficient for the duration
+2. If the payment is in the correct asset
+
+### `set_identity()`
+
+The owner of the name can change the identity to which the name resolves
+
+### `set_owner()`
+
+The owner of the name can change the owner of the name
```
Braqzen, line 44:
```suggestion
Allows the owner to transfer ownership of the registered name
```

name-registry/SPECIFICATION.md
```diff
@@ -0,0 +1,66 @@
...
+### `extend()`
+
+Any user can extend the registration duration of a given name
+
+1. If the payment is sufficient for the duration
+2. If the payment is in the correct asset
+
+### `set_identity()`
+
+The owner of the name can change the identity to which the name resolves
```
Braqzen, line 40:
```suggestion
Allows the owner to change the resolving identity
```

Unsure about this suggestions but I think it's better than the current one

name-registry/SPECIFICATION.md
```diff
@@ -0,0 +1,66 @@
...
+
+A user can register a name for themselves
+
+1. If the name is not already registered
+2. If the payment is sufficient for the duration
+3. If the payment is in the correct asset
+
+### `extend()`
+
+Any user can extend the registration duration of a given name
```
Braqzen, line 33: Should mention that the name must be registered because otherwise it will revert

name-registry/SPECIFICATION.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Actions that users are able to perform
+
+This sub-section details what a user is able to do e.g. click a button and "x, y, z" happens.
+
+### `register()`
+
+A user can register a name for themselves
+
+1. If the name is not already registered
```
Braqzen, line 27: This is slightly misleading. A name can be "registered" but expired. Does that mean that it is still registered? I would reword this

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name to extend the duration of
+    /// * `duration` - The duration to extend by
+    ///
+    /// # Reverts
+    ///
+    /// * If the name is not registered
+    /// * If the payment is not sufficient for the duration
```
Braqzen, line 14:
```suggestion
    /// * If the payment is insufficient to cover the cost for the duration
```

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    /// * If the wrong asset is sent
```
Braqzen, line 15:
```suggestion
    /// * If the incorrect asset is sent
```

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    #[storage(read, write)]
+    fn extend(name: str[8], duration: u64);
+
+    /// Registers an entry into the registry for the given name, assigns it to the given owner, and resolves the name to the given identity when queried
```
Braqzen, line 19: This is too verbose for a summary. It should be a single overview stating that a name is registered and then after an empty new line, on the 3rd line, it can contain additional information about the resolution etc.

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    /// # Arguments
+    ///
+    /// * `name` - The name to register
+    /// * `duration` - The duration to register for
+    /// * `owner` - The owner of the name, which will be able to control the ownership and the resolving identity of the name
+    /// * `identity` - The identity to which the name would resolve to when queried
+    ///
+    /// # Reverts
+    ///
+    /// * If the name is registered or if it has not expired
```
Braqzen, line 30: Similar comment regarding the meaning of "registered"

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    /// * If the payment is not sufficient for the duration
```
Braqzen, line 31:
```suggestion
    /// * If the payment is insufficient to cover the cost for the duration
```

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    /// * If the wrong asset is sent
```
Braqzen, line 32:
```suggestion
    /// * If the incorrect asset is sent
```

name-registry/src/interface.sw
```diff
@@ -0,0 +1,99 @@
...
+    /// Sets the identity to which the name will resolve to
+    ///
+    /// # Arguments
+    ///
+    /// * `name` - The name to set the identity for
+    /// * `identity` - The identity which the name will resolve to
+    ///
+    /// # Reverts
+    ///
+    /// * If the name is registered
```
Braqzen, line 45: I don't think that this revert statement is correct, including the functions below.

name-registry/tests/functions/expiry.rs
```diff
@@ -0,0 +1,41 @@
+mod success {
+    use crate::utils::{
```
Braqzen, line 2: If the same imports are being used then I would factor them out to be on the top level and then perhaps do specific imports in each module

name-registry/tests/functions/extend.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+mod revert {
+    use crate::utils::{
+        abi::{extend, register},
+        get_contract_instance,
+    };
+    use fuels::prelude::*;
+    #[tokio::test]
```
Braqzen, line 46:
```suggestion
    use fuels::prelude::*;
    
    #[tokio::test]
```

name-registry/tests/functions/expiry.rs
```diff
@@ -0,0 +1,41 @@
...
+    use fuels::prelude::*;
+
+    #[tokio::test]
+    async fn can_get_expiry() {
+        let (instance, _id, wallet, _wallet2) = get_contract_instance().await;
+
+        let wallet_identity = Identity::Address(Address::from(wallet.address()));
+        let name = String::from("SwaySway");
+
+        register(&instance, &name, 5000, &wallet_identity, &wallet_identity).await;
```
Braqzen, line 15: The 5000 seems to be used multiple times. I would consider moving this into a `setup` and referencing that variable.

name-registry/tests/functions/identity.rs
```diff
@@ -0,0 +1,46 @@
+mod success {
+    use crate::utils::{
+        abi::{identity, register, set_identity},
+        get_contract_instance,
+    };
+    use fuels::prelude::*;
+
+    #[tokio::test]
+    async fn can_get_identity() {
```
Braqzen, line 9: I would move as much of the setup to the top of the function so that I don't have to look for wherever any variable might be defined inside the middle of a test.

name-registry/tests/functions/identity.rs
```diff
@@ -0,0 +1,46 @@
...
+    use crate::utils::{abi::identity, get_contract_instance};
+
+    #[tokio::test]
+    #[should_panic(expected = "Revert(42)")]
+    async fn cant_get_identity() {
+        let (instance, _id, _wallet, _wallet2) = get_contract_instance().await;
+
+        let name = String::from("SwaySway");
+
+        let _expiry = identity(&instance, &name).await;
```
Braqzen, line 44: Instead of assigning to a variable I would `.unwrap()` and let it crash that way

name-registry/tests/functions/register.rs
```diff
@@ -0,0 +1,65 @@
+mod success {
+    use crate::utils::{
+        abi::register, get_contract_instance, string_to_ascii, NameRegisteredEvent,
+    };
+    use fuels::prelude::*;
+    #[tokio::test]
```
Braqzen, line 6:
```suggestion
    use fuels::prelude::*;
    
    #[tokio::test]
```

name-registry/tests/functions/register.rs
```diff
@@ -0,0 +1,65 @@
...
+                identity: wallet_identity
+            }]
+        )
+    }
+}
+
+mod revert {
+    use crate::utils::{abi::register, get_contract_instance};
+    use fuels::prelude::*;
+    #[tokio::test]
```
Braqzen, line 34:
```suggestion
    use fuels::prelude::*;
    
    #[tokio::test]
```

name-registry/tests/utils/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use fuels::{prelude::*, tx::ContractId};
+
+// Load abi from json
+abigen!(NameRegistry, "out/debug/name-registry-abi.json");
+
+pub async fn get_contract_instance() -> (NameRegistry, ContractId, WalletUnlocked, WalletUnlocked) {
```
Braqzen, line 8:
```suggestion
pub async fn setup() -> (NameRegistry, ContractId, WalletUnlocked, WalletUnlocked) {
```

This is doing a lot more than just returning a contract instance and I suggest you add more to this, as per comment(s) above

name-registry/tests/utils/abi.rs
```diff
@@ -0,0 +1,123 @@
+use core::fmt::Debug;
+use fuels::{
+    client::types::TransactionStatus,
+    contract::contract::{CallResponse, ContractCallHandler},
+};
+
+use crate::utils::*;
```
Braqzen, line 7: This only have 1 function inside... this is just lazy now. Import explicitly.

Braqzen (changes requested):
2 more comments

1. Can you add this to the main repo readme and the application progress docs?
2. This should not be changing the CI for the entire repo. Bumping the repo should be done in a separate PR. Given that everything seems to be passing here without any changes to the other apps I won't request a reversion of this however for future reference either request for the repo to be bumped and have whatever is broken fixed or do it yourself in a separate PR

.github/workflows/ci.yml
```diff
@@ -89,7 +89,8 @@ jobs:
       - name: Run Forc tests
         run: |
           cd ${{ matrix.project }}
-          forc test
+          forc build
+          cargo test
```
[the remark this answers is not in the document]
SwayStar123: forc test is deprecated in forc 0.26

name-registry/src/main.sw
SwayStar123: People may want to use different addresses to extend their names, so I would prefer not having the owner extend it as a requirement

name-registry/src/main.sw
SwayStar123: I dont think the SDK supports Options so testing would be impossible if I returned an Option

name-registry/tests/functions/expiry.rs
SwayStar123: Using super just adds more lines for no reason (If you import at the top level the mods dont get access to the import, so you have to use super to import them again, making the top level import kinda pointless)

name-registry/src/interface.sw
SwayStar123: I dont know what you are referring to

Braqzen (dismissed):
Oh, right. So this is doing more than just bumping the version. This should not be changing the other contracts. That should be done in a separate PR. All the other apps need to be reverted and those changes should be made in a new PR which bumps the entire repo.

Once these changes have been reverted then re-request a review - after the repo has been bumped.

SwayStar123:
> Oh, right. So this is doing more than just bumping the version. This should not be changing the other contracts. That should be done in a separate PR. All the other apps need to be reverted and those changes should be made in a new PR which bumps the entire repo.
> 
> Once these changes have been reverted then re-request a review - after the repo has been bumped.

Bumping to 0.26 doesnt need any changes to other repos so ill just stick to that

name-registry/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Tests
+
+In order to run the tests make sure that you are in the root of this project i.e. `/path/to/name-registry/<you are here>`
+
+Run the tests
+   ```bash
+   forc build
+   cargo test
+   ```
```
Braqzen, line 49: Once this PR goes in can you raise 1 issue which says that this change needs to be made for each specification? It should be labelled as a `Good First Issue`.

Braqzen (commented):
1. Needs to be added to the root readme so that people know what this is, moreover the version specified at the bottom of the readme needs to be updated to reflect this as well
2. Hm, I recall a PR for the application progress so I guess that can be omitted here
3. I recall there being issues with config time constants. What is the status of that issue? I don't want to see a PR go in that will sometimes fail CI and sometimes pass
4. In `main.sw` there are some `match` expressions used. They return results which are wrapped inside `{}` and thus they take up a lot of room. Can you remove the `{}` so that you have each arm take up a single line instead of 3?

I think I'm close to approving.

SwayStar123:
> 3. I recall there being issues with config time constants. What is the status of that issue? I don't want to see a PR go in that will sometimes fail CI and sometimes pass
The CI will always pass, the problem was with the test rust files not properly saving and recognizing the change in config time constants or something like that

name-registry/tests/functions/expiry.rs
```diff
@@ -0,0 +1,47 @@
...
+        abi::{expiry, extend, register},
+        setup, EXTEND_DURATION, REGISTER_DURATION,
+    };
+    use fuels::prelude::*;
+
+    #[tokio::test]
+    async fn can_get_expiry() {
+        let (instance, _id, wallet, _wallet2) = setup().await;
+        let wallet_identity = Identity::Address(Address::from(wallet.address()));
+        let name = String::from("SwaySway");
```
Braqzen, line 12: These seem to be getting reused here and there. Perhaps a default struct should be returned so that you can set it up once it the `setup()` and then use the values as you see fit. If you need an exceptional value then that can be defined in the test itself

name-registry/tests/functions/identity.rs
```diff
@@ -0,0 +1,48 @@
...
+        assert_eq!(new_identity.0.value.unwrap(), wallet_identity2);
+    }
+}
+
+mod revert {
+    use crate::utils::{abi::identity, setup};
+
+    #[tokio::test]
+    #[should_panic(expected = "`Result::unwrap()` on an `Err` value")]
+    async fn cant_get_identity() {
```
Braqzen, line 41: I think we can be a bit more specific here i.e. `_when_not_registered` or something like that. It makes it clearer as to why we cannot perform this operation. A few tests use this

name-registry/tests/functions/register.rs
```diff
@@ -0,0 +1,82 @@
...
+            REGISTER_DURATION,
+            &wallet_identity,
+            &wallet_identity,
+        )
+        .await;
+    }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn cant_register_infinite() {
```
Braqzen, line 68: infinite or `maximum value / duration` or something? infinite is incorrect since it's technically the max limit, right? it also doesn't convey much info

name-registry/tests/functions/expiry.rs
```diff
@@ -0,0 +1,47 @@
...
+            new_expiry_response.0.value.unwrap()
+        );
+    }
+}
+
+mod revert {
+    use crate::utils::{abi::expiry, setup};
+
+    #[tokio::test]
+    #[should_panic(expected = "`Result::unwrap()` on an `Err` value")]
```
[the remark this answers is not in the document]
SwayStar123: What suggestions?

Those expected strings caught a few bugs in the testing, although the bugs have already been caught now so they can be removed, they should still be kept to prove the correct revert is occurring. (We do care about these specific errors, if the test is panicing due to some other error that was not meant to occur then that means theres a bug that the testing is falsely accepting as expected)

.github/workflows/ci.yml
```diff
@@ -137,4 +137,5 @@ jobs:
 
       - name: Build Sway
         run: forc build --path ${{ matrix.project }}
+
```
Braqzen, line 141:
```suggestion

```

Let's try without the tabs / spaces

name-registry/tests/functions/set_owner.rs
```diff
@@ -48,30 +48,29 @@ mod success {
...
     };
     use fuels::prelude::*;
 
     #[tokio::test]
     #[should_panic(expected = "Revert(42)")]
     async fn cant_set_owner() {
         let (instance, _id, wallet, wallet2) = setup().await;
-        let wallet_identity = Identity::Address(Address::from(wallet.address()));
-        let name = String::from("SwaySway");
+        let acc1 = Account::new(wallet);
```
Braqzen, line 59: Why not move this _into_ the `setup()` so that it gets returned as well? 1 fewer line per test

name-registry/src/main.sw
```diff
@@ -0,0 +1,166 @@
...
+    logging::log,
+    storage::StorageMap,
+};
+
+storage {
+    /// A mapping of names to an option of records, with a none representing an unregistered name
+    names: StorageMap<str[8], Option<Record>> = StorageMap {},
+}
+
+const ASSET_ID = ~ContractId::from(ASSET_B256);
```
Braqzen, line 31:
```suggestion
const ASSET_ID = ~ContractId::from(ASSET_B256);

storage {
    /// A mapping of names to an option of records, with a none representing an unregistered name
    names: StorageMap<str[8], Option<Record>> = StorageMap {},
}
```

Only just noticed this.

matt-user (approved):
I have one small suggestion but overall lgtm

name-registry/src/main.sw
```diff
@@ -0,0 +1,166 @@
...
+use std::{
+    block::timestamp,
+    chain::auth::msg_sender,
+    context::call_frames::msg_asset_id,
+    context::msg_amount,
+    logging::log,
+    storage::StorageMap,
+};
+
+const ASSET_ID = ~ContractId::from(ASSET_B256);
```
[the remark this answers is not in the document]
SwayStar123: Looks like it automatically merged when you approved it

PR #240, opened 2022-09-29
Deleted unnecessary redirect file

Braqzen:
I think I understand the jumping around better now. No need for this file at all and I don't know why it even let me push without a PR the first time. It just automatically merged things in


Braqzen:
What, so I can just merge this in anyway. Why doesn't it auto merge like the first time then? I don't understand tools.

PR #245, opened 2022-09-29
Application book has a `run` feature that should be disabled in code snippets

### LINKED ISSUE #244: Application book has a `run` feature that should be disabled in code snippets
### Motivation

According to [mdbook](https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=runnab#outputhtmlplayground) there is a `runnable` parameter that can be set to `false` in order to prevent code snippets from displaying a `run` button (the dark triangle).

When running the book locally this works as expected and the button does not appear.
When looking at the [live](https://fuellabs.github.io/sway-applications/book/documentation/project-quality/documentation/code/abi.html) book every code snippet in the book is runnable.
This does not make sense and is incorrect. They should not be runnable because the code itself is not `Rust` although the snippets use the `Rust` codeblock for highlighting.

Braqzen:
## Type of change

- Bug fix
- Documentation

## Changes

- Removed book param for `runnable`. Hopefully a non-Rust language is not runnable by default
- Changed codeblocks to use `sway` instead of `rust`

## Related Issues

Closes #244

### DIFF
--- a/.docs/contributing-book/book.toml
+++ b/.docs/contributing-book/book.toml
@@ -5,9 +5,6 @@ multilingual = false
 src = "src"
 title = "Contributing to Sway Applications"
 
-[output.html.playground]
-runnable = false
-
 [output.html.fold]
 enable = true
 level = 100     # Allow them to fold but keep it all visible as the default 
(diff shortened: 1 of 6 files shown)


supiket (approved):
_hopefully_

PR #255, opened 2022-10-04
Update which branch the airdrop-distributor is using to build sway-libs

### LINKED ISSUE #254: Sway-Libs updates to master break CI
#### Description

An update to sway-libs was put in place where the Binary Merkle Proof tests were updated from a proof length of 2 to 8. This was not added to any release, yet CI broke due to the use of master as opposed to release.

#### Motiviation 

Fix CI to unblock https://github.com/FuelLabs/sway-applications/pull/249

bitzoic:
## Type of change

- Bug fix

## Changes

- Updates the target for the sway-libs dependency in the Forc.toml to use a tag as opposed to the version.

## Notes

- There should be a way to use libraries without the need to specify the git repository.

Closes #254

### DIFF
--- a/airdrop/airdrop-distributor/Forc.toml
+++ b/airdrop/airdrop-distributor/Forc.toml
@@ -5,4 +5,4 @@ license = "Apache-2.0"
 name = "airdrop-distributor"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Braqzen (approved):
:)

PR #256, opened 2022-10-10
chore: bump forc to v0.25.2 and sdk to v0.25

matt-user:
## Type of change

## Changes

- Updated the CI to forc version 0.25.2
- Update the fuel-rs dependency for each app and fix breaking changes

### DIFF
--- a/NFT/tests/functions/meta_data.rs
+++ b/NFT/tests/functions/meta_data.rs
@@ -1,11 +1,13 @@
 use crate::utils::{
     abi_calls::{constructor, meta_data, mint},
     test_helpers::setup,
-    Identity, TokenMetaData,
+    TokenMetaData,
 };
-use fuels::signers::Signer;
+use fuels::{prelude::Identity, signers::Signer};
 
 mod success {
+
+    use fuels::prelude::SizedAsciiString;
 
     use super::*;
 
@@ -23,7 +25,7 @@ mod success {
         assert_eq!(
             meta_data(&owner1.contract, 0).await,
             TokenMetaData {
-                name: "Example".to_string()
+                name: SizedAsciiString::<7>::new("Example".to_string()).unwrap()
             }
         );
     }
@@ -42,21 +44,21 @@ mod success {
         assert_eq!(
             meta_data(&owner1.contract, 0).await,
             TokenMetaData {
-                name: "Example".to_string()
+                name: SizedAsciiString::<7>::new("Example".to_string()).unwrap()
             }
         );
 
         assert_eq!(
             meta_data(&owner1.contract, 1).await,
             TokenMetaData {
-                name: "Example".to_string()
+                name: SizedAsciiString::<7>::new("Example".to_string()).unwrap()
             }
         );
 
         assert_eq!(
             meta_data(&owner1.contract, 2).await,
             TokenMetaData {
-                name: "Example".to_string()
+                name: SizedAsciiString::<7>::new("Example".to_string()).unwrap()
             }
         );
     }
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ If you wish to run any of the projects then clone this repository and go through
 Any instructions related to running a specific project should be found within the README.md of that project.
 
 > **Note**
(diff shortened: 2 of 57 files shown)


NFT/tests/functions/meta_data.rs
```diff
@@ -1,12 +1,14 @@
...
     test_helpers::setup,
-    Identity, TokenMetaData,
+    TokenMetaData,
 };
-use fuels::signers::Signer;
+use fuels::{prelude::Identity, signers::Signer};
 
 mod success {
 
+    use fuels::prelude::SizedAsciiString;
```
Braqzen, line 10: What is this ugly type?

fundraiser/src/main.sw
```diff
@@ -32,68 +32,89 @@ storage {
     /// Direct look-up for asset data if the user wants to check via a known ID
     asset_info: StorageMap<ContractId,
     AssetInfo> = StorageMap {
-    }, /// O(1) look-up to allow searching via asset_count
+    },
```
Braqzen, line 35: hm, looks like the formatter is broken here @eureka-cpu

NFT/tests/functions/meta_data.rs
matt-user: That's the only way I found to pass in a static sway string.
```
TokenMetaData {
      name: SizedAsciiString::<7>::new("Example".to_string()).unwrap()
}
```

airdrop/airdrop-distributor/tests/functions/end_block.rs
```diff
@@ -5,6 +5,8 @@ use crate::utils::{
 
 mod success {
 
+    use fuels::prelude::Bits256;
```
bitzoic, line 8: Was this meant to be added here or at the top of the file?

README.md
```diff
@@ -67,7 +67,7 @@ If you wish to run any of the projects then clone this repository and go through
 Any instructions related to running a specific project should be found within the README.md of that project.
 
 > **Note**
-> All projects currently use `forc 0.24.4` and `fuel-core 0.10.1`.
+> All projects currently use `forc 0.25.2` and `fuel-core 0.10.1`.
```
bitzoic, line 70: It may be worth considering adding the SDK version as I've done with Sway-Libs

airdrop/airdrop-distributor/tests/functions/end_block.rs
matt-user: yep good catch!

bitzoic: I think there's another one in a different file just as a heads up

README.md
Braqzen: I'm uncertain whether this is a good idea since the `forc` and `core` versions must match however the `sdk` does not. It may be that a new app comes in which requires new `sdk` features but not a newer release of `forc` or `core`. In this case we need to bump all the apps for the tests rather than have the new app use the newer `sdk`.

Braqzen: Furthermore, the most recent commit states that the version is `0.25.1` while the tests actually use `0.25.0`. I think this is a bad idea in general and it should be reverted.

Braqzen: Arbitrarily selecting a version and using a sliding window is a bit arbitrary... that being said my general approach is if nothing actually requires the version to be bumped and there have been significant enough changes / a couple major releases have been released then we should bump `forc`. The tests follow is needed. This is a looser approach to specifying a minimal version each time.

Braqzen: Ultimately I'd like each project to be completely self contained and have it be managed in the same way so that each project does not affect any other project, the decoupling makes PRs easier to handle, and versions are somewhat frequently updated so that nothing lags behind too much while not spending too much time constantly bumping.

matt-user: Yeah I agree that having each project self contained would be easier to mange PRs.  Also reading your first (@Braqzen ) comment I agree with you.  Perhaps a good rule of thumb for updating the SDK versions could be whenever a breaking change is introduced into the rust SDK.

bitzoic (approved):
lgtm!

Braqzen (approved):

PR #258, opened 2022-10-12
chore: update to latest sdk version

matt-user:
## Type of change

## Changes

- Updated the escrow tests to work with the newest sdk version


escrow/tests/functions/create_escrow.rs
```diff
@@ -118,16 +120,16 @@ mod revert {
...
+        // create_escrow(
+        //     defaults.asset_amount,
+        //     &arbiter_obj,
+        //     &defaults.asset_id,
+        //     vec![],
+        //     buyer.wallet.address(),
+        //     &seller.contract,
+        //     defaults.deadline,
+        // )
+        // .await;
```
Braqzen, line 132:
```suggestion
        create_escrow(
            defaults.asset_amount,
            &arbiter_obj,
            &defaults.asset_id,
            vec![],
            buyer.wallet.address(),
            &seller.contract,
            defaults.deadline,
        )
        .await;
```

Braqzen (changes requested):

escrow/tests/functions/create_escrow.rs
matt-user: Actually I can't do this because the create escrow function expects an array or 2 Assets and not a vec.  I think it makes the most sense to comment the code out instead of having code that passes 2 assets.

matt-user: Nvm I'm going to change the contract to use a vec and update the tests accordingly

escrow/src/data_structures.sw
```diff
@@ -58,14 +58,14 @@ pub struct EscrowInfo {
...
-        assets: [Asset; 2],
+        assets: Vec<Asset>,
         buyer: Identity,
         deadline: u64,
         seller: Identity,
     ) -> Self {
         Self {
             arbiter,
-            assets,
+            assets: [assets.get(0).unwrap(), assets.get(1).unwrap()],
```
Braqzen, line 68: This should not be hardcoding 2 assets, it should take the entire array

escrow/src/main.sw
```diff
@@ -138,12 +137,12 @@ impl Escrow for Contract {
...
         let mut index = 0;
-        // while index < escrow.assets.len() {
-        // let asset = escrow.assets.get(index).unwrap();
+        //while index < escrow.assets.len() {
+            //let asset = escrow.assets.get(index).unwrap();
         while index < 2 {
             let asset = escrow.assets[index];
-
             if asset.id == msg_asset_id() {
+                log(69);
```
Braqzen, line 145: This log should be removed

escrow/tests/functions/create_escrow.rs
```diff
@@ -98,6 +98,7 @@ mod revert {
 
     use super::*;
 
+    // TODO marked as ignore until we use vec in the contract and can properly test this case
```
Braqzen, line 101: This is not marked as ignored...

Braqzen (changes requested):

escrow/src/data_structures.sw
matt-user: We can't use vecs in structs yet so `assets` in `EscrowInfo` needs to be a hardcoded array.  The only alternative I can think of is to move `assets` out of the struct and use it separately in storage.

Braqzen: I believe we track the ID of the escrow, correct? What is the issue with moving the vector out of the struct and having an additional storage variable to index?

matt-user: Code quality I guess.  But functionality probably trumps such a nit.

matt-user: We can't use storage containers within each other yet so we can't track the storage vector as a separate storage variable inside of a storage map, so I think we are stuck with a static array inside EscrowInfo for now.

Braqzen: I don't understand.

matt-user: So ideally we would like the `assets` variable inside of `EscrowInfo` to be a `StorageVec<Asset>` but [this](https://github.com/FuelLabs/sway/issues/2465) is blocking us.  A workaround is to have a separate storage variable `assets" StorageMap<u64, StorageVec<Asset>`, but [this](https://github.com/FuelLabs/sway/issues/2639) is blocking as well.

matt-user: And assets can't be a regular vec it has to be a `StorageVec`

Braqzen: There is a solution but it's nasty and I'd rather not go down that path but then again I'd rather not go down this one either.

Struct keeps track of the number of assets and the index of the storagevec where the first asset is. You can then iterate from that point to get the assets.

Thoughts?

matt-user: Not a huge fan, but at this point I think it's the only option to get the contract fully functional which would be nice for the frontend and cool in general.  The Sway implementation should be pretty straightforward.  To display an Escrow's Asset information this way from the frontend perspective might be a bit of a pain although I think I could implement a temporary getter function for the contract to avoid this pain until the blockers are resolved.

escrow/src/main.sw
```diff
@@ -138,11 +145,8 @@ impl Escrow for Contract {
         // TODO: https://github.com/FuelLabs/sway/issues/2014
         //       `.contains() -> bool / .position() -> u64` would clean up the loop
```
bitzoic, line 146: Is this comment still valid since you're using the `.get()` function below? Not entirely sure what this is referring to here.

bitzoic (commented):
Could you add a test with more than 2 assets since there is the use of a `vec` now?

escrow/src/main.sw
Braqzen: Yes, the idea was to say "is `msg_asset_id()` in here?". This way the loop is abstracted into the vec and I can remove the loop in the contract itself.

matt-user:
> Could you add a test with more than 2 assets since there is the use of a `vec` now?

I modified the `deposit_to_two_escrows` test to test for this case.  lmk if you think this should be separated into its own test case instead

Braqzen:
> > Could you add a test with more than 2 assets since there is the use of a `vec` now?
> 
> I modified the `deposit_to_two_escrows` test to test for this case. lmk if you think this should be separated into its own test case instead

The cases that we'd like to cover are:

- No assets revert
- 1 asset
- 2 assets in 1 escrow
- 1 asset per escrow, we only need 2 different escrows

This way we can cover unit and system testing

matt-user:
> > > Could you add a test with more than 2 assets since there is the use of a `vec` now?
> > 
> > 
> > I modified the `deposit_to_two_escrows` test to test for this case. lmk if you think this should be separated into its own test case instead
> 
> The cases that we'd like to cover are:
> 
> * No assets revert
> * 1 asset
> * 2 assets in 1 escrow
> * 1 asset per escrow, we only need 2 different escrows
> 
> This way we can cover unit and system testing

`when_assets_are_not_specified` should already take care of the first test case you mentioned.  Also I think it is appropriate to remove the TODO on line 154 now.

The rest of the tests you mentioned should be covered now in the `success` mod in `create_escrow`.  Although I used 3 assets instead of 2 for the 3 test case you mentioned.

bitzoic (dismissed):
Could you update the APPLICATION_PROGRESS.md to reflect the use of `vec` now? 
Other than that lgtm!

Braqzen (approved):

bitzoic (approved):

PR #267, opened 2022-10-17
AMM scripts: Swap assets with exact input & swap assets with exact output along a route

### LINKED ISSUE #229: AMM: Swap exact input and swap exact output scripts
#### Description

Swap exact input script that:
- Takes an asset route (i.e. a vector of asset IDs) and a vector of amounts as input
- Calls `swap_exact_input` function ([ABI and documentation](https://github.com/FuelLabs/sway-applications/blob/master/AMM/project/libraries/src/interface.sw#L108-L123), [implementation](https://github.com/FuelLabs/sway-applications/blob/master/AMM/project/contracts/exchange-contract/src/main.sw#L226-L260)) for each subsequent asset pair to perform potentially (i.e. if the route has more than two assets) multiple swaps in a row 
- Returns the bought amount of the last asset

Swap exact output script that:
- Takes an asset route (i.e. a vector of asset IDs) and a vector of amounts as input
- Calls `swap_exact_output` function ([ABI and documentation](https://github.com/FuelLabs/sway-applications/blob/master/AMM/project/libraries/src/interface.sw#L125-L143), [implementation](https://github.com/FuelLabs/sway-applications/blob/master/AMM/project/contracts/exchange-contract/src/main.sw#L262-L300)) for each subsequent asset pair to perform potentially (i.e. if the route has more than two assets) multiple swaps in a row 
- Returns the sold amount of first asset

#### Motivation

These scripts will allow swapping two assets that do not have a direct liquidity pool.

However, the frontend of the AMM app (#233) may use this script to perform any swap operation, e.g. even if two assets have a pool or the asset route consists of only two assets. Another use case of this script can be finding an optimal route while swapping one asset for another, whether the assets have a pool or not. These routes should be calculated in the frontend and passed to the script.

supiket:
## Type of change

- New feature

## Changes

- Added a script that takes a path of assets and swaps with exact input.

## Notes

- Uses [feat: support main function arguments in scripts #620](https://github.com/FuelLabs/fuels-rs/pull/620)
   - Vector inputs are not yet supported, so the script takes assets individually. Will clarify the plan on this soon.
   - Test is failing right now due to not being able to set external contracts as dependencies, addressed in [this issue](https://github.com/FuelLabs/fuels-rs/issues/667).

## Related Issues

Closes #229


supiket:
The work for this functionality is now in #312.

Issue #14, opened 2022-05-05
Fuel Signatures


Braqzen:

K1-R1:
Closed as of https://github.com/FuelLabs/sway-applications/pull/272

Issue #12, opened 2022-05-05
Full test coverage


Braqzen:
All of the features in [Ethereum Signature Capability](https://github.com/FuelLabs/sway-applications/issues/1) should be fully tested.

- [x] #10 
- [x] #11

Braqzen:
Appears to be completed in #272.

Issue #11, opened 2022-05-05
EIP-191 signature support


Braqzen:
Support payloads (signatures) constructed from Metamask, Trezor & Ledger.

adlerjohn:
I don't think separate issues are needed for this. What's really needed is simply [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signature support, not specifically supporting different wallets.

K1-R1:
Closed as of https://github.com/FuelLabs/sway-applications/pull/272

PR #272, opened 2022-10-22
Multisig-wallet ETH signature support

### LINKED ISSUE #10: Supports both Ethereum & Fuel signature recovery

K1-R1:
## Type of change

- New feature

## Changes

- This PR adds the ability for both Fuel and EVM addresses to be owners/users of the multisig-wallet; by enabling signature recovery from signatures over EIP-191 formatted and Ethereum prefixed message hashes.
- It additionally adds testing and updated documentation.

## Related Issues

Closes #10


eth-account-abstraction/src/utils.sw
```diff
@@ -0,0 +1,56 @@
...
+    },
+    identity::Identity,
+    vm::evm::evm_address::EvmAddress,
+};
+
+//Applies the prefix used by Geth to a message hash.
+//Returns the prefixed hash.
+pub fn eth_prefix(msg_hash: b256) -> b256 {
+    let prefix = "\x19Ethereum Signed Message:\n32";
+    sha256((prefix, msg_hash))
```
adlerjohn, line 18: Not sure this will work correctly. See https://github.com/FuelLabs/sway/issues/2594#issuecomment-1221205410 for more context. It might be worth adding a bytes library as a requirement to implement this correctly.

K1-R1: Ah, I think I see; so the library would need to give an unpadded reference type for correct hashing with sha256?

K1-R1:
This PR additionally closes the following issues:
- Closes #7 
- Closes #11 
- Closes #12 
- Closes #13 
- Closes #14 

I have not included these in the description as to make reviewing easier.

Braqzen:
> This PR additionally closes the following issues:
> 
> * Closes [Full test coverage #7](https://github.com/FuelLabs/sway-applications/issues/7)
> * Closes [EIP-191 signature support #11](https://github.com/FuelLabs/sway-applications/issues/11)
> * Closes [Full test coverage #12](https://github.com/FuelLabs/sway-applications/issues/12)
> * Closes [Ethereum Signatures #13](https://github.com/FuelLabs/sway-applications/issues/13)
> * Closes [Fuel Signatures #14](https://github.com/FuelLabs/sway-applications/issues/14)
> 
> I have not included these in the description as to make reviewing easier.

I'll review this sometime tomorrow. Note that if you do not add these into the description they will not be closed once this PR is merged.

K1-R1:
No worries, I thought it'd be easier to review like this and then I'll manually close the relevant issues after

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,92 +1,73 @@
...
 
-// Standard library code
 use std::{
-    b512::B512,
-    call_frames::contract_id,
     constants::ZERO_B256,
     context::this_balance,
-    ecr::ec_recover_address,
-    hash::sha256,
+    identity::Identity,
```
matt-user, line 15: this should be included in the prelude so no need to import

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,92 +1,73 @@
...
     logging::log,
-    token::{
-        force_transfer_to_contract,
-        transfer_to_address,
-    },
+    revert::require,
```
matt-user, line 17: included in the prelude no need to import

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,92 +1,73 @@
...
 use events::{ExecutedEvent, TransferEvent};
+use utils::{create_hash, recover_signer};
 
 storage {
     /// Used to add entropy into hashing of Tx to decrease the probability of collisions / double
-    /// spending
+    /// spending.
     nonce: u64 = 0,
-    /// The number of approvals required in order to execture a Tx
+    /// The number of approvals required in order to execture a Tx.
```
matt-user, line 31:
```suggestion
    /// The number of approvals required in order to execute a Tx.
```

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+
+use std::{
+    call_frames::contract_id,
+    ecr::{
+        ec_recover_address,
+    },
+    hash::{
+        sha256,
+    },
+    identity::Identity,
```
matt-user, line 13: no need to import included in prelude

matt-user (changes requested):
Just took a quick look at the code (haven't looked at tests yet).  One addition I would make to the docs is to link EIP 191 somewhere in them as it is mentioned frequently in the code.  Also I don't see the spec in here, but I may have missed it.

Braqzen (changes requested):
1. SPECIFICATION.md?
2. Imports from the [prelude](https://fuellabs.github.io/sway/v0.31.1/introduction/standard_library.html#standard-library-prelude) should not be explicitly imported because they are automatically imported for us. Some imports have been pointed out in the previous review but not all files
3. When importing a single component from somewhere it should not be wrapped inside `{}` e.g. `utils.sw` for `erc`
4. The interface documentation references `Panics` in a few areas rather than `Reverts`
5. `mod.rs` uses `unsafe {}` blocks. Given that it comes from that is likely well tested it's probably fine; however, this may be problematic for maintenance because you're saying "I have checked this, trust me". Is there a workaround to not use `unsafe` code?
6. There are comments in `mod.rs` function `format_and_sign` that do not insert a space between the `//` and text proceeding e.g. `//Format` instead of `// Format`. That aside, those comments do not convey any insight into what is going on. It's clear enough from the code to see that something is formatted etc. Those comments should be removed or elaborated upon. This may apply to other files as well hence I'm writing it here instead of that specific section.
7. There are no tests for `execute_transaction`

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+
+    let prefixed_message = match signature_data.prefix {
+        MessagePrefix::None => formatted_message,
+        MessagePrefix::Ethereum => ethereum_prefix(formatted_message),
+    };
+
+    match signature_data.wallet_type {
+        WalletType::Fuel => {
+            let recover_result = ec_recover_address(signature_data.signature, prefixed_message);
+            require(recover_result.is_ok(), "ec_recover_address failed");
```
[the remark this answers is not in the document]
nfurfaro: If `ec_recover_address` fails, it already propagates an `EcRecoverError::UnrecoverablePublicKey` enum, so we could just use that here.

nfurfaro (commented):
Looking really good! So nice to see this feature on the brink of being merged. 
I left a few comments, mostly trivial stuff but one which may reduce the number of storage reads being used.

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,92 +1,73 @@
...
-        while user_index < 25 {
-            require(Address::from(ZERO_B256) != users[user_index].identity, InitError::AddressCannotBeZero);
-            require(users[user_index].weight != 0, InitError::WeightingCannotBeZero);
-            storage.weighting.insert(users[user_index].identity, users[user_index].weight);
-            user_index = user_index + 1;
+        while user_index < users.len() {
+            require(ZERO_B256 != users.get(user_index).unwrap().address, InitError::AddressCannotBeZero);
+            require(users.get(user_index).unwrap().weight != 0, InitError::WeightingCannotBeZero);
+
+            storage.weighting.insert(users.get(user_index).unwrap().address, users.get(user_index).unwrap().weight);
```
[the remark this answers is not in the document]
nfurfaro: Ah, ignore that last comment, I see that this is not doing what I initially thought and is not reading from storage at all.

multisig-wallet/project/multisig-contract/src/data_structures.sw
```diff
@@ -1,24 +1,50 @@
...
+
+pub enum MessageFormat {
+    None: (),
+    EIP191PersonalSign: (),
+}
+
+pub enum MessagePrefix {
+    None: (),
+    Ethereum: (),
+}
```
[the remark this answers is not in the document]
K1-R1: The `None` variant is used to specify that a message isn't formatted/prefixed. This allows `ec_recovery` from a 
variety of scenarios:

-  A signature from a Fuel wallet signing over a message that _was_ [EIP-191](https://eips.ethereum.org/EIPS/eip-191) formatted but _didn't_ have the Ethereum prefix applied.
- A signature from a Fuel wallet signing over a message that _wasn't_ formatted and _didn't_ have the Ethereum prefix applied.
- A signature from an EVM wallet signing over a message that _was_ [EIP-191](https://eips.ethereum.org/EIPS/eip-191) formatted and _had_ the Ethereum prefix applied.

These enums are used for the control flow in `utils.sw::recover_signer`.

multisig-wallet/project/multisig-contract/src/interface.sw
```diff
@@ -0,0 +1,85 @@
...
+    #[storage(read, write)]
+    fn constructor(threshold: u64, users: Vec<User>);
+
+    /// Execute a transaction formed from the `to, `value` and `data` parameters if the signatures meet the
+    /// threshold requirement.
+    ///
+    /// # Arguments
+    ///
+    /// * 'data' - The data field of the transaction.
+    /// * 'signatures_data' - The information for each user's signature for a specific transaction.
```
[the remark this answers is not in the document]
K1-R1: I see, I agree on the parameters being changed to `signatures`, but I think the struct being called `Signature` is also a bit confusing as it has the field `signature`.
Perhaps `SignatureInfo` for the struct?

multisig-wallet/project/multisig-contract/src/interface.sw
```diff
@@ -0,0 +1,85 @@
...
+use std::{contract_id::ContractId, identity::Identity};
+
+use data_structures::{SignatureData, User};
+
+abi MultiSignatureWallet {
+    /// The constructor initializes the necessary values and unlocks further functionlity.
+    ///
+    /// # Arguments
+    ///
+    /// * 'threshold' - The number of approvals required to enable a transasction.
```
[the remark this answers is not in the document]
K1-R1: Backticks are back

multisig-wallet/project/multisig-contract/src/interface.sw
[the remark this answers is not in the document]
Braqzen: Sure

multisig-wallet/project/multisig-contract/src/data_structures.sw
```diff
@@ -1,24 +1,50 @@
...
-    /// Amount of asset
+    /// Amount of asset.
     value: u64,
 }
+
+pub struct User {
+    /// The wallet address of a user.
+    address: b256,
+    /// Number of approvals the user provides when approving.
+    /// The default is usually 1.
```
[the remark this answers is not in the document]
Braqzen: When I was writing the initial documentation I believe the default multisig wallet implementation would not use weighting and thus the default is technically 1, 1 per signature. I may have been incorrect.

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+/// Creates an EIP-191 compliant transaction hash, of the version:
+/// 0x45, personal sign.
+/// It takes a data_to_sign to represent the <data to sign> in the following EIP-191 format:
+/// 0x19 <1 byte version> <version specific data> <data to sign>
+fn eip_191_personal_sign_format(data_to_sign: b256) -> b256 {
+    let initial_byte = 0x19u8;
+    let version_byte = 0x45u8;
```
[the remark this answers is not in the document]
Braqzen: Yes, on the one hand that may save memory per call but on the other hand it brings them into global scope where everything has access. I _assume_ that `const` is better for runtime cost.

multisig-wallet/project/multisig-contract/src/data_structures.sw
```diff
@@ -1,24 +1,50 @@
...
+    /// Value used to prevent double spending.
     nonce: u64,
-    /// Amount of asset
+    /// Amount of asset.
     value: u64,
 }
+
+pub struct User {
+    /// The wallet address of a user.
+    address: b256,
```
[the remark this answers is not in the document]
Braqzen: This is something that I've noticed but it slipped my mind. Looking at the code it's been changed to save on some conversions so it should make it slightly cheaper to use and possibly deploy. The `Address` is correct thought because it prevents someone from accidentally using the `b256` of a contract as a user and then finding out that they cannot sign etc. For clarity and correctness, even at a slight cost, `Address` should be used.

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,212 +1,141 @@
...
-    /// - When the total approval count is less than the required threshold for execution
     #[storage(read, write)]
-    fn execute_transaction(to: Identity, value: u64, data: b256, signatures: [B512; 25]) {
+    fn execute_transaction(
+        data: b256,
+        signatures: Vec<SignatureInfo>,
+        to: Identity,
+        value: u64,
+    ) {
         require(storage.nonce != 0, InitError::NotInitialized);
```
simonr0204, line 56: Should we also check that transactions are being executed in nonce order? There are scenarios where a signer would want to approve a transaction on the condition that all previous transactions in the queue have already been executed. Many multisig wallets (e.g. Gnosis Safe) have this requirement.

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+fn decompose(val: b256) -> (u64, u64, u64, u64) {
+    asm(r1: __addr_of(val)) { r1: (u64, u64, u64, u64) }
+}
+
+/// Applies the prefix used by Geth to a message hash.
+/// Returns the prefixed hash.
+fn ethereum_prefix(msg_hash: b256) -> b256 {
+    let prefix = "\x19Ethereum Signed Message:\n32";
+
+    sha256((prefix, msg_hash))
```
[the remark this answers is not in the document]
simonr0204: hmm yes I think this should be keccak.

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,212 +1,141 @@
...
+        while user_index < users.len() {
+            require(ZERO_B256 != users.get(user_index).unwrap().address, InitError::AddressCannotBeZero);
+            require(users.get(user_index).unwrap().weight != 0, InitError::WeightingCannotBeZero);
+
+            storage.weighting.insert(users.get(user_index).unwrap().address, users.get(user_index).unwrap().weight);
+            user_index += 1;
         }
 
         storage.nonce = 1;
         storage.threshold = threshold;
```
simonr0204, line 46: Should check that threshold <= sum of weights, otherwise wallet can be initialized in an unusable state.

simonr0204:
I think this would be the object of a future PR, but we should consider whether there's functionality here that should be lifted into `std::vm::evm` . Particularly the EIP191 prefix stuff.

multisig-wallet/project/multisig-contract/src/main.sw
K1-R1: As it stands, I don't think it's possible for transactions to be executed in anything but nonce order; the user never inputs a nonce value to either `transfer` or `execute_transaction`, the value is only ever called from storage when executing a transaction. 
A function to cancel an order ( by incrementing the nonce ) would be helpful; I'll add that in

multisig-wallet/project/multisig-contract/src/utils.sw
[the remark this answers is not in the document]
K1-R1: 👍

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+
+/// Creates an EIP-191 compliant transaction hash, of the version:
+/// 0x45, personal sign.
+/// It takes a data_to_sign to represent the <data to sign> in the following EIP-191 format:
+/// 0x19 <1 byte version> <version specific data> <data to sign>
+fn eip_191_personal_sign_format(data_to_sign: b256) -> b256 {
+    let initial_byte = 0x19u8;
+    let version_byte = 0x45u8;
+
+    let encoded_data = encode_data(initial_byte, version_byte, data_to_sign);
```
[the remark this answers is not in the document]
K1-R1: I've changed these to `encode_and_pack_signed_data` and `signed_data`, as the `EIP-191` docs refer to this data as `signed_data`.
Open to suggestions if this is still unclear.

multisig-wallet/project/multisig-contract/src/main.sw
simonr0204: That makes sense. It's a feature that's rarely used in practice, but in a scenario where the signers don't all trust each other, being able to explicitly render a previously shared signature useless (by using its nonce) is necessary.

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,113 @@
...
+    data.push((message_2 << 48) + (message_3 >> 16));
+    data.push((message_3 << 48) + (message_4 >> 16));
+    data.push(message_4 << 48);
+
+    data
+}
+
+/// Get a tuple of 4 u64 values from a single b256 value.
+fn decompose(val: b256) -> (u64, u64, u64, u64) {
+    asm(r1: __addr_of(val)) { r1: (u64, u64, u64, u64) }
```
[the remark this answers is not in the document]
K1-R1: Yes

multisig-wallet/project/multisig-contract/src/utils.sw
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Applies the prefix used by Geth to a message hash.
```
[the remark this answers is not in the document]
K1-R1: No; EIP-191 isn't and nor is the prefix, but the prefix originates from Geth as I understand it.

multisig-wallet/project/multisig-contract/src/utils.sw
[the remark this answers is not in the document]
K1-R1: Yh you two are correct; both ethers-rs and go-ethereum use keccak256. Looks like the older references I was using are just outdated or wrong. Will switch to keccack256

multisig-wallet/project/multisig-contract/src/utils.sw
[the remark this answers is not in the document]
simonr0204: Presumably there are other ethereum clients that implement this signature scheme though? I think the prefixes are a standard across implementations. It may be that the client usually attaches the prefix but it's not necessarily the case. One could prepare a signature without using any client software if they follow the spec.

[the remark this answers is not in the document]
K1-R1: Yh; this scheme is common across clients and wallets, but can be followed manually with the spec. I meant that as the EIP-191 document itself states:

Additionally, 0x19 has been chosen because since ethereum/go-ethereum#2940 , the following is prepended before hashing in personal_sign:
`"\x19Ethereum Signed Message:\n" + len(message)`

It seemed that it's use, with regards to EIP-191, stems from it's integration into Geth

multisig-wallet/project/multisig-contract/tests/functions/transaction_hash.rs
```diff
@@ -0,0 +1,103 @@
...
+        let tx = Transaction {
+            contract_identifier: contract.get_contract_id().try_into().unwrap(),
+            data,
+            destination: to.clone(),
+            nonce,
+            value,
+        };
+
+        //Set tokens for encoding with ABIEncoder
+        let contract_identifier_token = Token::Struct(vec![Token::B256(
```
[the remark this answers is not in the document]
K1-R1: I'm unaware of another way to match Sway's encoding without just hardcoding a `u8` array. I can look more deeply into this; opening an issue on the SDK and if a simpler solution arrises I can add it in a future PR. 

Assuming there currently isn't a simpler approach, it's a decent example of manually using the SDK's `ABIEncoder`.

multisig-wallet/project/multisig-contract/tests/functions/transfer.rs
```diff
@@ -0,0 +1,464 @@
...
+        // Assertions
+        assert_eq!(initial_contract_balance, 200);
+        assert_eq!(initial_receiver_balance, 0);
+
+        assert_eq!(final_contract_balance, 0);
+        assert_eq!(final_receiver_balance, 200);
+    }
+}
+
+mod revert {
```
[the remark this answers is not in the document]
K1-R1: I've written a test for unrecoverable key, but I don't think it's panicking for the expected reason. I'd have to make the `ecr` opcode panic, and I couldn't see any examples of this in the Sway repo tests. I'll do some digging on this, but won't add the test until it's definitely working

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+            coins_per_asset,
+            amount_per_coin,
+        );
+
+        let (provider, _socket_addr) = setup_test_provider(coins.clone(), vec![], None, None).await;
+
+        wallet.set_provider(provider);
+
+        let contract_id = Contract::deploy(
+            "./out/debug/multisig-contract.bin",
```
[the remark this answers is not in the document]
K1-R1: I'll add a paths module

K1-R1:
> 5\. `mod.rs` uses `unsafe {}` blocks. Given that it comes from that is likely well tested it's probably fine; however, this may be problematic for maintenance because you're saying "I have checked this, trust me". Is there a workaround to not use `unsafe` code?

Currently theres no workaround, but there is an open issue related to removing the use of unsafe here:
https://github.com/FuelLabs/fuels-rs/issues/698
I'll reference the issue in `mod.rs`

K1-R1:
> 7\. There are no tests for `execute_transaction`

Currently `execute_transaction` doesn't do anything, as it requires the ability to make arbitrary contract calls, which isn't yet supported.

Braqzen:
> > 7. There are no tests for `execute_transaction`
> 
> Currently `execute_transaction` doesn't do anything, as it requires the ability to make arbitrary contract calls, which isn't yet supported.

Template tests should be written and an `#[ignore]` should be added with comments in the tests to describe why we cannot fully test at the moment. If these are not added then someone has to remember that the entire suite is missing. If the tests are partially written, or at least stubbed, then when the tests are run they will come up as ignored and anyone will be able to pick on the tests needing to be written when possible.

K1-R1:
I've bumped the versions on this PR to align with #325, CI will fail until it's merged.

multisig-wallet/project/multisig-contract/tests/functions/balance.rs
```diff
@@ -0,0 +1,39 @@
...
+            )
+            .await
+            .unwrap();
+
+        let final_balance = balance(&deployer.contract, base_asset_contract_id())
+            .await
+            .value;
+
+        assert_eq!(initial_balance, 0);
+        assert_eq!(final_balance, DEFAULT_TRANSFER_AMOUNT);
```
Braqzen, line 37: Looking at the imported constant we can see that the default is non-zero; however, this test itself does not guarantee that there is any change in the balance before and after because it does not assert that the initial balance and final balance are not the same. 1 more assertion should be added to check this because someone may change the constant in the other file and this test would still pass.

multisig-wallet/project/multisig-contract/tests/functions/cancel_transaction.rs
```diff
@@ -0,0 +1,41 @@
...
+        constructor(&deployer.contract, constructor_users(), DEFAULT_THRESHOLD).await;
+
+        let initial_nonce = nonce(&deployer.contract).await.value;
+
+        cancel_transaction(&deployer.contract).await;
+
+        let final_nonce = nonce(&deployer.contract).await.value;
+
+        assert_eq!(initial_nonce, 1);
+        assert_eq!(final_nonce, 2);
```
Braqzen, line 24:
```suggestion
        assert_eq!(final_nonce, initial_nonce + 1);
```

multisig-wallet/project/multisig-contract/tests/functions/constructor.rs
```diff
@@ -0,0 +1,86 @@
...
+            .unwrap(),
+            weight: 3,
+        };
+
+        constructor(&deployer.contract, users, DEFAULT_THRESHOLD).await;
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "ThresholdCannotBeZero")]
+    async fn threshold_cannot_be_zero() {
```
Braqzen, line 56: I would order the tests to be in the same order as the requirements so that it's easier to see requirement by requirement which test should be tested and thus if anything is missing

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        let non_owner = Caller {
+            contract: MultiSigContract::new(multisig_contract_id, non_owner_wallet.clone()),
+            wallet: non_owner_wallet,
+        };
+
+        Ok((private_key, deployer, non_owner))
+    }
+
+    pub fn constructor_users() -> Vec<User> {
```
Braqzen, line 172: Unsure about the name, what about something like `default_users`?

multisig-wallet/project/multisig-contract/tests/functions/constructor.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        constructor(&deployer.contract, users, DEFAULT_THRESHOLD).await;
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "TotalWeightCannotBeLessThanThreshold")]
+    async fn total_weight_cannot_be_less_than_threshold() {
+        let (_private_key, deployer, _non_owner) = setup_env(VALID_SIGNER_PK).await.unwrap();
+
+        constructor(&deployer.contract, constructor_users(), 100).await;
```
Braqzen, line 84: This is nice for a one liner; however, you can make the suite more flexible by counting the weight of the users that are being provided into the constructor and then do `weight + 1` for the threshold. This way you can dynamically alter the weights in the setup without worry of brittle tests.

multisig-wallet/project/multisig-contract/tests/functions/execute_transaction.rs
```diff
@@ -0,0 +1,39 @@
...
+    async fn executes_transaction() {
+        // Currently ignored and unimplemented, awaiting the ability to implement https://github.com/FuelLabs/sway-applications/issues/22
+    }
+}
+
+mod revert {
+
+    #[ignore]
+    #[tokio::test]
+    #[should_panic]
```
Braqzen, line 14: I would also add the `expected = "Error"` to make it even clearer for future devs which line should fail.

multisig-wallet/project/multisig-contract/tests/functions/nonce.rs
```diff
@@ -0,0 +1,24 @@
...
+        let (_private_key, deployer, _non_owner) = setup_env(VALID_SIGNER_PK).await.unwrap();
+
+        let initial_nonce = nonce(&deployer.contract).await.value;
+
+        constructor(&deployer.contract, constructor_users(), DEFAULT_THRESHOLD).await;
+
+        let final_nonce = nonce(&deployer.contract).await.value;
+
+        assert_eq!(initial_nonce, 0);
+        assert_eq!(final_nonce, 1);
```
Braqzen, line 22:
```suggestion
        assert_eq!(final_nonce, initial_nonce + 1);
```

multisig-wallet/project/multisig-contract/tests/functions/transfer.rs
```diff
@@ -0,0 +1,274 @@
...
+            .unwrap()
+            .get_asset_balance(receiver_wallet.address(), BASE_ASSET_ID)
+            .await
+            .unwrap();
+
+        assert_eq!(initial_contract_balance, 200);
+        assert_eq!(initial_receiver_balance, 0);
+
+        assert_eq!(final_contract_balance, 0);
+        assert_eq!(final_receiver_balance, 200);
```
Braqzen, line 90: Could use `DEFAULT_TRANSFER_AMOUNT`, no? Also similar comment where the pre-condition is different to the post-condition.

multisig-wallet/project/multisig-contract/tests/functions/transfer.rs
```diff
@@ -0,0 +1,274 @@
...
+    use super::*;
+
+    #[tokio::test]
+    #[should_panic(expected = "NotInitialized")]
+    async fn not_initialized() {
+        let (private_key, deployer, _non_owner) = setup_env(VALID_SIGNER_PK).await.unwrap();
+
+        let (_receiver_wallet, receiver, data) = transfer_parameters();
+
+        deployer
```
Braqzen, line 105: Not needed

multisig-wallet/project/multisig-contract/src/main.sw
```diff
@@ -1,212 +1,168 @@
...
-    ///
-    /// # Panics
-    ///
-    /// - When the constructor is called more than once
-    /// - When the user address is the 0th address (0x00000...)
-    /// - When the threshold is set to 0
-    /// - When an owner has an approval weight of 0
     #[storage(read, write)]
-    fn constructor(users: [User; 25], threshold: u64) {
+    fn cancel_transaction() {
```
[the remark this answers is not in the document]
simonr0204: It's really for "peace of mind" for a signer who has previously signed a transaction (off-chain, but that signed transaction is now out "in the wild" ), and would like to be certain that the transaction can _never_ be executed. So they spend the nonce on-chain.

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+};
+
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
+use sha3::{Digest, Keccak256};
+
+abigen!(
+    MultiSigContract,
+    "./project/multisig-contract/out/debug/multisig-contract-abi.json"
+);
```
Braqzen, line 15:
```suggestion
abigen!(
    MultiSig,
    "./project/multisig-contract/out/debug/multisig-contract-abi.json"
);
```

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+pub mod paths {
+    pub const MULTISIG_CONTRACT_BINARY_PATH: &str = "./out/debug/multisig-contract.bin";
+    pub const MULTISIG_CONTRACT_STORAGE_PATH: &str =
+        "./out/debug/multisig-contract-storage_slots.json";
+}
+
+pub struct Caller {
+    pub contract: MultiSigContract,
+    pub wallet: WalletUnlocked,
+}
```
Braqzen, line 29:
```suggestion
pub struct Caller {
    pub contract: MultiSigContract,
    pub wallet: WalletUnlocked,
}

pub mod paths {
    pub const MULTISIG_CONTRACT_BINARY_PATH: &str = "./out/debug/multisig-contract.bin";
    pub const MULTISIG_CONTRACT_STORAGE_PATH: &str =
        "./out/debug/multisig-contract-storage_slots.json";
}
```

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,300 @@
+use fuels::{
+    contract::call_response::FuelCallResponse,
+    prelude::*,
+    signers::fuel_crypto::{Message, SecretKey, Signature},
+    tx::{Bytes32, Bytes64},
+};
+
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
```
Braqzen, line 9:
```suggestion
```

multisig-wallet/project/multisig-contract/tests/utils/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub const MULTISIG_CONTRACT_STORAGE_PATH: &str =
+        "./out/debug/multisig-contract-storage_slots.json";
+}
+
+pub struct Caller {
+    pub contract: MultiSigContract,
+    pub wallet: WalletUnlocked,
+}
+
+pub mod abi_calls {
```
Braqzen, line 31:
```suggestion
pub mod interface {
```

multisig-wallet/project/multisig-contract/tests/functions/cancel_transaction.rs
```diff
@@ -0,0 +1,41 @@
...
+    test_helpers::{default_users, setup_env, DEFAULT_THRESHOLD},
+    VALID_SIGNER_PK,
+};
+
+mod success {
+
+    use super::*;
+
+    #[tokio::test]
+    async fn cancels_transaction() {
```
Braqzen, line 12: Can we check that the tests also emit the correct logs? [Example](https://github.com/FuelLabs/sway-applications/pull/322/files#diff-3bf4205ff279244340076b24dce7c26d8ba6c8767c54a9c47566046f3d83c09eR37-R42)

Braqzen (approved):

matt-user (commented):
lgtm, will approve once more people get a chance to review

Braqzen (approved):

simonr0204 (approved):

PR #273, opened 2022-10-25
Resolved Issues #247 and #268.

### LINKED ISSUE #247: Remove `CONTRIBUTING.md` from repo README block
### Motivation

[Here](https://github.com/FuelLabs/sway-applications#repository-structure) is a block which shows the repo structure. The contributing section no longer a markdown file and has been removed from the repo. This block of text should no longer reference that file because it does not exist.

Remove `CONTRIBUTING.md` from the block.

### LINKED ISSUE #268: Remove outdated comment in Escrow
### Motivation

The following line no longer applies and it should be deleted.

https://github.com/FuelLabs/sway-applications/blob/adec21e57e7b8dc09e881fbe24d8fdea1124b49f/escrow/src/main.sw#L3

GeckoSplat:
## Type of change

- Other :
- Resolved Issues #247 and #268. Removed text as stated in issues.

## Changes

- Change 1 : Issue #247 resolved. Removed line 46.
- Change 2: Issue #268 resolved. Removed line 3.

Closes #247 
Closes #268

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,6 @@ sway-applications/
 |    └── P-1 Directories & files
 ├── Project-2
 |    └── P-2 Directories & files
-├── CONTRIBUTING.md
 ├── LICENSE
 └── README.md
 ```
--- a/escrow/src/main.sw
+++ b/escrow/src/main.sw
@@ -1,6 +1,5 @@
 contract;
 
-// TODO: commented-out code is there because the SDK does not support Vec yet so an array is used
 // Our library dependencies
 dep data_structures;
 dep errors;


Braqzen (approved):
Lgtm. Thanks for helping out!

There is a minor note to keep in mind.

It's usually a good practice to create 1 pull request per issue, rather than multiple issues into 1 pull request, because there may be a time when some work needs to be reverted (undone) and it's easier to do this when each piece of work is self contained. Moreover, it's easier to categorize and search for issues by labels and it's easier to review 1 piece of work at a time.

On the other hand, if the issues are related then it's reasonable to group them together into a single pull request although if there is a lot of work it may be beneficial to split them up into separate pull requests.

In this scenario two lines have been changed, one line in each file, so reviewing was not a problem however if this contained many files with many changes throughout each file then it becomes more difficult to review.

PR #280, opened 2022-10-31
remove tilde from method calls for compatibility with the latest sway compiler

sezna:
## Type of change

- Other (describe below)

## Changes

- Remove `~` from all method applications to be compatible with the latest version of Sway.

## Notes

Once https://github.com/FuelLabs/sway/pull/3218 goes in and `0.29.0` is released, code that contains `~`, like `~Address::from()`, will no longer compile. As a show of good faith, respect for our consumers, and in apology for breaking your repo, I'm submitting this PR. ❤️


Braqzen:
The repository has been bumped to version `0.31.1` via #298. The `~` should produce errors when used therefore under that assumption this PR can be closed as CI would have failed if `~` was used anywhere.

PR #282, opened 2022-11-02
Update forc to correct version in README shield

bitzoic:
## Type of change

## Changes

- The shield would state the incorrect version of forc compared to what we have at the end of the README


Braqzen (approved):
I've been using this as a check to see when a new release has come out but that may be misleading to others so this is probably better.

PR #284, opened 2022-11-05
refactor: build script

### LINKED ISSUE #283: Change ci to put formatting in the build script
### Motivation

The oracle node pr needs the ci to be changed in order to test the oracle node.  This includes changing how the ci checks formatting,

matt-user:
## Type of change

- Other (describe below)

## Changes

- Refactored the build script to allow the oracle node to be properly tested

## Notes

- Note 1

## Related Issues

Closes #283

### DIFF
--- a/.github/scripts/forc-handler.sh
+++ b/.github/scripts/forc-handler.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+PROJECT=$1
+COMMAND=$2
+FLAGS=$3
+
+if [ $PROJECT = 'airdrop/airdrop-distributor' ]; then
+    forc $COMMAND --path $PROJECT $FLAGS
+    forc $COMMAND --path $PROJECT/../simple-asset/ $FLAGS
+elif [ $PROJECT = 'airdrop/simple-asset' ]; then
+    forc COMMAND --path $PROJECT FLAGS
+elif [ $PROJECT = 'auctions/english-auction' ]; then
+    forc $COMMAND --path $PROJECT $FLAGS
+    forc $COMMAND --path $PROJECT/tests/artifacts/asset $FLAGS
+    forc $COMMAND --path $PROJECT/../../NFT/ $FLAGS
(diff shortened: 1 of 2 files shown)


.github/scripts/forc-handler.sh
```diff
@@ -0,0 +1,36 @@
...
+    forc $COMMAND --path $PROJECT/../simple-asset/ $FLAGS
+elif [ $PROJECT = 'airdrop/simple-asset' ]; then
+    forc $COMMAND --path $PROJECT $FLAGS
+elif [ $PROJECT = 'auctions/english-auction' ]; then
+    forc $COMMAND --path $PROJECT $FLAGS
+    forc $COMMAND --path $PROJECT/tests/artifacts/asset $FLAGS
+    forc $COMMAND --path $PROJECT/../../NFT/ $FLAGS
+elif [ $PROJECT = 'dao-voting' ]; then
+    forc $COMMAND --path $PROJECT $FLAGS
+    forc  $COMMAND --path $PROJECT/tests/artifacts/gov_token $FLAGS
```
Braqzen, line 18:
```suggestion
    forc $COMMAND --path $PROJECT/tests/artifacts/gov_token $FLAGS
```

Braqzen (commented):
Could you verify that `$FLAGS` takes multiple flags and not just one? I'm unsure if that will take multiple flags. I believe it will only take 1 but I may be incorrect. If it only takes 1 then either

1. Rename to `$FLAG` to indicate a single flag
2. Change to allow any number of flags to be tagged on at the end

.github/scripts/forc-handler.sh
matt-user: I was able to get the script to work locally with multiple `FLAGS` by setting `FLAGS="${@:3}"` but it didn't seem to work in github and after roughly an hour of debugging it still doesn't and I don't know why.  Because of that and seeing as no command in the ci requires multiple flags I'm just gonna make it work with a single flag.

Braqzen (approved):

PR #288, opened 2022-11-08
AMM: Atomic deposit and add liquidity script

### LINKED ISSUE #287: AMM: Atomic deposit and add liquidity script
### Motivation

Currently, adding liquidity to an exchange contract requires three steps:
1. Deposit some amount of asset A using `deposit()`
2. Deposit some amount of asset B using `deposit()`
3. Add liquidity using `add_liquidity()`

A Sway script can do this atomically.

#### Specification

The script should accept asset IDs and the amounts to be deposited to add liquidity by using up at least one of the assets (this depends on the ratio of the values of the two assets). Should refund any extra amount.

supiket:
## Type of change

- Improvement 

## Changes

- A script that atomically deposits and adds liquidity to an exchange contract has been added.

## Notes

- The tests depend on the [PR for script main arguments](https://github.com/FuelLabs/fuels-rs/pull/620).

## Related Issues

Closes #287

### DIFF
--- a/APPLICATION_PROGRESS.md
+++ b/APPLICATION_PROGRESS.md
@@ -79,7 +79,6 @@ The information in this section is split into subsections in order to conceptual
 
 - <h3>Rust ✅</h3>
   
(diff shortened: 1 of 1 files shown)


supiket:
The work for this functionality is now in #304.

PR #295, opened 2022-11-15
Removed std prelude imports

### LINKED ISSUE #269: Remove `std` prelude imports from all apps
### Motivation

The applications have imports from the `std`.
The `std` has a prelude which is code that is automatically imported for us.
This means that we do not have to explicitly import that code anymore.
Any code that is automatically imported for us should be removed from the imports in the apps.

To accomplish this:

1. Check the version of `forc` specified [here](https://github.com/FuelLabs/sway-applications#running-a-project)
2. Navigate to [this link](https://fuellabs.github.io/sway/v0.26.0/index.html) and in the URL replace the version with the one we use in the point above
3. Navigate to section `1.5 Standard Library` and look at the prelude at the bottom
4. Go through each `.sw` file for every app in the repo and if any app imports a module specified in the prelude then remove that import

brandonsurh:
## Type of change

## Changes

- Removed redundant `std` prelude imports from all apps

## Notes

- `std` import groupings that were left with one import were reformatted
- Example: 
```bash
- use std::{block::height, contract_id::ContractId, identity::Identity};
+ use std::block::height;
```

## Related Issues

Closes #269


Braqzen (changes requested):
Hey, it looks like this is close to being merged in but we need to make CI pass so that it allows us to merge. If you take a look at the screenshot that I've attached you'll see there is an X on the left of some apps and `Details` to the right side. 

We need to make sure that there are no X's so if you click on the `Details` it will tell you what the issue is. I've glanced at the multisig and it looks like it's a formatting issue.

Could you do the following:

1. Run `forc fmt` on each application that has an X and check if CI has passed
2. If it looks like the formatting tool is not working as expected (i.e. CI is not passing after the command is run) then you could manually make the changes

brandonsurh:
After running `forc fmt` on each failed application, here are the results:

NFT:
```
cd NFT/
forc fmt
This file: "/home/<myUser>/sway-applications/NFT/src/main.sw" is not part of the build
Error parsing file: Unable to parse: invalid character
```

Multisig-Wallet:
```
forc fmt
This file: "/home/<myUser>/sway-applications/multisig-wallet/src/main.sw" is not part of the build
Error parsing file: Unable to parse: invalid character
invalid character
invalid character
```

Oracle:
```
cd oracle/packages/contract/
forc fmt
This file: "/home/<myUser>/sway-applications/oracle/packages/contract/src/main.sw" is not part of the build
Error parsing file: Unable to parse: invalid character
invalid character

```


I looked at each `.sw` file individually for the applications that did not pass, but did not see any formatting issues or invalid characters. It seems to me that the formatting tool is not working as expected, but would like to hear your input on this before taking any action.

brandonsurh:
I realize I think I was running the latest version of forc and not 0.26.0. That could be the reason for the errors.

Braqzen:
> I realize I think I was running the latest version of forc and not 0.26.0. That could be the reason for the errors.

The versions need to match and if the formatter has a problem then the changes need to be manually implemented to make CI happy

brandonsurh:
Installed the correct version of `forc` (0.26.0) and `forc fmt` shows no errors for any of the failed applications. To make the changes manually how should someone go about doing that?

Braqzen:
> Installed the correct version of `forc` (0.26.0) and `forc fmt` shows no errors for any of the failed applications. To make the changes manually how should someone go about doing that?

Manually means that you do it yourself. You look at the error in CI and apply the appropriate fix.

brandonsurh:
The fundraiser application still spits out an error on forc fmt. Need to check this out later possibly.

brandonsurh:
All builds seem to be successful and tests pass. Let me know if there needs to be any more changes made.

Braqzen (approved):
The changes look good to me, I hope the other apps and other files in these apps did not require such changes

brandonsurh:
> The changes look good to me, I hope the other apps and other files in these apps did not require such changes

I made sure to search through every file in `sway-applications` for an instance of `use std::` so there shouldn't be any more changes to be done.  Thank you for checking out out my changes!

PR #312, opened 2022-12-02
AMM: Swap exact input script

### LINKED ISSUE #229: AMM: Swap exact input and swap exact output scripts

supiket:
## Type of change

- New feature

## Changes

- Swap exact input script has been added

## Notes

- Tests will follow

## Related Issues

Closes #229

### DIFF
--- a/AMM/project/scripts/swap-exact-input/Forc.toml
+++ b/AMM/project/scripts/swap-exact-input/Forc.toml
@@ -1,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "swap-exact-input"
+
+[dependencies]
+libraries = { path = "../../libraries" }
--- a/AMM/project/scripts/swap-exact-input/src/main.sw
+++ b/AMM/project/scripts/swap-exact-input/src/main.sw
@@ -1,0 +1,37 @@
+script;
+
+use core::num::*;
+use libraries::{AMM, Exchange};
+use std::{constants::ZERO_B256, logging::log};
+
+pub enum SwapError {
+    PairExchangeNotRegistered: (ContractId, ContractId),
+}
+
+// It is assumed that the path validity is checked while running this script such that consecutive assets have a valid exchange contract.
+fn main(
+    amm_contract_address: ContractId,
+    assets: Vec<ContractId>,
+    input_amount: u64,
+) -> u64 {
+    let amm_contract = abi(AMM, amm_contract_address.into());
+    let (mut i, mut output_amount, path_length) = (0, Option::None::<u64>(), assets.len);
+    while i < path_length - 1 {
+        let asset_pair = (assets.get(i).unwrap(), assets.get(i + 1).unwrap());
+        let exchange_contract_id = amm_contract.pool {
+            gas: 100_000,
+            coins: 0,
+            asset_id: ZERO_B256,
+        }(asset_pair);
+        require(exchange_contract_id.is_some(), SwapError::PairExchangeNotRegistered(asset_pair));
+        let exchange_contract = abi(Exchange, exchange_contract_id.unwrap().into());
+        let minimum_output: Option<u64> = Option::None;
+        output_amount = Option::Some(exchange_contract.swap_exact_input {
+            gas: 10_000_000,
+            coins: input_amount,
+            asset_id: asset_pair.0.into(),
+        }(minimum_output, 1000));
+        i += 1;
+    }
+    output_amount.unwrap()
+}
--- a/AMM/Forc.toml
+++ b/AMM/Forc.toml
@@ -2,4 +2,5 @@ members = [
   "./project/contracts/AMM-contract",
   "./project/contracts/exchange-contract",
   "./project/contracts/exchange-contract/tests/artifacts/malicious-implementation",
+  "./project/scripts/swap-exact-input",
 ]


supiket:
@Braqzen you can take a look at the test-utils :)

AMM/SPECIFICATION.md
```diff
@@ -135,41 +133,63 @@ This sub-section details what a user is able to do, e.g., click a button and "x,
...
     2. Requires the asset identifier to return the balance for
 
-##### `pool_info()`
+#### `pool_info()`
 
 1. Returns the pool info, i.e., the identifiers and amounts of assets and the liquidity pool asset amount 
     1. If the asset pair of the pool is set
 
-## Sequence Diagram
+## Atomic Add Liquidity Script
```
Braqzen, line 171: How about a `### Script` heading and then nesting the scripts in there? You could exclude the `Script` post-fix and have `Atomic Add Liquidity` etc. or better yet the name of the script itself to follow the convention of the functions above.

AMM/project/contracts/exchange-contract/src/main.sw
```diff
@@ -77,8 +77,8 @@ impl Exchange for Contract {
         let mut added_liquidity = 0;
 
         // checking this because this will either result in a math error or adding no liquidity at all
-        require(asset_a_in_deposit != 0, TransactionError::DepositCannotBeZero);
-        require(asset_b_in_deposit != 0, TransactionError::DepositCannotBeZero);
+        require(asset_a_in_deposit != 0, TransactionError::ExpectedNonZeroDeposit);
```
Braqzen, line 80: I'm wondering if we should distinguish between the assets so that the error will tell the user which asset is 0.

AMM/project/scripts/atomic-add-liquidity/src/main.sw
```diff
@@ -0,0 +1,44 @@
...
+    desired_liquidity: u64,
+) -> u64 {
+    require(desired_liquidity > 0, InputError::DesiredLiquidityZero);
+
+    let exchange_contract = abi(Exchange, exchange_contract_id.into());
+
+    // deposit first asset
+    if deposit_amounts.0 != 0 {
+        exchange_contract.deposit {
+            gas: 100_000,
```
Braqzen, line 23: Why should I trust this gas value (and others)? The SDK has a way to check the gas; however, that might change in the future. Can we avoid hardcoding gas at all?

AMM/project/scripts/atomic-add-liquidity/src/main.sw
```diff
@@ -0,0 +1,44 @@
...
+    // deposit second asset
+    if deposit_amounts.1 != 0 {
+        exchange_contract.deposit {
+            gas: 100_000,
+            coins: deposit_amounts.1,
+            asset_id: assets.1.into(),
+        }();
+    }
+
+    let deadline = height() + 5;
```
Braqzen, line 38: Where does this deadline come from? Why 5?

Braqzen (commented):
I haven't validated the logic itself - likewise, tests. It looks good otherwise. Would approve soon.

AMM/project/scripts/atomic-add-liquidity/src/main.sw
simonr0204: Ideally this should be a script parameter.

AMM/project/contracts/exchange-contract/src/main.sw
```diff
@@ -228,19 +228,17 @@ impl Exchange for Contract {
...
-        require(exact_input > 0, InputError::AmountCannotBeZero);
+        require(exact_input > 0, InputError::ExpectedNonZeroAmount);
 
         let sender = msg_sender().unwrap();
         let input_asset_in_reserve = storage.reserves.get(input_asset);
         let output_asset_in_reserve = storage.reserves.get(output_asset);
 
         let bought = minimum_output_given_exact_input(exact_input, input_asset_in_reserve, output_asset_in_reserve, LIQUIDITY_MINER_FEE);
 
-        require(bought <= output_asset_in_reserve, TransactionError::InsufficientLiquidity);
```
matt-user, line 242: Why was this require removed?

AMM/project/scripts/atomic-add-liquidity/src/main.sw
```diff
@@ -0,0 +1,44 @@
...
+    DesiredLiquidityZero: (),
+}
+
+fn main(
+    exchange_contract_id: ContractId,
+    assets: (ContractId, ContractId),
+    deposit_amounts: (u64, u64),
+    desired_liquidity: u64,
+) -> u64 {
+    require(desired_liquidity > 0, InputError::DesiredLiquidityZero);
```
matt-user, line 16: should we also add requires to ensure the deposit amounts are non-zero?  This would also enable us to remove the if statements checking if the deposit amounts are 0

AMM/project/scripts/swap-exact-input/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+}
+
+enum SwapError {
+    PairExchangeNotRegistered: (ContractId, ContractId),
+}
+
+fn main(
+    amm_contract_address: ContractId,
+    assets: Vec<ContractId>,
+    amounts: Vec<u64>,
```
matt-user, line 18: IMO it would be cleaner to take in a single `Vec<Amounts>` where `Amounts` (or whatever you want to name it) is defined as

```
struct Amounts {
    id: ContractId,
    amount: u64,
}
```

AMM/project/scripts/swap-exact-input/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+) -> u64 {
+    require(assets.len() >= 2, InputError::RouteTooShort);
+    require(amounts.len() == assets.len(), InputError::SwapAmountsNotExact);
```
matt-user, line 21: Then we could remove this check if you consider the above comment

AMM/project/scripts/swap-exact-input/src/main.sw
```diff
@@ -0,0 +1,58 @@
...
+
+    let amm_contract = abi(AMM, amm_contract_address.into());
+
+    let mut latest_bought = amounts.get(1).unwrap();
```
matt-user, line 25: would this work if there are more than 2 amounts? specifically the `.get(1)`

AMM/project/scripts/swap-exact-output/src/main.sw
```diff
@@ -0,0 +1,59 @@
...
+    SwapAmountsNotExact: (),
+}
+
+enum SwapError {
+    PairExchangeNotRegistered: (ContractId, ContractId),
+}
+
+fn main(
+    amm_contract_address: ContractId,
+    assets: Vec<ContractId>,
```
matt-user, line 17: consider what I mentioned above about a single vec parameter

AMM/project/scripts/swap-exact-output/src/main.sw
```diff
@@ -0,0 +1,59 @@
...
+    amounts: Vec<u64>,
+) -> u64 {
+    require(assets.len() >= 2, InputError::RouteTooShort);
+    require(amounts.len() == assets.len(), InputError::SwapAmountsNotExact);
```
matt-user, line 21: again see above

AMM/project/scripts/swap-exact-output/src/main.sw
```diff
@@ -0,0 +1,59 @@
...
+
+        require(exchange_contract_id.is_some(), SwapError::PairExchangeNotRegistered(asset_pair));
+
+        let exchange_contract = abi(Exchange, exchange_contract_id.unwrap().into());
+
+        latest_sold = Option::Some(amounts.get(bought_asset_index - 1).unwrap());
+        let maximum_sell_amount = amounts.get(bought_asset_index).unwrap();
+
+        // swap by specifying the exact amount to buy
+        let _sold = exchange_contract.swap_exact_output {
```
matt-user, line 49:
```suggestion
        exchange_contract.swap_exact_output {
```

AMM/project/scripts/swap-exact-input/src/main.sw
simonr0204, line 18: Is there a way to make this function just take a single input amount? The output of one swap should always be exactly the input of the next, otherwise the user will end up with small amounts of intermediate tokens (if the swap price is better than expected), or failing swaps (if the price swap price is worse than expected).

You could compare the balance of the sender before and after each swap, and then use the delta as the input amount for the next swap. 

e.g. For swap A -> B -> C -> D etc.

User provides input amount of token A, `a`. Script gets the pre-swap balance of token B,  `b`. Swap A->B occurs. Script gets new balance `b'`, and sets input amount for the B->C swap to `(b' - b)`. Script gets pre-swap balance of token C, `c`. Swap B->C occurs. Script gets new balance `c'`, and sets input amount for the C->D swap to `(c' - c)`. .. .etc.

AMM/project/scripts/swap-exact-input/src/main.sw
simonr0204: If my proposed change (function takes a single `input_amount` parameter, and intermediate input amounts are set to the `latest_bought` from the previous swap) then this can become:

`let mut latest_bought = input_amount;`

AMM/project/scripts/swap-exact-output/src/main.sw
```diff
@@ -0,0 +1,59 @@
...
+enum InputError {
+    RouteTooShort: (),
+    SwapAmountsNotExact: (),
+}
+
+enum SwapError {
+    PairExchangeNotRegistered: (ContractId, ContractId),
+}
+
+fn main(
```
simonr0204, line 15: This is trickier to do than the "exact input" case, as here you have no choice but to calculate all the swaps ahead of time. This can be done with helper functions on the AMM contract (see `get_amounts_in` on the [uniswap router contract](https://etherscan.io/address/0xf164fc0ec4e93095b804a4795bbe1e041497b92a#code), or basically what you already have in your test code)

This would enable the user to enter just the amount they want out at the end of the swap, as the script would calculate all the intermediate amounts.

Same as for the "exact input" case, the main function should take a deadline parameter, and a `maximum_amount_in` parameter to protect against excessive total slippage.

simonr0204 (changes requested):

AMM/.gitignore
```diff
@@ -2,3 +2,6 @@ target
 project/contracts/AMM-contract/out
 project/contracts/exchange-contract/out
 project/contracts/exchange-contract/tests/artifacts/malicious-implementation/out
+project/scripts/atomic-add-liquidity/out
```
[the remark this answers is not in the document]
supiket: Sure.

AMM/SPECIFICATION.md
supiket: Done.

AMM/project/contracts/exchange-contract/src/main.sw
supiket: Done.

AMM/project/scripts/atomic-add-liquidity/src/main.sw
supiket: There is an issue with the estimate gas function but I applied your suggestion to the best I could. We can have further discussion on this.

AMM/project/scripts/atomic-add-liquidity/src/main.sw
supiket: Changed it to a script parameter. FYI: none of the contracts or scripts have something like `+ 5` now, but the test files have.

AMM/project/contracts/exchange-contract/src/main.sw
supiket: `bought` is calculated based on `output_asset_in_reserve`, so that `require` is never violated. This is something I caught thanks to the error messages in revert tests :)

AMM/project/scripts/atomic-add-liquidity/src/main.sw
supiket: Decided to go with checking that deposit amounts are not 0 on the frontend side and removed the if statements checking whether they are 0.

AMM/project/scripts/swap-exact-input/src/main.sw
supiket: Removed `amounts` altogether.

AMM/project/scripts/atomic-add-liquidity/src/main.sw
matt-user: yeah that seems fine.  I would just double check to make sure someone sending a tx with deposit amounts of 0 does not break any of the contract logic.

Braqzen: yeah... we need to defensively code against everything. for an MVP it's fine to make assumptions; however, this will likely be used as a template therefore we cannot make such assumptions as someone might take the contract and it gets exploited if the logic is not carefully validated to not cause any issues

AMM/SPECIFICATION.md
```diff
@@ -135,41 +134,65 @@ This sub-section details what a user is able to do, e.g., click a button and "x,
...
+
+1. Swaps assets along a route by specifying exact output for each swap
+    1. If the route has at least 2 assets
+    2. If the exact output amount is specified for each swap
+    3. If the AMM has a pool for each subsequent asset pair in route
+    4. If [`swap_exact_output`](#swap_exact_output) conditions are met
+
+# Sequence Diagram
 
 ![AMM Sequence Diagram](.docs/amm-sequence-diagram.png)
```
Braqzen, line 198: Shouldn't the diagram be altered to show that you can construct the exchange and then deposit, the deposit is used to create the pool and then you can keep depositing / do other things. 
The way this reads to me is that you can construct and somehow during the construction you add a pool, after you add a pool you can do things that do not link to that pool (no lines) and hence there is the assumption of knowledge that you can do those things.

Braqzen (dismissed):
1) The test structure for the scripts is inconsistent. Previously the repo followed a `harness` monolithic structure which was separated into files.
There is a newer structure in a PR that needs to be reviewed, that aside, these scripts mix the setup and testing while the `atomic` script introduces another new structure which separates into cases.
I'm uncertain about that structure as a general rule; however, I think it's probably okay for most cases, including this PR.
I would like to see the `harness` tests be separated into the current repo structure or into the new success/revert cases for the sake of consistency.

2) I'm not a fan of the `test-utils/src/abi.rs` naming convention. I like the name but it is a keyword in sway so we cannot have a library called `abi.sw` and thus we have `interface.sw`. I would prefer to be able to use the abi keyword as a file name. For consistency I think `abi.rs` should be renamed to `interface.rs`.

There are like 6 files for tests across the AMM/Exchange that I did not review because this review has already taken a few hours. Can take a look in the next session.

AMM/project/scripts/swap-exact-input/src/main.sw
```diff
@@ -0,0 +1,60 @@
...
+    RouteTooShort: (),
+}
+
+enum SwapError {
+    ExcessiveSlippage: u64,
+    PairExchangeNotRegistered: (ContractId, ContractId),
+}
+
+fn main(
+    amm_contract_address: ContractId,
```
simonr0204, line 15: Could this be a config time constant? The user should never need to point at different AMM addresses (ironically this did happen on ethereum (naming no names), thus forever fragmenting liquidity. But I hope that's an exception rather than a rule)

AMM/project/scripts/swap-exact-input/src/main.sw
```diff
@@ -0,0 +1,60 @@
...
+    input_amount: u64,
+    minimum_output_amount: Option<u64>,
+    deadline: u64,
+) -> u64 {
+    require(assets.len() >= 2, InputError::RouteTooShort);
+
+    let amm_contract = abi(AMM, amm_contract_address.into());
+
+    let mut latest_bought = input_amount;
+    let intermediate_minimum_amount: Option<u64> = Option::None;
```
simonr0204, line 26: This variable is never changed, as we only check the final amount out. Suggest removing and just passing `Option::None` directly to the swap call.

AMM/project/scripts/swap-exact-input/tests/harness.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    let transaction_parameters =
+        transaction_inputs_outputs(&wallet, &provider, &contracts, &asset_ids, None).await;
+
+    let deadline = provider.latest_block_height().await.unwrap() + 10;
+
+    (wallet, amm, asset_ids, transaction_parameters, deadline)
+}
+
+mod success {
```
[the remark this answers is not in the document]
supiket: Added boundary tests.

AMM/project/scripts/swap-exact-output/tests/harness.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    let transaction_parameters =
+        transaction_inputs_outputs(&wallet, &provider, &contracts, &asset_ids, None).await;
+
+    let deadline = provider.latest_block_height().await.unwrap() + 10;
+
+    (wallet, amm, asset_ids, transaction_parameters, deadline)
+}
+
+mod success {
```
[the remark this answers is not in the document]
supiket: Added boundary tests for the input script. This script / `swap_exact_output` function has a math issue I am trying to resolve, will add other tests later.

AMM/project/scripts/swap-exact-input/src/main.sw
supiket: Done! Hardcoded for now since as far as I know we cannot set config-time constants via the SDK. But the AMM should not be updated too often that this becomes a chore until they can be set programatically.

AMM/project/scripts/swap-exact-input/src/main.sw
supiket: Right.

simonr0204 (dismissed):

simonr0204 (approved):

simonr0204:
Any extra thoughts here @Braqzen or can we click the big green button?

Braqzen:
> Any extra thoughts here @Braqzen or can we click the big green button?

Sorry, I was planning on reviewing this tomorrow. Will get around to this soon

AMM/project/contracts/exchange-contract/src/events.sw
```diff
@@ -1,53 +1,45 @@
...
+    /// Identifier of the other asset
+    asset_b_id: ContractId,
 }
 
 pub struct DepositEvent {
-    /// Deposited amount of the asset that may be withdrawn of used to add liquidity
-    amount: u64,
-    /// Deposited asset that is either asset A or asset B
-    asset: ContractId,
+    /// Deposited asset that may be withdrawn of used to add liquidity
```
Braqzen, line 20: of used?

AMM/project/libraries/src/data_structures.sw
```diff
@@ -1,39 +1,118 @@
...
+    }
+}
+
+pub struct AssetPair {
+    a: Asset,
+    b: Asset,
+}
+
+impl AssetPair {
+    pub fn of(a: Asset, b: Asset) -> Self {
```
Braqzen, line 22: `of` -> `new`?

AMM/project/contracts/exchange-contract/src/utils.sw
```diff
@@ -54,22 +49,18 @@ pub fn minimum_output_given_exact_input(
     result_wrapped.unwrap()
 }
 
-pub fn multiply_div(a: u64, b: u64, c: u64) -> u64 {
+pub fn multiply_divide(a: u64, b: u64, c: u64) -> u64 {
```
Braqzen, line 52: The name of this function doesn't really convey much as to what it is trying to accomplish. I would rename and potentially add some documentation to outline what is the multiplication and division achieving.

AMM/project/contracts/exchange-contract/src/main.sw
```diff
@@ -52,382 +53,346 @@ storage {
...
     }
 
     #[storage(read, write)]
-    fn constructor(pair: (ContractId, ContractId)) {
-        require(storage.pair.is_none(), InitError::CannotReinitialize);
-        require(pair.0 != pair.1, InitError::PoolAssetsCannotBeIdentical);
-
-        storage.pair = Option::Some(pair);
-        log(DefineAssetPairEvent { pair });
+    fn constructor(asset_a_id: ContractId, asset_b_id: ContractId) {
```
Braqzen, line 145: I think the `_id` in the name of the parameters is redundant

AMM/project/scripts/atomic-add-liquidity/tests/cases/revert.rs
```diff
@@ -0,0 +1,134 @@
...
+            },
+        )
+        .call()
+        .await
+        .unwrap();
+}
+
+#[tokio::test]
+#[should_panic(expected = "Revert(18446744073709486080)")]
+// the contract call in the script fails with "DesiredAmountTooHigh" but that message is not propagated
```
Braqzen, line 39: Noting that we need to wait for a new PR / new release for this to be propagated. Could use a branch until then via `staging`

AMM/project/scripts/swap-exact-input/Forc.toml
```diff
@@ -0,0 +1,11 @@
...
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "swap-exact-input"
+
+[dependencies]
+libraries = { path = "../../libraries" }
+
+[constants]
+AMM_ID = { type = "b256", value = "0x7c1420f4d30178fdff677e7b718ab20fdf4c0dbca78d3b9602ee8cabe50349cb" }
```
Braqzen, line 11: Constants above dependencies?

AMM/project/scripts/swap-exact-input/tests/harness.rs
```diff
@@ -0,0 +1,301 @@
+use fuels::prelude::*;
```
Braqzen, line 1: The following 2 scripts contain a `harness.rs` which has the setup and test cases. Splitting into `revert.rs` and `success.rs`, similar to the atomic script, should be fine

Braqzen (commented):
I have not carefully reviewed the tests to see if the logic covers all cases or if they make sense - via this review.

AMM/project/contracts/exchange-contract/src/utils.sw
supiket: Cannot come up with a better name now, will rename by the time this is merged to master.

PR #322, opened 2022-12-11
Fundraiser code quality and testing

### LINKED ISSUE #219: Finish testing the Fundraiser
### Motivation

Recently the projects have been bumped to a newer SDK release which supports [block manipulation](https://fuellabs.github.io/fuels-rs/v0.23.0/testing/chains.html). There are a few tests that need to be completed.

Look in `harness.rs` to see if there is a note mentioning which tests must be updated otherwise look through all the functions.

### LINKED ISSUE #318: Fundraiser code quality
### Motivation

The contract is large and it would benefit from splitting up the structure for maintainability.

We can split the contract into 2 `abi`s, similarly adapt the testing directory.
We can also bump the SDK while we are at it.

Braqzen:
## Type of change

## Changes

- Separated `data_structures.sw` into directory of libraries
- Split contract `abi` into 2 abis: `core` and `info` (getters) for maintenance
- Split `tests/functions` into `core` and `info`, similar changes to `utils`
- Bumped SDK to `0.33.0` and changed Panics / Reverts to use the Sway error enums instead of generic messages
- Added code to check that Sway events are logged properly
- Any tests that were previously marked with `#[ignore]` are now tested
- Changed `deadline` in tests to be slightly more flexible
  - Instead of hardcoding a number we use the current block height plus an arbitrary number based on the testing setup

## Related Issues

Closes #318 
Closes #219


fundraiser/project/fundraiser-contract/src/data_structures/asset_info.sw
```diff
@@ -0,0 +1,13 @@
+library asset_info;
+
+/// Used to track the total amount pledged to an asset
+pub struct AssetInfo {
+    /// The amount that is currently pledged
+    amount: u64,
```
simonr0204, line 6: If a struct only has one field, is the struct abstraction still useful? Could it just be a `u64` ?

Braqzen: It could be a `u64` though standardizing into a structure also has some meaning. I do not recall right now if additional information is meant to be included in the struct in future updates since I have narrowed the scope to these improvements rather than feature updates.

fundraiser/project/fundraiser-contract/tests/functions/core/cancel_campaign.rs
```diff
@@ -135,7 +177,8 @@ mod revert {
     #[should_panic(expected = "CampaignEnded")]
     async fn when_calling_after_deadline() {
```
matt-user, line 178: For this test case I would utilize the `produce_blocks` [function](https://fuellabs.github.io/fuels-rs/v0.33.0/testing/chains.html) provided by the provider along with some variable `offset`.  Then we have
```
let deadline = block_height + offset;
create_campaign(...);
provider.produce_blocks(offset);
cancel_campaign(...);
```

fundraiser/project/fundraiser-contract/src/interface.sw
```diff
@@ -1,59 +1,16 @@
...
-    ///
-    /// * `id` - Unique campaign identifier which is a number from the storage.total_campaigns range
-    ///
-    /// # Reverts
-    ///
-    /// * When the `id` is either 0 or greater than the total number of campaigns created
-    #[storage(read)]
-    fn campaign_info(id: u64) -> CampaignInfo;
-
     /// Marks a campaign as cancelled preventing further pledges or a claim to be made
```
matt-user, line 14: I'm not a huge fan of passing deadline (a block height) here.  Imo it seems annoying to a user (or frontend) to first have to query the block-height and then add it to a desired offset.  It seems better to take in an offset here and then add the block height internally in `CampaignInfo::new()`

fundraiser/project/fundraiser-contract/src/data_structures/campaign_info.sw
```diff
@@ -0,0 +1,43 @@
...
+    /// The current amount pledged used to measure against the target_amount
+    total_pledge: u64,
+}
+
+impl CampaignInfo {
+    pub fn new(
+        asset: ContractId,
+        author: Identity,
+        beneficiary: Identity,
+        deadline: u64,
```
matt-user, line 30: See my other comment, but take in offset here and then store is as a deadline by adding the current block height in in the body of the function.

fundraiser/project/fundraiser-contract/src/data_structures/campaign_info.sw
```diff
@@ -0,0 +1,43 @@
...
+        target_amount: u64,
+    ) -> Self {
+        Self {
+            asset,
+            author,
+            beneficiary,
+            state: State::Funding,
+            deadline,
```
matt-user, line 38:
```suggestion
            deadline: block_height() + offset,
```

fundraiser/project/fundraiser-contract/tests/functions/core/create_campaign.rs
```diff
@@ -63,24 +74,22 @@ mod success {
...
             &author.contract,
             &defaults.asset_id,
             &defaults.beneficiary,
             defaults.deadline,
             defaults.target_amount,
         )
         .await;
-
-        create_campaign(
+        let resposne2 = create_campaign(
```
matt-user, line 92:
```suggestion
        let response2 = create_campaign(
```
nit

fundraiser/project/fundraiser-contract/tests/functions/core/pledge.rs
```diff
@@ -377,11 +463,11 @@ mod revert {
     }
 
     #[tokio::test]
-    #[ignore]
     #[should_panic(expected = "CampaignEnded")]
     async fn when_pledging_after_deadline() {
         let (author, user, asset, _, defaults) = setup().await;
-        let deadline = 5;
+        let provider = author.wallet.get_provider().unwrap();
+        let deadline = provider.latest_block_height().await.unwrap() + 3;
```
matt-user, line 470: see my other comment, but I would consider using the `produce_blocks` function for this test case

matt-user (changes requested):

fundraiser/project/fundraiser-contract/src/interface.sw
matt-user: uhh `create_campaign` that is not `cancel_campaign` I must've put this comment in the wrong spot

fundraiser/project/fundraiser-contract/src/data_structures/pledge.sw
```diff
@@ -0,0 +1,15 @@
+library pledge;
+
+/// Used to track the amount pledged by a user to a specific campaign
+pub struct Pledge {
+    /// The amount pledged to a campaign
+    amount: u64,
+    /// The unique identifier for the campaign
+    id: u64,
```
supiket, line 8:
```suggestion
    campaign_id: u64,
```

fundraiser/project/fundraiser-contract/tests/utils/interface/core.rs
```diff
@@ -0,0 +1,65 @@
...
+pub async fn create_campaign(
+    contract: &Fundraiser,
+    asset: &ContractId,
+    beneficiary: &Identity,
+    deadline: u64,
+    target_amount: u64,
+) -> FuelCallResponse<()> {
+    contract
+        .methods()
+        .create_campaign(asset.clone(), beneficiary.clone(), deadline, target_amount)
```
supiket, line 32:
```suggestion
        .create_campaign(*asset, *beneficiary, deadline, target_amount)
```

supiket (commented):
Looks great, though didn't look into the tests with a magnifying glass.

Some additional comments:
- `interface.sw`
  - `create_campaign` says that it reverts when `asset` is the BASE_ASSET, is that still so? There is no test for that.
  - `pledge` is missing that it reverts when asset amount is 0
 

These comments are not for changed files but rather further changes:
- Specification typos
  - `pledge_count()` items 1 and 2 have "pledgd", suggesting to change them to "pledged"
  - `user_campaign_count()`  item 2 has a "." in the middle of  the sentences, suggesting to remove that
- Errors `TargetReached` and `ContractNotInitialized` are not used anywhere

fundraiser/project/fundraiser-contract/tests/utils/interface/core.rs
Braqzen: My usage requires cloning. 
At least, the `asset` can be deref'd because it's a `ContractId` and it compiles. The `Identity` does not implement `Copy` so it does not compile. Given the usage I think it makes sense to leave them as clones.

fundraiser/project/fundraiser-contract/src/interface.sw
```diff
@@ -125,48 +82,84 @@ abi Fundraiser {
...
+    /// # Arguments
     ///
-    /// * When the `id` is either 0 or greater than the total number of campaigns created
-    /// * When the user attempts to unpledge after the deadline and `target_amount` have been reached
-    /// * When the user has not pledged to the campaign represented by the `id`
-    #[storage(read, write)]
-    fn unpledge(id: u64, amount: u64);
+    /// * `id` - Unique identifier which is a number starting from 1...storage.user_campaign_count
+    #[storage(read)]
+    fn campaign(campaign_history_index: u64, user: Identity) -> Option<Campaign>;
```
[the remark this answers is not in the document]
Braqzen: No, it's convoluted. 

We need to track two things here:

1) Total number of campaigns from all users (global `CampaignInfo` data)
2) The number of unique campaigns per user
2.1) When a new campaign is created by a user the unique counter is incremented by 1

We use the unique counter and their identity to retrieve the global campaign identifier.
This allows us to query for campaigns in two ways:

1) By the actual campaign ID directly via `CampaignInfo`
2) We can check how many campaigns a user has created and use those numbers to then query the global store to get their specific campaign

If we don't do this then we need some alternative way to allow anyone to query the global data while somehow limiting the number of queries that they must make. If we only have a global store then we must iterate over the entire data. With this additional step we can ask how many campaigns a user has created and only query those limited campaigns. 

From the perspective of a UI we probably don't need this if we log and can retrieve global logs. This is a problem for contracts because they cannot get this data so they wouldn't be able to query. At least this allows contracts to also retrieve user data; but, it means they may need to loop a lot if the user has a lot of campaigns.

I don't see a good solution here except to scrap the idea completely and simplify the entire app to only contain a global store for a UI with events. 
This is probably the correct approach for us since the work here is past the MVP and has a bunch of functionality in preparation for a fully off-the-shelf application (although not yet complete).

Perhaps a future PR may just nuke most of the functionality and go with the simpler approach. cc @simonr0204

fundraiser/project/fundraiser-contract/src/main.sw
```diff
@@ -42,18 +48,18 @@ storage {
     user_campaign_count: StorageMap<Identity, u64> = StorageMap {},
     /// Campaigns that have been created by a user
     /// Map(Identity => Map(1...user_campaign_count => Campaign)
-    campaign_history: StorageMap<(Identity, u64), Campaign> = StorageMap {},
+    campaign_history: StorageMap<(Identity, u64), Option<Campaign>> = StorageMap {},
```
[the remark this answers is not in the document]
Braqzen: Ref [comment](https://github.com/FuelLabs/sway-applications/pull/322#discussion_r1071430929). They would need to know the ID in the first place and/or iterate over a global store of all IDs

fundraiser/project/fundraiser-contract/src/interface.sw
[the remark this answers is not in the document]
matt-user: ah gotcha thanks for the the explanation

matt-user (approved):
lgtm

PR #332, opened 2022-12-21
Basic devops scripts

### LINKED ISSUE #331: Basic devops scripts
### Motivation

In preparation for automation and various milestones we'd like to be able to build and test all projects.

This issue is meant to introduce a couple basic scripts

- `build.sh` which uses `forc` to build each application according to the workspaces that have been recently introduced
- `test.sh` which uses `cargo` to run the test suite in each app

The scripts will likely be completely repurposed into Rust binaries in the near future.

Braqzen:
## Type of change

- New feature

## Changes

- Added `build.sh` which uses `forc` to build each application in the repo. Reports APPs that error'd at the end for UX
- Added `test.sh` which uses `cargo` to test each application in the repo. Reports APPs that error'd at the end for UX

## Related Issues

Closes #331


supiket:
Noting for people on macOS and getting the error `./build.sh: line 7: realpath: command not found`, `brew install`ing `coreutils` allowed `realpath` to be recognized.

Question: Do the scripts work with whatever versions of the toolchain are installed/default in the environment they are running on? Is there a way to specify the versions for these scripts, of `forc` and `fuel-core` specifically?

My comments for 
- The build script (not related to the script itself)
  - There are some warnings in some apps that I didn't see before and can and should be fixed, I will do that shortly.
  - In some apps (e.g. auction and AMM), storage modification after external contract interaction is necessary. Should we have a Sway annotation to disable warnings for those? We could document why doing that is necessary and why the warnings are disabled.
  - `AMM/project/contracts/exchange-contract/tests/artifacts/malicious-implementation` uses the ABI of the correct implementation but doesn't read or write to the storage since it is not intended to be used. For this reason, almost every function in that contract has the warning `This function's storage attributes declaration does not match its actual storage access pattern: 'read, write' attribute(s) can be removed.`. But since the ABI has those attributes for storage, removing them from the malicious implementation is not possible. There are some options:
      1. This can remain as-is.
      2. Similar to my question above, we could request an annotation to disable these warnings since this is the intended usage.
      3. The malicious implementation be modified to read and write to the storage where necessary to avoid these warnings. The functions will not be called anyway, so doing this would only increase the bytecode size a bit.

- The test script
   - Works and looks great.

Braqzen:
> Noting for people on macOS and getting the error `./build.sh: line 7: realpath: command not found`, `brew install`ing `coreutils` allowed `realpath` to be recognized.
> 
> Question: Do the scripts work with whatever versions of the toolchain are installed/default in the environment they are running on? Is there a way to specify the versions for these scripts, of `forc` and `fuel-core` specifically?
> 
> My comments for
> 
> * The build script (not related to the script itself)
>   
>   * There are some warnings in some apps that I didn't see before and can and should be fixed, I will do that shortly.
>   * In some apps (e.g. auction and AMM), storage modification after external contract interaction is necessary. Should we have a Sway annotation to disable warnings for those? We could document why doing that is necessary and why the warnings are disabled.
>   * `AMM/project/contracts/exchange-contract/tests/artifacts/malicious-implementation` uses the ABI of the correct implementation but doesn't read or write to the storage since it is not intended to be used. For this reason, almost every function in that contract has the warning `This function's storage attributes declaration does not match its actual storage access pattern: 'read, write' attribute(s) can be removed.`. But since the ABI has those attributes for storage, removing them from the malicious implementation is not possible. There are some options:
>     
>     1. This can remain as-is.
>     2. Similar to my question above, we could request an annotation to disable these warnings since this is the intended usage.
>     3. The malicious implementation be modified to read and write to the storage where necessary to avoid these warnings. The functions will not be called anyway, so doing this would only increase the bytecode size a bit.
> * The test script
>   
>   * Works and looks great.

1. ugh, MAC. I have no further comments here. Feel free to suggest something otherwise I have no intention of doing anything about this.
2. The binaries used are the ones that are currently under the default toolchain. You cannot pin a version, yet; however, Fuelup is currently investigating [this feature](https://github.com/FuelLabs/fuelup/issues/320).
3. New annotations: there have been conversations here and there; however, that is not currently a priority afaik.
4. I don't see the point of increasing the size and adding more code (which has to be maintained) for the sake of not seeing some _warnings_, and only at times. Once these become errors we can modify accordingly.

supiket:
> 1. ugh, MAC. I have no further comments here. Feel free to suggest something otherwise I have no intention of doing anything about this.
> 2. The binaries used are the ones that are currently under the default toolchain. You cannot pin a version, yet; however, Fuelup is currently investigating [this feature](https://github.com/FuelLabs/fuelup/issues/320).
> 3. New annotations: there have been conversations here and there; however, that is not currently a priority afaik.
> 4. I don't see the point of increasing the size and adding more code (which has to be maintained) for the sake of not seeing some _warnings_, and only at times. Once these become errors we can modify accordingly.

1. No suggestions, these scripts will probably be used in non-mac environments anyway. I get your "ugh" but there are many external developers that are / will be using macOS, so it's good to try things we develop on macOS as well. It's not going to disappear anytime soon.
2. Amazing.
3. Yes, I have brought similar things up several times, wanted to open to discussion once more :D Though I am aware it is not a priority.
4. I am not 100% comfortable with warnings generated by intended implementations. Simply reading from storage and writing would do the trick, therefore I do not agree with the maintaining part, but I do agree with the size part.

Braqzen:
> > 1. ugh, MAC. I have no further comments here. Feel free to suggest something otherwise I have no intention of doing anything about this.
> > 2. The binaries used are the ones that are currently under the default toolchain. You cannot pin a version, yet; however, Fuelup is currently investigating [this feature](https://github.com/FuelLabs/fuelup/issues/320).
> > 3. New annotations: there have been conversations here and there; however, that is not currently a priority afaik.
> > 4. I don't see the point of increasing the size and adding more code (which has to be maintained) for the sake of not seeing some _warnings_, and only at times. Once these become errors we can modify accordingly.
> 
> 1. No suggestions, these scripts will probably be used in non-mac environments anyway. I get your "ugh" but there are many external developers that are / will be using macOS, so it's good to try things we develop on macOS as well. It's not going to disappear anytime soon.
> 2. Amazing.
> 3. Yes, I have brought similar things up several times, wanted to open to discussion once more :D Though I am aware it is not a priority.
> 4. I am not 100% comfortable with warnings generated by intended implementations. Simply reading from storage and writing would do the trick, therefore I do not agree with the maintaining part, but I do agree with the size part.

A future upgrade to using Rust binaries will resolve the platform dependency here.

supiket:
> A future upgrade to using Rust binaries will resolve the platform dependency here.

Why future?
*cough* time isn't real *cough*
It is for practical reasons though, so my question persists.

Braqzen:
> 

I have ideas; but, I'd like to have a discussion with the team before I jump in and do too much work in a direction which we may not want to take. People are currently away so sometime after the holidays we can schedule a meeting to clarify some ideas.

supiket (approved):
Could not come up with a cleaner version of `$(dirname $(dirname $(realpath $0)))`. Do we need the `This can probably be cleaner` comments in these scripts?

Braqzen:
> Could not come up with a cleaner version of `$(dirname $(dirname $(realpath $0)))`. Do we need the `This can probably be cleaner` comments in these scripts?

I think so, it points something out that isn't really important but can draw attention for someone to improve

Braqzen:
Intentionally not merging until further reviews.

matt-user (approved):
Is there any way to DRY the `APPS` variable between the two files.  It's only two locations so I'm not super concerned, but it's still two locations instead of one we need to remember to change when adding a new app.

K1-R1 (approved):
Overall it works great, watching all the testing pass is particularly satisfying. Just a few comments on above discussions:

1. I'm on a mac and didn't get the error, it seems that `realpath` is included in macOS 13?
2. I agree with @supiket regarding the `This can probably be cleaner` comment.
3. Regarding @matt-user 's comment on DRY for `APPS`; a `config.sh` file in which to put `APPS`, in combination with the `source` command to 'import' `APPS` into the other files could work nicely, as future iterations that deploy to testnet etc. would require configuration values as well.

Only a nit; It might be worth considering outputting errors to a log file as well as the terminal.

PR #377, opened 2023-01-23
Pin `forc` and `fuel-core` versions in `fuel-toolchain.toml`

### LINKED ISSUE #376: Pin `forc` and `fuel-core` versions in `fuel-toolchain.toml`
### Motivation

We can now have separate versions for applications using the `fuel-toolchain.toml` files: https://fuellabs.github.io/fuelup/v0.16.1/overrides.html

This will require some changes to the CI.

For instance, we can specify the toolchain as the `latest`
```
[toolchain]
channel = "latest"
```
and override the versions:
```
[components]
forc = "0.33.1"
fuel-core = "0.15.3"
```

supiket:
## Type of change

## Changes

- Added `fuel-toolchain.toml` files to application roots
- Replaced `forc`  and `fuel-core` versions in CI with `latest 2023-01-18`
- Removed `forc` badge from repo root README 
- Removed the note stating all apps use the same versions from repo root README
- Added `forc`, `fuel-core` and `fuels` badges to apps' READMEs
- nit: Removed unused `.gitignore` in `games/`

## Related Issues

Closes #376


.template/fuel-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+[toolchain]
+channel = "latest-2023-01-17"
```
bingcicle, line 2:
```suggestion
channel = "latest-2023-01-18"
```

This should be [18th instead](https://github.com/FuelLabs/fuelup/tree/gh-pages/channels/latest). There should probably be a feature to easily list the dates you can declare in fuelup though

bingcicle (changes requested):
This should work properly now on `v0.16.2` - try it with `fuelup self update` - although you might need to run that command twice to re-link the binaries found in `~/.fuelup/bin`

.template/fuel-toolchain.toml
supiket: Nice catch, missed that one. I strongly agree with the necessity of such a feature

bingcicle: https://github.com/FuelLabs/fuelup/pull/375

PR up for such a feature

Braqzen (commented):
1. Why does the oracle use a different channel?
2. The README has a badge which states which version we are on. What should be done about this now? Remove the badge?
3. The README mentions that all projects are under the same [CI](https://github.com/FuelLabs/sway-applications/blob/master/README.md#repository-structure). That's no longer going to be the case. This should be updated.

supiket:
> 1. Why does the oracle use a different channel?

Outdated as you were drafting this message I'm guessing.

> 2. The README has a badge which states which version we are on. What should be done about this now? Remove the badge?

Moved the badges to apps' READMEs.

> 3. The README mentions that all projects are under the same [CI](https://github.com/FuelLabs/sway-applications/blob/master/README.md#repository-structure). That's no longer going to be the case. This should be updated.

Any updates to a project still must make sure that the other projects continue to pass CI though.

oracle/fuel-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+[toolchain]
+channel = "beta-1"
```
Braqzen, line 2:
```suggestion
channel = "latest-2023-01-18"
```

I don't know about that one chief

supiket: Ughh I was testing locally, thanks for catching this.

Braqzen (approved):

.github/workflows/ci.yml
```diff
@@ -62,8 +60,7 @@ jobs:
       - name: Install Fuel toolchain
         uses: FuelLabs/action-fuel-toolchain@v0.2.0
         with:
-          name: my-toolchain
-          components: forc@${{ env.FORC_VERSION }}, fuel-core@${{ env.CORE_VERSION }}
+          toolchain: latest
```
bingcicle, line 63: You may also use this like so, provided you upgrade to **v0.3.0** ([link](https://github.com/FuelLabs/action-fuel-toolchain#pinned-toolchains)):

```suggestion
          toolchain: latest
          date: 2023-01-18
```

bingcicle (dismissed):
LGTM on the usage of the overrides side! Just added an optional note regarding the `action-fuel-toolchain` usage.

.github/workflows/ci.yml
Braqzen: This isn't ideal. I was hoping for a decoupling of the apps. This means that if an app needs to be bumped then all the others need to be bumped as well.

Braqzen (commented):
The CI matrix is long. Could you split that array up onto a few lines to prevent scrolling to the right?

Braqzen (approved):

supiket:
> This isn't ideal. I was hoping for a decoupling of the apps. This means that if an app needs to be bumped then all the others need to be bumped as well.

It does not mean that. Bumping an app would be a process like this:

- Bump the `fuel-toolchain.toml` of _the_ app
  1. Changing the pinned versions and the pinned toolchain date, or
  2. Changing the pinned versions but not the pinned toolchain date
- If there are problems in the CI, bump the pinned toolchain date in the CI

PR #385, opened 2023-01-25
Add owner functionality in the `Multisig`

### LINKED ISSUE #383: Add owner functionality in the `Multisig`
### Motivation

Some features that would be nice to have in the application are

1. Change the weight of a user
  1.1. This is equivalent to adding a new user, removing an existing user or changing the weight of a user to a different non-zero value
2. Approval weight getter

Braqzen:
## Type of change

- New feature

## Changes

- `approval_weight()` getter returns the approval weight of the user
- Any non-zero `weight` requirements have been removed
- `constructor()`
  - Removed unnecessary tests for `AddressCannotBeZero` & `WeightingCannotBeZero`
- Split `data_structures` into directory of libraries since the file was starting to become large and cluttered
- `set_threshold()` changes the threshold required for execution of transactions
- `set_weight()` changes the approval weights of users which are summed and compared to the threshold in order to check if enough votes have approved the Tx
- `threshold_hash()` used to hash the arguments in the `set_threshold()` function in order to recover correctly
- `weight_hash()` used to hash the arguments in the `set_weights()` function in order to recover correctly

## Notes

- `set_weight()` takes a single user instead of a vector of users because the SDK does not properly support vectors

## Related Issues

Closes #383


K1-R1 (commented):
WRT `Vec` support and weights; I still think that for now taking a single `User` as a parameter in `weight_hash ` and `set_weights ` and then immediately placing that `User` in a `Vec` is the best option. This way requires minimal changes when SDK support arrives, and as I understand it, should allow testing. I don't think we should be pushing untested functionality.

Braqzen:
> WRT `Vec` support and weights; I still think that for now taking a single `User` as a parameter in `weight_hash ` and `set_weights ` and then immediately placing that `User` in a `Vec` is the best option. This way requires minimal changes when SDK support arrives, and as I understand it, should allow testing.

The issue isn't the abi, it's that `Vec` is used at all. If I change it to a single user and wrap it in a vec inside the functions then the issue will persist.

K1-R1:
> The issue isn't the abi, it's that `Vec` is used at all. If I change it to a single user and wrap it in a vec inside the functions then the issue will persist.

Ah I think I see, so the issue is that the `Weight` data structure contains `users: Vec<User>`? If so then `Weight`, `hash_weight` and `set_weights` could all be refactored to work on a single `User`, which would allow testing? I still think this is important functionality, even for a single user at a time, and personally would prefer only tested functionality. Until SDK support arrives.

Out of curiosity would it make any difference if the `Weight` type was exposed in the ABI?

Braqzen:
> > The issue isn't the abi, it's that `Vec` is used at all. If I change it to a single user and wrap it in a vec inside the functions then the issue will persist.
> 
> Ah I think I see, so the issue is that the `Weight` data structure contains `users: Vec<User>`? If so then `Weight`, `hash_weight` and `set_weights` could all be refactored to work on a single `User`, which would allow testing? I still think this is important functionality, even for a single user at a time, and personally would prefer only tested functionality. Until SDK support arrives.
> 
> Out of curiosity would it make any difference if the `Weight` type was exposed in the ABI?

I could change it so that only a single user can have their weight changed at a time. No tuple wrapping needed although it's inconsistent with other functionality. Doesn't really matter as long as it is documented. 

As for the weight being exposed, no, that would not make a difference because it contains a vector. Once the vec is removed everything will be okay.

Braqzen:
Will PR again once a single weight is set at a time.

K1-R1 (approved):
Looks great, LGTM

PR #399, opened 2023-02-04
OTC swap frontend

### LINKED ISSUE #400: OTC swap frontend
### Motivation

A frontend for creating and executing offers using the OTC swap predicate

simonr0204:
## Type of change
- New feature: frontend

## Changes

- Adding a basic frontend for the OTC swap predicate
- The frontend uses a precompile of the predicate bytecode, and substitutes in the offer details in the appropriate places. 

### Things that are working:
- Generating bytecode root from offer details
- Displaying any coins found at the corresponding predicate address
- Automatically detecting network change and updating coins
- Input validation

### Known limitations [See also, `TODO` comments in project]
- A readme. Will add via subsequent PR before merging to master. 
- Spending the predicate's coins, either through order fulfillment or cancelling by order creator. (blocked by SDK)

## To test: 

- From OTC-swap-predicate root, `cd frontend && npm install && npm start`
- To make an offer: Enter `amount`, `asset ID` and `receiver`. Addresses can be hex or bech32 fuel addresses. A predicate address will be displayed. You can copy this and send funds to it using the fuel wallet. Invalid inputs will cause no predicate address to be displayed, and will be highlighted.
- To take an offer. Enter `amount`, `asset ID` and `receiver` of an existing offer. On the beta-2 testnet, offer (`0`, `0x00...`, `0x00...`) is funded for convenience.

## Related Issues
Closes #400


OTC-swap-predicate/frontend/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
Braqzen, line 23:
```suggestion
yarn-error.log*

```

OTC-swap-predicate/frontend/src/hooks/useFuel.ts
```diff
@@ -0,0 +1,26 @@
...
+      setLoading(false);
+    }, 500);
+    return () => clearTimeout(timeout);
+  }, []);
+
+  return [fuel as NonNullable<Window["fuel"]>, error, isLoading] as const;
+}
+
+
+export default useFuel;
```
Braqzen, line 26:
```suggestion
export default useFuel;

```

OTC-swap-predicate/frontend/src/hooks/useNetwork.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        fuel?.on("network", handleNetworkChange);
+    
+        return () => {
+            fuel?.off("network", handleNetworkChange);
+        };
+    }, [fuel]);
+  }
+
+export default useNetwork;
```
Braqzen, line 44:
```suggestion
export default useNetwork;

```

Braqzen (commented):
1. I recommend creating a new PR which takes master and merging that into staging. Next, adapt this code structure to the rest of the apps
2. There are a few too many lines in various places, could remove some
3. The `css` uses `px` and `%` in some places. Afaik that's not great, there's various ways of setting width / height but I don't remember what the go to is. I think using `rem` or `em` is meant to be the standard.
4. There's at least one console log (for changing network). I think we should avoid logging things to the console unless there's a good reason to do it. Arbitrary debug-esque calls reduce the code quality imo. Perhaps use `toast` instead?
5. Unsure about that blob of bytecode. It says there is some manipulation there too. That begs some questions. I think it's better to generate that and automatically sub values in. Seems more credible and trustworthy that way because I can see how things are created.
6. Target for compilation is `es5` as opposed to the proposed `es6` for other apps
7. `npm` should work but I suggest using `pnpm` and `vite`
8. I have yet to run the PR

matt-user (commented):
Just a couple of comments.  In general files should follow their respective component names `ComponentNameExample.tsx`.  Also if we encounter an error or have a successful contract interaction we should alert the user.  We are using `toast` for this.  You can import it from the `fuel-ui` library.

simonr0204:
> Just a couple of comments. In general files should follow their respective component names `ComponentNameExample.tsx`. Also if we encounter an error or have a successful contract interaction we should alert the user. We are using `toast` for this. You can import it from the `fuel-ui` library.

Currently there are no contract interactions or transactions in the app. It simply calculates the predicate ID to send the funds to.

simonr0204:
Thanks for the review, @Braqzen 

> 1. I recommend creating a new PR which takes master and merging that into staging. Next, adapt this code structure to the rest of the apps
> 2. There are a few too many lines in various places, could remove some
> 3. The `css` uses `px` and `%` in some places. Afaik that's not great, there's various ways of setting width / height but I don't remember what the go to is. I think using `rem` or `em` is meant to be the standard.
> 4. There's at least one console log (for changing network). I think we should avoid logging things to the console unless there's a good reason to do it. Arbitrary debug-esque calls reduce the code quality imo. Perhaps use `toast` instead?
> 5. Unsure about that blob of bytecode. It says there is some manipulation there too. That begs some questions. I think it's better to generate that and automatically sub values in. Seems more credible and trustworthy that way because I can see how things are created.
> 6. Target for compilation is `es5` as opposed to the proposed `es6` for other apps
> 7. `npm` should work but I suggest using `pnpm` and `vite`
> 8. I have yet to run the PR

1. I'll take care of conflicts with master when we merge staging into master. 
2. Empty lines?
3. You're probably right here, but if this functionally works then I'd like to merge and potentially tidy up later.
4. Noted - will remove `console.log` s
5. I wanted to avoid re-compiling every time a constant changed in order for the predicate calculation to be absolutely instantaneous and dynamic. There's also no nice way of setting compile-time constants just before compilation. I think if the user wants to check the predicate root they can always build the predicate themselves.
6. Hmm. I'll look into this.
7. Will look into migrating in a future PR.

Braqzen:
> 

2. I should have said "new lines". There are multiple places where there are multiple new / empty lines in a row when there should be fewer e.g. inputValidation with the try / catch has a line between them
5. There should be instructions written in the subsequent PR telling the user how to create the bytecode so that they can match on it with some editor to see if it is exactly the same when they reproduce

Braqzen (dismissed):

supiket (approved):
Saw it working, cool

Braqzen:
`console.log()` -> `toast.error()`

PR #459, opened 2023-02-12
Removed StorageMap

### LINKED ISSUE #449: Name Registry: remove `StorageMap` from imports in `main.sw`
### Motivation

`StorageMap` is included in the prelude therefore we do not need to import it.
Remove [this line](https://github.com/FuelLabs/sway-applications/blob/0132a4a89bb0f9c1a5d016db609adead87c43626/name-registry/project/registry-contract/src/main.sw#L23).

Kirito-Excalibur:
## Type of change

- Bug fix

## Changes

- Removed `StorageMap` from imports in `main.sw`

## Related Issues

Closes #449 >

### DIFF
--- a/name-registry/project/registry-contract/src/main.sw
+++ b/name-registry/project/registry-contract/src/main.sw
@@ -20,7 +20,7 @@ use std::{
     call_frames::msg_asset_id,
     context::msg_amount,
     logging::log,
-    storage::StorageMap,
+   
 };
 
 // TODO: Replace the B256 config-time constant with a ContractId when possible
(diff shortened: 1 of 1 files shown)


Braqzen (changes requested):
Thanks for the quick PR!

If you take a look at the CI below it will say that it is failing because the formatting is not what it is expected to be. If you remove the line where the `StorageMap` used to be entirely then that should fix the issue!

Kirito-Excalibur:
Okay thanks, will do.

Braqzen (approved):
Sweet, this should pass CI, thanks!

PR #467, opened 2023-02-12
New devops script to bump the fuel-toolchain.toml

### LINKED ISSUE #466: New devops script to bump the fuel-toolchain.toml
### Motivation

It would be nice to have a script similar to `build.sh` and `test.sh` which takes a `fuel-toolchain.toml` that has the next pinned versions that we want to use and go through each app and test whether we can bump to those versions in those apps.

It would

1. move the current toolchain to a different file name in the same directory
2. move the new toolchain configuration into the app
3. try to build with the new toolchain
  i. if it builds successfully then it removes the file from step 1.
  ii. if it failed to build then it moves the file from step 1. back so that we can continue to build from the same version

Braqzen:
## Type of change

- New feature

## Changes

- Script which replaces the toolchain with a newer one
  - if it fails to build then it reverts back to the current toolchain
- reports which apps have failed / succeeded

## Related Issues

Closes #466


supiket:
Just to clarify, do we want the apps to be on the same version all the time given that the CI stopped imposing the same versions? Can you provide more context about the motivation of this script?

Braqzen:
> Just to clarify, do we want the apps to be on the same version all the time given that the CI stopped imposing the same versions? Can you provide more context about the motivation of this script?

We do not need nor should we attempt to keep apps on the same version.
We should aim to keep the apps relatively up to date but that's about it.
This script is the first step of helping us bump the repo at a glance. It allows us to bump everything up to the latest toolchain that we specify and anything that doesn't work will remain on the same toolchain.
This can be expanded in the future to specify individual apps with toolchains rather than running through the entire repo.

nfurfaro:
>It allows us to bump everything up to the latest toolchain that we specify and anything that doesn't work will remain on the same toolchain.

So this would allow some apps to be upgraded to the newer specified toolchain, while keeping other apps on their current toolchain if they fail to upgrade successfully? 
If so, that seems reasonable. Will it work for downgrading as well?

Braqzen:
> > It allows us to bump everything up to the latest toolchain that we specify and anything that doesn't work will remain on the same toolchain.
> 
> So this would allow some apps to be upgraded to the newer specified toolchain, while keeping other apps on their current toolchain if they fail to upgrade successfully? If so, that seems reasonable. Will it work for downgrading as well?

I have not attempted to downgrade but in theory it should work.
What it does is it renames the current toolchain in case we need to revert and then it copies the one from `.devops` in place of the current one. It will build and if it builds successfully then it will remove the old toolchain. If it fails then it will move back the old (actually current) toolchain to where it was so it's as if nothing has happened.

This means that if something can be upgraded without errors then it will do so otherwise it will revert back to whatever the toolchain was.
This script provides no guarantees for anything other than `forc build` so the tests may no longer be compatible.

nfurfaro:
>This script provides no guarantees for anything other than forc build so the tests may no longer be compatible.

Hmm. I wonder about the value of automating the toolchain upgrade if we then have to manually revert it if the tests no longer pass? Could this be extended to also run the tests before deciding whether to keep the upgrade or revert back to the old one?

Braqzen:
> > This script provides no guarantees for anything other than forc build so the tests may no longer be compatible.
> 
> Hmm. I wonder about the value of automating the toolchain upgrade if we then have to manually revert it if the tests no longer pass? Could this be extended to also run the tests before deciding whether to keep the upgrade or revert back to the old one?

We have a file to `build` and another to `test`. 
This one is to `bump` for building. 

I don't think we can get away from needing to fix things and only bump if everything works seamlessly. I could add a check to also run the tests and potentially add it as a flag such that we can bump things automatically if both the fuel and cargo tools produce no errors, at the same time I think this is fine imo. I don't mind updating to do that but I kind of expect tests to go out of sync if the forc version is bumped enough times. With the speed of development this script might be pretty useless if that restriction is the only way it functions.

Lmk what you think.

nfurfaro (approved):
uTack

PR #538, opened 2023-02-20
made preludes import explicit as stated in issue #434

### LINKED ISSUE #434: TicTacToe: change `fuels::prelude::*` to be explicit
### Motivation

The application has `tests` which use the `*` glob operator to import everything. 
We should aim to explicitly import.

1. Comment out / remove that import
2. Build the tests
3. Check to see which imports are recommended by the compiler
4. Use the suggested imports instead of `fuels::prelude::*;`
  i. If there is a single import then use the following style `fuels::prelude::<name of import>;`
  ii. If there are multiple imports then use the following style `fuels::prelude::{<name of first import>, <name of second import>};`

GeckoSplat:
## Type of change

## Changes

- Change 1 : made preludes import explicit as stated in issue #434

Closes #434


Braqzen (approved):
LGTM!

PR #545, opened 2023-02-20
F-NFT: Split the abi into core and info

### LINKED ISSUE #432: Fractional NFT: change `fuels::prelude::*` to be explicit
### Motivation

The application has `tests` which use the `*` glob operator to import everything. 
We should aim to explicitly import.

1. Comment out / remove that import
2. Build the tests
3. Check to see which imports are recommended by the compiler
4. Use the suggested imports instead of `fuels::prelude::*;`
  i. If there is a single import then use the following style `fuels::prelude::<name of import>;`
  ii. If there are multiple imports then use the following style `fuels::prelude::{<name of first import>, <name of second import>};`

### LINKED ISSUE #535: F-NFT: Split the abi into core and info
### Motivation

Restructure the application into `core` functionality and `info` functionality. 
This means taking the `abi` and creating a new one `Info` with the getters and adjusting the tests to follow the same structure.

May need to adjust specification structure.

Braqzen:
## Type of change

## Changes

- Split into `core` and `info` (getters)
- Restructured tests

## Related Issues

Closes #535
Closes #432


nfurfaro:
I don't love the idea of splitting the `abi` like this; to me it seems like an atomic thing that should be self-contained. 
Testing the abi can be organized as needed (i.e: separating read-only from storage writes, etc...).
That said, maybe the ship has already sailed on this, in which case I guess it's ok.

Braqzen:
> I don't love the idea of splitting the `abi` like this; to me it seems like an atomic thing that should be self-contained. Testing the abi can be organized as needed (i.e: separating read-only from storage writes, etc...). That said, maybe the ship has already sailed on this, in which case I guess it's ok.

There's cases like an NFT extending various extensions: minting, burning, etc., so splitting things isn't bad. It's more about maintenance in general. It sets us up for scaling things by separating concerns. Anyone can easily combine them if they want to.

nfurfaro:
>There's cases like an NFT extending various extensions: minting, burning, etc., so splitting things isn't bad. It's more about maintenance in general. It sets us up for scaling things by separating concerns. Anyone can easily combine them if they want to.

True. I can live with this for now. When there is an agreed-upon NFT standard for Fuel, I would say at that point the NFT abi should match what is in the standard, which means both getters and setters. 
Seeing as we don't have that standard yet I don't need to die on this hill :)

nfurfaro (approved):

PR #556, opened 2023-02-22
Convert `.devops` scripts into a rust project

### LINKED ISSUE #554: Convert `.devops` scripts into a rust project
### Motivation

Scripts are becoming repetitive. Expand them to a similar system that the sway repository has with commands.

The project can work as follows in the first iteration:

1. \<name of project\> --build
2. \<name of project\> --fmt
3. \<name of project\> --test
4. \<name of project\> --bump

#### build

Runs `forc build` on each project and reports which projects failed to build at the end

#### fmt

Runs `forc fmt` and `cargo fmt` on each project. Once `forc fmt` is fixed to return an error code of a non-zero value then we can report which projects failed to build at the end for both `forc` and `cargo`

#### test

Runs `cargo test --color always -q` on each project and reports which projects failed tests at the end

#### bump

Runs the `bump_toolchain.sh` equivalent i.e. tries to upgrade each project to the next toolchain but if it fails then it returns the project back to the previous toolchain, then reports which projects failed to bump at the end

Subsequent additions may allow specifying which apps are bumped instead of all of them

Braqzen:
## Type of change

## Changes

- Added `build` script
- Added `fmt` script
- Added `test` script
- Added `bump` script
- Removed shell scripts because the crate replicates and expands functionality

## Notes

Take a look at the README.md for usage

## Related Issues

Closes #554


.devops/repo/src/commands/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    let apps = read_applications();
+
+    let mut errors: Vec<String> = vec![];
+
+    for app in apps {
+        println!("\nBuilding {}", app);
+
+        set_current_dir(format!("{}/{}/project", root, app));
+
+        let build = Command::new("forc").arg("build").status();
```
prestwich: instead of using `set_current_dir` on the parent process, use 
```
let working_dir = std::fs::canonicalize(format!("{}/{}/project", root, app)?;

let build = Command::new("forc")
    .current_dir(working_dir)
    .arg("build")
    .status();
```

.devops/repo/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "repo"
+version = "0.1.0"
+edition = "2021"
+authors=["Fuel Labs <contact@fuel.sh>"]
```
bingcicle, line 5:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
publish = false
```

Spacing, and since this package probably isn't meant to be published, let's add `publish = false` here.

.devops/repo/README.md
```diff
@@ -0,0 +1,47 @@
+## Instructions
```
bingcicle, line 1: Perhaps there should be a note here that there are several assumptions here, in case someone wants to run this locally? (rust toolchain installed, forc installed, etc.)

.devops/repo/src/commands/bump.rs
```diff
@@ -0,0 +1,70 @@
...
+                format!("{}/{}", project.clone().display(), toolchain.clone()),
+            ])
+            .status();
+
+        let result = Command::new("forc")
+            .current_dir(project.clone())
+            .arg("build")
+            .status();
+
+        match result {
```
bingcicle, line 37: This can also be

```suggestion
        match Command::new("forc")
            .current_dir(project.clone())
            .arg("build")
            .status() {
```

.devops/repo/src/commands/build.rs
```diff
@@ -0,0 +1,21 @@
...
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).unwrap();
+
+        execute(
+            Command::new("forc").current_dir(project).arg("build"),
+            &mut errors,
+            &app,
+        );
+    }
+
+    print_applications(errors, "Errors found in".to_string());
```
bingcicle, line 20: It could be worth creating a const for the error messages that's passed into `print_applications`, i.e. 

`const ERRORS_FOUND_MESSAGE: &str = "Errors found in";`

.devops/repo/src/commands/build.rs
```diff
@@ -0,0 +1,21 @@
...
+use std::process::Command;
+
+pub(crate) fn run(apps: Vec<String>, root: String) {
+    let mut errors: Vec<String> = vec![];
+
+    for app in apps {
+        println!("\nBuilding {}", app);
+
+        // TODO: safety
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).unwrap();
```
bingcicle, line 11: You may use `anyhow` here to quickly throw an error:

```suggestion
        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).map_err(|e| anyhow!("Failed to canonicalize path to project for app '{}': {}", app, e));
```

.devops/repo/src/commands/bump.rs
```diff
@@ -0,0 +1,70 @@
...
+    let mut success: Vec<String> = vec![];
+
+    let toolchain = "fuel-toolchain.toml".to_string();
+    let tmp_toolchain = "fuel-toolchain2.toml".to_string();
+
+    for app in apps {
+        println!("\nBumping {}", app);
+
+        // TODO: safety
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).unwrap();
```
bingcicle, line 14: Same as above

.devops/repo/src/commands/fmt.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub(crate) fn run(apps: Vec<String>, root: String) {
+    let mut sway_errors: Vec<String> = vec![];
+    let mut cargo_errors: Vec<String> = vec![];
+
+    for app in apps {
+        println!("\nFormatting {}", app);
+
+        // TODO: safety
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).unwrap();
```
bingcicle, line 12: Ditto

.devops/repo/src/commands/test.rs
```diff
@@ -0,0 +1,23 @@
...
+use std::process::Command;
+
+pub(crate) fn run(apps: Vec<String>, root: String) {
+    let mut errors: Vec<String> = vec![];
+
+    for app in apps {
+        println!("\nTesting {}", app);
+
+        // TODO: safety
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app)).unwrap();
```
bingcicle, line 11: Ditto

.devops/repo/README.md
Braqzen: Good point. This is meant for internal use so the assumption was that people have what we use but yh, I'll add some things

.devops/repo/src/commands/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+        let project = std::fs::canonicalize(format!("{}/{}/project", root, app))
+            .map_err(|error| {
+                anyhow!(
+                    "Failed to canonicalize path to project for app '{}': {}",
+                    app,
+                    error
+                )
+            })
+            .unwrap();
```
bingcicle, line 21:
```suggestion
```

Ah, to make full use of `anyhow` and avoid `unwrap()`, we probably want to use `?` here and make `run()` return an `anyhow::Result`.

.devops/repo/src/main.rs
```diff
@@ -0,0 +1,40 @@
...
+    Build,
+    /// Bump each project from its current `fuel-toolchain.toml` to the one in this repository
+    Bump,
+    /// Format the Sway and Rust files in each project
+    Fmt,
+    /// Run the Rust tests for each project
+    Test,
+}
+
+fn main() {
```
bingcicle, line 28: That also means we return `anyhow::Result` here:

```suggestion
fn main() -> Result<()> {
```

eureka-cpu (approved):

PR #591, opened 2023-03-12
Multisig: change optional data fields to be mandatory

### LINKED ISSUE #590: Multisig: change optional data fields to be mandatory
### Motivation

In order to recover addresses (and have a consistent UI/UX) we need to sign some data meaning we cannot have that data be optional.
The users will need to sign data even if it's garbage data.

Change `set_weight` and `set_threshold` and anything else that may be needed.

Braqzen:
## Type of change

## Changes

- Changed `set_threshold` and `set_weight` to require data instead of being optional
- Adjusted tests

## Notes

- It was either this or make everything optional to be consistent. I think it makes more sense to require data even if it's garbage although that may not be the best.

## Related Issues

Closes #590

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
         run: rustup component add rustfmt
 
       - name: Install Fuel toolchain
-        uses: FuelLabs/action-fuel-toolchain@v0.4.0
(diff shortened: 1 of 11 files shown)


Braqzen:
Gonna make them all optional instead.

PR #604, opened 2023-03-19
Change the type of price in storage

### LINKED ISSUE #448: Oracle: change the type of `price` in `storage`
### Motivation

The `Option` type is supported by the SDK. 
Change the type to be an option containing a u64 [here](https://github.com/FuelLabs/sway-applications/blob/0132a4a89bb0f9c1a5d016db609adead87c43626/oracle/project/oracle-contract/src/main.sw#L17-L18). 
Adjust contract and tests.

fgimenez:
## Type of change

## Changes

- Change the type of `price` in `storage` to `Option<u64>`
- Update tests

## Notes

- I tried to follow the readme at https://github.com/FuelLabs/sway-applications/blob/master/oracle/README.mdbut could not spin up the the node:
```
$ fuel-core run --chain project/oracle-node/.chainConfig.json
Error: an error occurred while loading the chain config file project/oracle-node/.chainConfig.json

Caused by:
    missing field `consensus` at line 35 column 1
$ fuel-core --version                                        
fuel-core 0.17.3
```
Happy to create an issue and work on a fix if needed.

## Related Issues

Closes #448

### DIFF
--- a/oracle/project/Forc.lock
+++ b/oracle/project/Forc.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = 'core'
-source = 'path+from-root-B0D5F718107CCC86'
+source = 'path+from-root-894BF76E6DA2FFD3'
 
 [[package]]
 name = 'oracle-contract'
@@ -9,5 +9,5 @@ dependencies = ['std']
 
 [[package]]
 name = 'std'
-source = 'git+https://github.com/fuellabs/sway?tag=v0.35.2#f05ecaf2de1ebe3b3da80ecd8eb6053a84145cf4'
+source = 'git+https://github.com/fuellabs/sway?tag=v0.35.3#5d2b10bd83791d2eaff04206dbd45bfdd9cf23ff'
 dependencies = ['core']
--- a/oracle/project/contracts/oracle-contract/src/interface.sw
+++ b/oracle/project/contracts/oracle-contract/src/interface.sw
@@ -8,7 +8,7 @@ abi Oracle {
 
     /// Return price of asset
     #[storage(read)]
-    fn price() -> u64;
+    fn price() -> Option<u64>;
 
     /// Changes the price in storage to the value of `price`
     ///
--- a/oracle/project/contracts/oracle-contract/src/main.sw
+++ b/oracle/project/contracts/oracle-contract/src/main.sw
@@ -14,8 +14,7 @@ use interface::Oracle;
 
 storage {
     // Current price of tracked asset
-    // TODO use option when https://github.com/FuelLabs/fuels-rs/issues/415 is fixed
-    price: u64 = 0,
+    price: Option<u64> = Option::None,
 }
 
 // TODO treat owner as an identity once https://github.com/FuelLabs/sway/issues/2647 is fixed
@@ -25,7 +24,7 @@ impl Oracle for Contract {
     }
 
     #[storage(read)]
-    fn price() -> u64 {
+    fn price() -> Option<u64> {
         storage.price
     }
 
@@ -33,7 +32,7 @@ impl Oracle for Contract {
     fn set_price(price: u64) {
         require(msg_sender().unwrap() == Identity::Address(Address::from(OWNER)), AccessError::NotOwner);
 
-        storage.price = price;
+        storage.price = Option::Some(price);
 
         log(PriceUpdateEvent { price });
     }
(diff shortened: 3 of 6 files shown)


Braqzen (commented):
I would consider looking into using `matches!()` to compare enums but that's not really an issue.

As for your comment about the client not running you could try it with [this config](https://github.com/FuelLabs/sway-applications/blob/master/.template/ui/config/chainConfig.json). Feel free to file an issue!

In terms of CI it's failing because the command `forc update` needs to be run inside `/project`

fgimenez:
hey @Braqzen thanks a lot for your feedback! :heart: 

> I would consider looking into using matches!() to compare enums but that's not really an issue.

Sure done, I get a couple of warnings though:
```
warning: unused variable: `set_price_amount`
  --> contracts/oracle-contract/tests/functions/price.rs:15:38
   |
15 |         assert!(matches!(price, Some(set_price_amount)));
   |                                      ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_set_price_amount`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `set_price_amount`
  --> contracts/oracle-contract/tests/functions/set_price.rs:27:38
   |
27 |         assert!(matches!(price, Some(set_price_amount)));
   |                                      ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_set_price_amount`
```

> As for your comment about the client not running you could try it with [this config](https://github.com/FuelLabs/sway-applications/blob/master/.template/ui/config/chainConfig.json). Feel free to file an issue!

Thanks, just checked and only oracle is using a node, so I'll fix it here if that's ok. I've pushed changes copy the chain config you linked to the oracle project, and also to update the readme and the example env file. With this in place everything seems to work, I still see an error in the node logs after deploying the contract, not sure if this is ok?:
```
2023-03-20T08:53:29.327829Z ERROR new{name=}:initialize_loop{service="P2P"}: fuel_core_p2p::service: 184: Got an error during transaction broadcasting MessageTooLarge
```

> In terms of CI it's failing because the command forc update needs to be run inside /project

hmm I had already pushed an updated Forc.lock, when I execute `forc update` in `oracle/project` the file is not changed, am I missing something?

Braqzen:
> hey @Braqzen thanks a lot for your feedback! heart
> 
> > I would consider looking into using matches!() to compare enums but that's not really an issue.
> 
> Sure done, I get a couple of warnings though:
> 
> ```
> warning: unused variable: `set_price_amount`
>   --> contracts/oracle-contract/tests/functions/price.rs:15:38
>    |
> 15 |         assert!(matches!(price, Some(set_price_amount)));
>    |                                      ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_set_price_amount`
>    |
>    = note: `#[warn(unused_variables)]` on by default
> 
> warning: unused variable: `set_price_amount`
>   --> contracts/oracle-contract/tests/functions/set_price.rs:27:38
>    |
> 27 |         assert!(matches!(price, Some(set_price_amount)));
>    |                                      ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_set_price_amount`
> ```
> 
> > As for your comment about the client not running you could try it with [this config](https://github.com/FuelLabs/sway-applications/blob/master/.template/ui/config/chainConfig.json). Feel free to file an issue!
> 
> Thanks, just checked and only oracle is using a node, so I'll fix it here if that's ok. I've pushed changes copy the chain config you linked to the oracle project, and also to update the readme and the example env file. With this in place everything seems to work, I still see an error in the node logs after deploying the contract, not sure if this is ok?:
> 
> ```
> 2023-03-20T08:53:29.327829Z ERROR new{name=}:initialize_loop{service="P2P"}: fuel_core_p2p::service: 184: Got an error during transaction broadcasting MessageTooLarge
> ```
> 
> > In terms of CI it's failing because the command forc update needs to be run inside /project
> 
> hmm I had already pushed an updated Forc.lock, when I execute `forc update` in `oracle/project` the file is not changed, am I missing something?

To make the warnings disappear you could add an `if guard` on `set_price_amount` but that's a lot more verbose than we need to be now that I think about it.
I think your initial implementation is nicer or alternatively we could unwrap `price` and compare it directly to `set_price_amount`. Unwrapping is not ideal but we don't really care in this case.
My bad.

I've checked out your branch and started the client and deployed without any issues from inside `/project` so I'm unsure with what is going on here. I assume we did different things but since I got it to work then I wouldn't worry about it. If it's an issue in the future then we can fix it then.

As for the `.lock` file if you take a look at the `diff` it states that the file has changed from `.2` to `.3`. I've also run `forc update` after changing versions and it goes back to `.2` so I would change into `/project` and run `forc update` and check the `.lock` file. Its version should change to `.2`. I assume CI will be happy at that point.

PR #669, opened 2023-05-17
Bump DAO

### LINKED ISSUE #649: Bump `DAO`
To:

- `forc 0.42.1`
- `fuel-core 0.18.3`
- `fuels-rs 0.42.0`

K1-R1:
## Type of change

## Changes

- Bump application to `forc 0.42.1` and `fuel-core 0.18.3`
- Bump test harness to `fuels 0.42.0`

## Related Issues

Closes #649


DAO/project/contracts/DAO-contract/tests/utils/setup.rs
```diff
@@ -10,7 +10,7 @@ abigen!(
     ),
     Contract(
         name = "GovToken",
-        abi = "./contracts/DAO-contract/tests/artifacts/gov_token/out/debug/gov_token-abi.json"
+        abi = "./contracts/test-artifacts/gov_token/out/debug/gov_token-abi.json"
```
Braqzen, line 13: Given the structure of the AMM it made sense to extract some code but here it does not. I would move the code back to where it was.

Iirc the token isn't even needed and can be replaced by the functionality from the SDK so I recommend either moving it back or deleting it and generating an asset to deposit via the SDK

K1-R1: This is required due to changes with nested projects and forc workspaces. An error is produced if the test artifacts are kept in their old places.

Braqzen: ugh, so you cannot have a project inside a project. they can only be siblings...
okay, two options, replace with the SDK or raise issues to address this when we have time.
thoughts?

K1-R1: I've replaced with the SDK

DAO/project/contracts/DAO-contract/tests/functions/core/deposit.rs
```diff
@@ -60,71 +53,32 @@ mod success {
...
-        .await;
+        let (gov_token_id, _other_token_id, _deployer, user, asset_amount) = setup().await;
 
         let call_params = CallParameters::new(asset_amount, AssetId::from(*gov_token_id), 100_000);
         deposit(&user.dao_voting, call_params).await;
     }
 
     #[tokio::test]
-    #[should_panic(expected = "IncorrectAssetSent")]
+    #[should_panic]
```
Braqzen, line 67: This looks like a regression

K1-R1: Nice catch, forgot to add back in

DAO/project/contracts/DAO-contract/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "Apache-2.0"
 
 [dependencies]
 fuels = { version = "0.42.0", features = ["fuel-core-lib"] }
+rand = { version = "0.8.5" }
```
Braqzen, line 10: We shouldn't be introducing entropy into the suite unless it's needed.

DAO/project/contracts/DAO-contract/tests/utils/setup.rs
```diff
@@ -32,16 +34,41 @@ pub(crate) fn proposal_transaction(asset_id: ContractId) -> Proposal {
...
+    let num_coins = 1;
+    let coin_amount = 1_000_000;
+
+    let base_asset = AssetConfig {
+        id: BASE_ASSET_ID,
+        num_coins,
+        coin_amount,
+    };
+    let mut gov_token_id = AssetId::zeroed();
+    gov_token_id.try_fill(&mut rng).unwrap();
```
Braqzen, line 47: The randomness can be replaced by creating the `AssetId` using the `::from(array)` syntax

DAO/project/contracts/DAO-contract/tests/utils/setup.rs
```diff
@@ -32,16 +34,41 @@ pub(crate) fn proposal_transaction(asset_id: ContractId) -> Proposal {
...
+    let mut other_token_id = AssetId::zeroed();
+    other_token_id.try_fill(&mut rng).unwrap();
+    let other_token = AssetConfig {
+        id: other_token_id,
+        num_coins,
+        coin_amount,
+    };
+    let assets = vec![base_asset, gov_token, other_token];
+
+    let num_wallets = 2;
```
Braqzen, line 62: I would hoist this to the top where the other vars are declared and rename to not use abbreviations

DAO/project/contracts/DAO-contract/tests/utils/setup.rs
K1-R1: I don't mind making this change, and understand that generally it's not desirable to introduce entropy to the test suite but in this case I'm unsure why the entropy is an issue when creating wallet addresses; the [rust SDK suggests this in the docs](https://rust.fuel.network/v0.44.0/wallets/test-wallets.html#setting-up-a-test-wallet-with-multiple-custom-assets), and the [rust SDK's functions](https://docs.rs/fuels-test-helpers/0.44.0/src/fuels_test_helpers/lib.rs.html#50-55) also use `rand`.

Braqzen: It's unnecessary. Adding entropy when it's not needed is akin to adding complexity when it's not needed. Things may go wrong and it may be difficult to debug. If you can make it static then it should be static.

Braqzen (approved):

PR #703, opened 2023-07-08
Name-Registry: accept multiple assets for payment

### LINKED ISSUE #699: Name-Registry: accept multiple assets for payment
### Motivation

The contract accepts a single asset which may be configured.
Update to accept any asset set via ownership over the registry.

Braqzen:
## Type of change

- New feature

## Changes

- New functions: `set_asset()` to add a new payment method & `rate()` to check the cost
- Adjusted `extend()` & `register()` to allow the user to specify an asset for payment
- SRC-2 docs

## Related Issues

Closes #699


name-registry/project/contracts/registry-contract/src/main.sw
```diff
@@ -152,4 +172,10 @@ impl Info for Contract {
             Option::None => Result::Err(RegistrationValidityError::NameNotRegistered),
         }
     }
+
+    #[storage(read)]
+    fn rate(id: ContractId) -> Option<u64> {
+        // SAFETY TODO: will bump in newer release PR
+        storage.assets.get(id).unwrap()
```
bitzoic, line 179: This will fail if I pass an invalid contract id. There should be a check for `None` before unwrapping.

bitzoic (commented):
I don't believe offering the ability to pay with multiple types of tokens really enhances the application in any meaningful way. Tokens historically have been volatile and the ratio between them can fluctuate rapidly. By offering multiple token payments a user may be over or under paying by using one token over another. This is bad for UX.

Furthermore, I don't see how this is particularly relevant to an example of a domain name resolver application.

Braqzen:
> I don't believe offering the ability to pay with multiple types of tokens really enhances the application in any meaningful way. Tokens historically have been volatile and the ratio between them can fluctuate rapidly. By offering multiple token payments a user may be over or under paying by using one token over another. This is bad for UX.
> 
> Furthermore, I don't see how this is particularly relevant to an example of a domain name resolver application.

You've countered your own point regarding volatility. Whether you have 1 asset or 2 makes no difference. Both are volatile.

Being able to pay for something in multiple ways is inherently an improvement. I don't think that needs an explanation.

Over / under paying isn't an issue. Pick the asset that you want. You can check the price / the app would tell you how much you are paying.

As for the last sentence, it's just an improvement to the scope of the app and not specifically the design of the registry. It's easier for people to pick this up and use it as a starting point, not that that was my reasoning for adding this in.

name-registry/project/contracts/registry-contract/src/main.sw
```diff
@@ -8,21 +8,28 @@ mod interface;
...
+    call_frames::msg_asset_id,
+    constants::ZERO_B256,
+    context::msg_amount,
+};
 
-// Amount of units of the asset to charge per 100 seconds of registration duration for every name/entry
 configurable {
-    ASSET_ID: ContractId = ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000),
-    PRICE_PER_HUNDRED: u64 = 1,
+    OWNER: Identity = Identity::Address(Address::from(ZERO_B256)),
```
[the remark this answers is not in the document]
Braqzen: After looking through the standard I've decided to not implement it in this PR.
It's a large library that requires a bunch of boilerplate code that would also needs to be tested.
The configurable is sufficient for this improvement. In the future we may raise another issue to implement the standard.

Braqzen:
Wait wtf why did the tests pass here, noooo

name-registry/project/contracts/registry-contract/src/main.sw
```diff
@@ -6,37 +6,49 @@ mod events;
...
+    call_frames::msg_asset_id,
+    constants::ZERO_B256,
+    context::msg_amount,
+};
 
-// Amount of units of the asset to charge per 100 seconds of registration duration for every name/entry
 configurable {
-    ASSET_ID: ContractId = ContractId::from(0x0000000000000000000000000000000000000000000000000000000000000000),
-    PRICE_PER_HUNDRED: u64 = 1,
+    /// The priviledged user with the ability to set assets for payment for the registry
```
K1-R1, line 27:
```suggestion
    /// The privileged user with the ability to set assets for payment for the registry
```

K1-R1 (changes requested):
Can you update the versions in the README. Otherwise looks good

Braqzen:
> Can you update the versions in the README. Otherwise looks good

Sorry, could you elaborate on why the versions need to be updated? I have not changed the versions in this PR.

K1-R1:
> > Can you update the versions in the README. Otherwise looks good
> 
> Sorry, could you elaborate on why the versions need to be updated? I have not changed the versions in this PR.

My mistake, apologies

K1-R1 (approved):

PR #715, opened 2023-07-27
Update links for multi-token standard

bitzoic:
## Type of change

- Documentation

## Changes

- Updated links to point to sway multi token standard in the RFCS repo. This should instead point to either the [fuel spec changes](https://github.com/FuelLabs/fuel-specs/pull/491) or the [SRC-20 standard](https://github.com/FuelLabs/sway-standards/issues/1). I have opted for the SRC-20 standard as it is more user facing.

## Notes

- The previous links pointed to an RFC

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,12 @@ sway-applications/
 - [Escrow](./escrow) is a third party that keeps an asset on behalf of multiple parties.
 
 > **Warning**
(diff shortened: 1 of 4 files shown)


Braqzen (approved):
Ok. We should update these to point at official docs instead of issues when things are finalized. It's better to point at a living doc.

PR #716, opened 2023-08-02
Implement low_level_call into the multisig

### LINKED ISSUE #22: Call contracts without knowing their ABI at compile time of the multisig
Once a Tx has been approved past a threshold it should be executable.

In order to accomplish execution we need to be able to call arbitrary contracts without knowing anything about them at the compilation of the multi-signature wallet.
This means that we need to be able to make generic calls at runtime.

### LINKED ISSUE #424: Multisig: interface bullet point style
### Motivation

The [`interface of the .template`](https://github.com/FuelLabs/sway-applications/blob/0132a4a89bb0f9c1a5d016db609adead87c43626/.template/project/template-contract/src/interface.sw#L8) uses a `*` before each parameter and a `:` after it.

Update the interface in the application to follow the same style.

### LINKED ISSUE #717: Multisig: SRC-2
Implement SRC-2 inline docs standard

### LINKED ISSUE #718: Multisig: remove arbitary data from hash functions
Arbitrary bytes in hash functions can be removed; they served to add additional entropy however this is not necessary as knowing the hash does not create a vulnerability, the hash must still be signed over by a multisig user.

K1-R1:
## Type of change

- New feature

## Changes

- `transfer` and `execute_transaction` have been merged into a single function: `execute_transaction`
-  [SRC-2](https://github.com/FuelLabs/sway-standards/tree/master/standards/src_2) has been implemented
-  A single hash function `compute_hash` has replaced the separate hash functions.
-  The use of arbitrary bytes in hash functions has been removed, more detail available in notes below.
- pub(crate)

## Notes

- Arbitrary bytes in hash functions have been removed as they served to add additional entropy; however this is not necessary as knowing the hash does not create a vulnerability, the hash must still be signed over by a multisig user.
- The test `functions::info::compute_hash::gets_call_hash` has been written, but is ignored with the following reason:
> Encoding of nested structs with Bytes fields in the RustSDK doesn't match the encoding in Sway. This is not an issue for the use of compute_hash when making a contract call; as the user would not be creating the hash in the SDK

## Related Issues

Closes #22
Closes #717
Closes #718 

Implementing SRC-2 Closes #424


multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -56,67 +83,125 @@ impl MultiSignatureWallet for Contract {
...
-    ) {
-        require(storage.nonce.read() != 0, InitError::NotInitialized);
+    fn set_threshold(signatures: Vec<SignatureInfo>, threshold: u64) {
+        let nonce = storage.nonce.read();
+        require(nonce != 0, InitError::NotInitialized);
         require(threshold != 0, InitError::ThresholdCannotBeZero);
         require(threshold <= storage.total_weight.read(), InitError::TotalWeightCannotBeLessThanThreshold);
 
-        let transaction_hash = hash_threshold(data, storage.nonce.read(), threshold);
+        let transaction_hash = compute_hash(TypeToHash::Threshold(Threshold {
```
Braqzen, line 171: May as well `impl` a `new(id, nonce, threshold)` for CQ

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -56,67 +83,125 @@ impl MultiSignatureWallet for Contract {
...
-
         let previous_threshold = storage.threshold.read();
+        require(previous_threshold <= approval_count, ExecutionError::InsufficientApprovals);
 
-        storage.nonce.write(storage.nonce.read() + 1);
+        storage.nonce.write(nonce + 1);
         storage.threshold.write(threshold);
 
         log(SetThresholdEvent {
+            nonce,
```
Braqzen, line 185: Irrelevant info, no need to log the nonce.

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -182,24 +243,56 @@ impl Info for Contract {
...
+/// The struct will be a variant of [TypeToHash].
+///
+/// # Arguments
+///
+/// * `type_to_hash` : [TypeToHash] - The struct to hash.
+///
+/// # Returns
+///
+/// * [b256] - The hash.
+fn compute_hash(type_to_hash: TypeToHash) -> b256 {
```
Braqzen, line 261: This should not be in the main file

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -182,24 +243,56 @@ impl Info for Contract {
...
+/// # Additional Information
+///
 /// Takes in a transaction hash and signatures with associated data.
 /// Recovers a b256 address from each signature;
 /// it then increments the number of approvals by that address' approval weighting.
 /// Returns the final approval count.
+///
+/// # Arguments
+///
+/// * `argument_1`: [Identity] - This argument is a user to be hashed.
```
Braqzen, line 280: Arg 1 should be removed again

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
```diff
@@ -1,75 +1,69 @@
...
-    value: u64,
-    data: Bits256,
-    signatures_data: Vec<SignatureInfo>,
-) -> FuelCallResponse<()> {
-    contract
+    contract_call_params: Option<ContractCallParams>,
+    signatures: Vec<SignatureInfo>,
+    target: Identity,
+    transfer_params: TransferParams,
+) -> Result<FuelCallResponse<()>, &'static str> {
```
Braqzen, line 19: Why do we need error handling here?
I imagine there is a test where you check the `Err(string)` defined below appears in a panic msg. It's a string though and we shouldn't be tested for that in contracts so I imagine this is propagated from elsewhere. Either way it fails. 
I haven't seen the usage yet so it may be valid but my immediate impression is to remove since based on this code alone and how we usually test we wouldn't care to handle it

multisig-wallet/project/contracts/multisig-contract/tests/utils/setup.rs
```diff
@@ -3,31 +3,43 @@ use fuels::{
...
 
 const MULTISIG_CONTRACT_BINARY_PATH: &str = "./out/debug/multisig-contract.bin";
 const MULTISIG_CONTRACT_STORAGE_PATH: &str = "./out/debug/multisig-contract-storage_slots.json";
+const CALLABLE_CONTRACT_BINARY_PATH: &str =
+    "../test-artifacts/callable-contract/out/debug/callable-contract.bin";
+const CALLABLE_CONTRACT_STORAGE_PATH: &str =
+    "../test-artifacts/callable-contract/out/debug/callable-contract-storage_slots.json";
 
-pub(crate) const DEFAULT_THRESHOLD: u64 = 5;
+pub(crate) const DEFAULT_CALLDATA_VALUE_PARAM: u64 = 1;
```
Braqzen, line 39: No need for `param` in the name. It doesn't convey much

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
+contract;
+
+use std::context::msg_amount;
+
+abi CallableContract {
```
Braqzen, line 5: I would continue with your convention of `Target` instead of `Callable`.

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+    #[storage(read)]
+    fn check_counter_map(address: Address) -> u64;
+
+    #[storage(read)]
+    fn check_deposit_map(address: Address) -> u64;
+}
+
+storage {
+    counter_map: StorageMap<Address, u64> = StorageMap {},
+    deposit_map: StorageMap<Address, u64> = StorageMap {},
```
Braqzen, line 22: No need to specify the type in the name, poor convention

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl CallableContract for Contract {
+    #[storage(write)]
+    fn change_mapping_without_value(address: Address, value: u64) {
```
Braqzen, line 27:
```suggestion
    fn update_counter(address: Address, value: u64) {
```

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+        storage.counter_map.insert(address, value);
+    }
+
+    #[payable]
+    #[storage(write)]
+    fn change_mapping_with_value(address: Address, value: u64) {
```
Braqzen, line 33:
```suggestion
    fn update_deposit(address: Address, value: u64) {
```

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+        storage.counter_map.insert(address, value);
+        storage.deposit_map.insert(address, msg_amount());
+    }
+
+    #[storage(read)]
+    fn check_counter_map(address: Address) -> u64 {
```
Braqzen, line 39:
```suggestion
    fn count(address: Address) -> u64 {
```

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
```diff
@@ -0,0 +1,47 @@
...
+        storage.counter_map.get(address).try_read().unwrap_or(0)
+    }
+
+    #[storage(read)]
+    fn check_deposit_map(address: Address) -> u64 {
```
Braqzen, line 44:
```suggestion
    fn deposit(address: Address) -> u64 {
```

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -56,67 +83,125 @@ impl MultiSignatureWallet for Contract {
 
     #[storage(read, write)]
     fn execute_transaction(
-        data: b256,
+        contract_call_params: Option<ContractCallParams>,
```
Braqzen, line 86: Yh, these call / transfer params can probably enum'd to and typed to remove optionals. This in addition to `new()` would clean this up

multisig-wallet/project/contracts/multisig-contract/tests/functions/core/set_threshold.rs
```diff
@@ -129,30 +152,25 @@ mod revert {
 
         constructor(&deployer.contract, default_users()).await;
 
-        let data = Bits256([2u8; 32]);
+        let _data = Bits256([2u8; 32]);
```
Braqzen, line 155: Unused, remove

multisig-wallet/project/contracts/multisig-contract/tests/functions/info/compute_hash.rs
```diff
@@ -0,0 +1,180 @@
...
+            &deployer.contract,
+            TypeToHash::Transaction(transaction_instance),
+        )
+        .await
+        .value;
+
+        assert_eq!(Bits256(expected_hash.into()), response);
+    }
+
+    #[ignore = "Encoding of nested structs with Bytes fields in the RustSDK doesn't match the encoding in Sway. This is not an issue for the use of compute_hash when making a contract call; as the user would not be creating the hash in the SDK"]
```
Braqzen, line 98: Wasn't the param commented out and therefore this test should run?

multisig-wallet/project/contracts/multisig-contract/tests/functions/info/compute_hash.rs
```diff
@@ -0,0 +1,180 @@
...
+            nonce,
+            target,
+            transfer_params,
+        };
+
+        let transaction_instance_token = Token::Struct(vec![
+            transaction_instance
+                .contract_call_params
+                .clone()
+                .into_token(), //This causes test to fail: tokenizing Some(ContractCallParams), does not encode the same as Sway
```
Braqzen, line 128: hm, maybe it's the SDK causing issues regardless of the sway param

multisig-wallet/project/contracts/multisig-contract/tests/utils/setup.rs
```diff
@@ -56,6 +68,24 @@ pub(crate) fn default_users() -> Vec<User> {
     vec![fuel_user_1, evm_user_1]
 }
 
+pub(crate) async fn deploy_callable_contract(
```
Braqzen, line 71:
```suggestion
pub(crate) async fn setup_target_contract(
```

Braqzen (commented):
I'll review the setup, execution and spec after some comments have been addressed

multisig-wallet/.docs/sequence-diagram.drawio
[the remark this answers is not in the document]
K1-R1: Drawio allows you to drag and drop a .png file, this is what I did to convert from the old to the new.

multisig-wallet/project/contracts/multisig-contract/src/data_structures/hashing.sw
```diff
@@ -1,41 +1,137 @@
 library;
 
 use ::data_structures::user::User;
+use std::{bytes::Bytes, constants::ZERO_B256};
 
-pub struct Transaction {
+impl Bytes {
```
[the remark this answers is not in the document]
K1-R1: This functionality is needed to hash the structs containing `Bytes` fields. It has not been added to the standard library as we are looking to move away from relying on memory representation for serialization as per the following from an internal discussion: 

> Relying on the memory representation for serialization is a nice hack, but we're actively trying to not support this to replace it with some defined encoding.

However this defined coding has not yet been implemented and the functionality is required for serialisation of Bytes here

multisig-wallet/project/contracts/multisig-contract/src/data_structures/hashing.sw
```diff
@@ -1,41 +1,137 @@
...
+    }
+}
+
+pub trait IntoBytes {
+    /// Converts self into [Bytes].
+    fn into_bytes(self) -> Bytes;
+}
+
+/// Parameters for calling a contract.
+pub struct ContractCallParams {
```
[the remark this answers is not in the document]
K1-R1: > I'm uncertain if the params should be in this file or a separate file

I've included them here as they primarily serve as a field in the data structure Transaction, the purpose of which is to be hashed.

> Also `low_level_call::CallParams` has a similar name which may be confusing

Understandable, but ContractCallParams is about as explicit as possible.

multisig-wallet/project/contracts/multisig-contract/src/errors.sw
```diff
@@ -1,14 +1,27 @@
 library;
 
+/// Error log for when transaction execution panics.
 pub enum ExecutionError {
+    /// Emitted when attempting to call when `target` is not a [Identity::ContractId].
```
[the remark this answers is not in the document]
K1-R1: I've updated this to:

>  Emitted when the called `target` is not a [Identity::ContractId].

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
```diff
@@ -1,149 +1,171 @@
...
 abi MultiSignatureWallet {
     /// The constructor initializes the necessary values and unlocks further functionality.
     ///
     /// # Arguments
     ///
-    /// * `users` - The users of the multisig, who can sign transactions to add their approval.
+    /// * `users`: [Vec<User>] - The users of the multisig, who can sign transactions to add their approval.
     ///
-    /// # Reverts
+    /// # Panics
```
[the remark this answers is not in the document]
K1-R1: I've used `Panics` as per https://github.com/FuelLabs/sway-standards/tree/master/standards/src_2 . This PR will remain compliant with SRC-2, but it would be worth bringing up the issue on the standard itself.

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
```diff
@@ -1,149 +1,171 @@
...
     /// * When the total approval count is less than the required threshold for execution.
+    /// * When attempting to call when `target` is not a [Identity::ContractId].
+    ///
+    /// # Number of Storage Accesses
+    ///
+    /// * Reads: `3`
+    /// * Writes: `2`
     #[storage(read, write)]
-    fn execute_transaction(data: b256, signatures: Vec<SignatureInfo>, to: Identity, value: u64);
+    fn execute_transaction(contract_call_params: Option<ContractCallParams>, signatures: Vec<SignatureInfo>, target: Identity, transfer_params: TransferParams);
```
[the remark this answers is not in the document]
K1-R1: I disagree about `contract_call_params`, it is very explicit; these are the parameters for calling a contract. The same for `transfer_params`. 
The enum suggestion would not work as a transfer does not require `contract_call_params`, but a call _may_ include transfer params.

multisig-wallet/project/contracts/multisig-contract/src/main.sw
K1-R1: Great shout

multisig-wallet/project/contracts/multisig-contract/src/main.sw
K1-R1: The enum suggestion would not work as a transfer does not require `contract_call_params`, but a call _may_ include `transfer params`.

multisig-wallet/project/contracts/multisig-contract/src/main.sw
K1-R1: I've included as it's necessary to know the nonce of the wallet in order to execute transactions, so a user can use this logged response instead of having to call the `nonce` getter when making multiple transactions in sequence.

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -126,40 +211,12 @@ impl MultiSignatureWallet for Contract {
...
-
-        transfer(value, asset_id, to);
-
-        log(TransferEvent {
-            asset: asset_id,
-            nonce: storage.nonce.read() - 1,
-            to,
-            value,
-        });
+        log(SetWeightEvent { nonce, user })
```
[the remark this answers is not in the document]
K1-R1: As above:
I've included as it's necessary to know the nonce of the wallet in order to execute transactions, so a user can use this logged response instead of having to call the nonce getter when making multiple transactions in sequence.

multisig-wallet/project/contracts/multisig-contract/tests/functions/info/compute_hash.rs
K1-R1: Your comment below is correct

multisig-wallet/project/contracts/multisig-contract/tests/functions/info/compute_hash.rs
K1-R1: Yh

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
K1-R1: Its because of the need to use `.set_contract_ids()` on a call, which only accepts `ContractId`s, however `target` is an `Identity`, as such I have to match on `target` and handle the case where `Identity` is not a `ContractId`; I chose to return an error here. The actual panicking test for making a call when `target` isn't a `ContractId` doesn't use this function

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
K1-R1: Although I'd normally agree with your suggestions for these function names; as this is a test artifact having the overly-explicit `change_mapping_without_value` makes it much clearer in the tests whats supposed to be happening when calling these functions.

multisig-wallet/.docs/sequence-diagram.drawio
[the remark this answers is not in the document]
Braqzen: Woah, really. Sugoi

multisig-wallet/project/contracts/multisig-contract/src/data_structures/hashing.sw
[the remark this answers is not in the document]
Braqzen: I see. This definitely should not be at the application level. We don't want to encourage users to play with `asm` blocks. There should be an issue raised to move this to some external place so that we can import it. Perhaps sway libs.

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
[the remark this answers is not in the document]
Braqzen: Is there an issue with having an enum with multiple variants and types? Pretty sure you could use an enum

multisig-wallet/project/contracts/multisig-contract/src/main.sw
Braqzen: The getter is free while this pollutes the state with irrelevant info. The UI can automatically fetch the nonce and update itself to always show the current nonce. This should be removed.

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
Braqzen: My understanding is that this is purely used via our set up so we need to set things up properly or this error would catch that, is that correct?
If it is then it should be removed. We ought to ensure tests follow the correct branch flow.

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
Braqzen: No. There's already an equivalent amount of prior information that you would need to use your verbose name.

`change` or `update`, `update` is canonical.
`mapping`, which one? Is there more than one? What does the mapping do?
`without_value` implies there is one with a value but what is the value? Payable, parameter?

If anything, someone who may not know what the contract does `update_counter` conveys more information than changing some mapping with some value.

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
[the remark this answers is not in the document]
K1-R1: It's a small amount of work to make this change, so far now I'll comply with SRC-2, and _if_ the standard changes I'll update

multisig-wallet/project/contracts/multisig-contract/src/main.sw
K1-R1: done

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
K1-R1: > My understanding is that this is purely used via our set up so we need to set things up properly or this error would catch that, is that correct?

The error would be triggered if `target` was set to an `Identity::Address` in a test when calling `execute_transaction()`, in which case the actual contract method can't be called because it requires the use of `set_contract_ids()`. As such when matching on `target`, I returned an error if it's an `Identity::Address`, as the call cannot be made.

multisig-wallet/project/contracts/test-artifacts/callable-contract/src/main.sw
K1-R1: Fair point, will implement your suggetsions

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
K1-R1: I suppose I could take `target` as a `ContractId` instead of `Identity` for this utility function instead.

K1-R1: An alternative would be to just extract the underlying `Address` from an `Identity:Address` and convert it into a `contract_id`

K1-R1: > An alternative would be to just extract the underlying `Address` from an `Identity:Address` and convert it into a `contract_id`

I've implemented this

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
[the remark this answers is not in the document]
Braqzen: The enum allows you to omit the optional types, no? It also wraps it into a single param.

multisig-wallet/project/contracts/multisig-contract/tests/utils/interface/core.rs
Braqzen: Chotto matte, so we're not supposed to set an `Address` and instead only use `ContractId`, is that correct? If so we shouldn't be hacking the address into an id and instead just `panic!("<error>")`

multisig-wallet/project/contracts/multisig-contract/src/interface.sw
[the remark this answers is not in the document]
K1-R1: The enum would remove the option for `ContractCallParams`, but the `Call` variant's type would then need to have some of the transfer parameters as options, as you may want to call a contract with or without value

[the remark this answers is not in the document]
Braqzen: Reasonable to me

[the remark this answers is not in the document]
K1-R1: Implemented the enum as suggested

K1-R1:
Comments addressed, @Braqzen

multisig-wallet/project/contracts/multisig-contract/tests/utils/setup.rs
```diff
@@ -227,40 +226,40 @@ pub(crate) fn call_parameters(
 ) -> Transaction {
     match with_value {
         false => Transaction {
-            contract_call_params: Some(ContractCallParams {
+            contract_identifier: deployer.contract.contract_id().try_into().unwrap(),
```
Braqzen, line 229: Looking at this function it creates a struct that is almost identical in both cases. 
I suggest creating a default struct that can be mutated and then match on the `with_value`.
It looks like 2 values are changed between the two arms so you can return those into a tuple and then mutate the struct. This way you're not duplicating a lot of code which also increases the cognitive load, reduces DRY-ness.

multisig-wallet/project/contracts/multisig-contract/src/main.sw
```diff
@@ -82,69 +81,61 @@ impl MultiSignatureWallet for Contract {
...
+
+                let transaction_hash = compute_hash(TypeToHash::Transaction(Transaction::new(contract_id(), nonce, target, transaction_parameters)));
+                let approval_count = count_approvals(signatures, transaction_hash);
+                require(storage.threshold.read() <= approval_count, ExecutionError::InsufficientApprovals);
+
+                storage.nonce.write(nonce + 1);
+
+                transfer(value, transfer_params.asset_id, target);
+
+                log(ExecuteTransactionEvent { nonce, target
```
Braqzen, line 135: Might as well move the log to be outside of the match expression, at the end of the function

Braqzen (commented):
Niceu, almosto thereo my friendo

multisig-wallet/project/contracts/multisig-contract/tests/utils/setup.rs
K1-R1: I've done something similar, but have had to account for the fact that the fields that change are in the `transaction_parameters` field, which takes a variant of the enum `TransactionParameters`. Still DRY-er though:
```rust
if with_value {
        transaction.transaction_parameters = TransactionParameters::Call(ContractCallParams {
            function_selector: Bytes(fn_selector!(update_deposit(Address, u64))),
            transfer_params: TransferParams {
                asset_id: base_asset_contract_id(),
                value: Some(DEFAULT_TRANSFER_AMOUNT),
            },
            ..contract_call_params
        });
    }
```

Braqzen (approved):

PR #721, opened 2023-08-12
counter-script docs

SwayStar123:
## Type of change

- Documentation

## Changes

Added documentation according to the src-2 standard

### DIFF
--- a/counter-script/project/contracts/counter/src/main.sw
+++ b/counter-script/project/contracts/counter/src/main.sw
@@ -3,6 +3,7 @@ contract;
 use libraries::Counter;
 
 storage {
+    /// The internal count to be manipulated.
     count: u64 = 0,
 }
 
--- a/counter-script/project/contracts/counter/src/main.sw
+++ b/counter-script/project/contracts/counter/src/main.sw
@@ -3,7 +3,7 @@ contract;
 use libraries::Counter;
 
 storage {
-    /// The internal count to be manipulated.
+    /// Internal counter updated via calls from a script.
     count: u64 = 0,
 }
 
--- a/counter-script/project/scripts/interaction_script/src/main.sw
+++ b/counter-script/project/scripts/interaction_script/src/main.sw
@@ -2,6 +2,16 @@ script;
 
 use libraries::Counter;
 
+/// Calls the methods of the counter contract at the given contract_id.
+///
+/// # Arguments
+///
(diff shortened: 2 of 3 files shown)


counter-script/project/scripts/interaction_script/src/main.sw
```diff
@@ -2,6 +2,16 @@ script;
 
 use libraries::Counter;
 
+/// Calls the methods of the counter contract at the given contract_id.
```
Braqzen, line 5:
```suggestion
/// Calls the functions of the counter contract at the given contract_id.
```

counter-script/project/contracts/counter/src/main.sw
```diff
@@ -3,6 +3,7 @@ contract;
 use libraries::Counter;
 
 storage {
+    /// The internal count to be manipulated.
```
Braqzen, line 6:
```suggestion
    /// Internal counter updated via calls from a script.
```

Braqzen (approved):

counter-script/project/scripts/interaction_script/src/main.sw
SwayStar123: Why?

Braqzen: https://fuellabs.github.io/sway/master/reference/documentation/language/functions/index.html

SwayStar123: https://fuellabs.github.io/sway/master/reference/documentation/examples/counter.html?highlight=Contract%20method#implementation

Braqzen: d84b3210af89d64439c07e7c7b3c073367b834c65c4ed76f4c72a77c87a011d9_1-457759858

Yh, that should be fixed too

SwayStar123: No, the term methods is correct. Even in the rust sdk you do .methods on the contract object to access the methods remember?

Braqzen: It's an associated function. For UX you're going to call it a `.method()` in the SDK.

K1-R1: Moving ahead with method, as per internal discussion

K1-R1 (approved):

PR #738, opened 2023-09-16
DAO bump

SwayStar123:
## Type of change

## Changes

- Update DAO app to forc 0.46.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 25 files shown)


SwayStar123:
Had to change some minor stuff in escrow and multisig because i needed to update rust version. (Clippy added new checks)

Braqzen (approved):
The cross-application changes are minor but I've said it many times that you should avoid intertwining application changes.

CI bumping should really be done on its own and then all the apps that need it can be updated alongside it. Either way, could you bump the forc dates in CI as well?

PR #746, opened 2023-09-18
Timelock-bump

SwayStar123:
## Type of change

## Changes

- Update timelock app to use forc 0.46.0


timelock/project/contracts/timelock-contract/src/utils.sw
```diff
@@ -1,7 +1,28 @@
...
 
 use ::data_structures::Asset;
-use std::{bytes::Bytes, hash::sha256};
+use std::{bytes::Bytes, hash::{Hash, Hasher, sha256}};
+
+impl Hash for Option<Bytes> {
+    fn hash(self, ref mut state: Hasher) {
+        match self {
+            Some(bytes) => bytes.hash(state),
+            None => 0.hash(state),
```
Braqzen, line 10: Unsure about the `0`s here.
I assume that hashing 0 is the equivalent of None somehow and that it's also akin to multiplying by 0. If that is the case then lets save on costs by returning ZERO_B256? Or does that not work?

Braqzen: Secondly, We ought to have this implementation in the std in the future

Braqzen (commented):
Readme versions

timelock/project/contracts/timelock-contract/src/utils.sw
SwayStar123: The hash of 0 is not 0

Braqzen (approved):

PR #747, opened 2023-10-05
Reintroduce the NFT Application with Native Assets

bitzoic:
## Type of change

- New Application

## Changes

The following changes have been made:

- Reintroduces the NFT App using Fuel Native Assets
- Uses the newly introduced standards
- Implements the new token library
- Documents the use of libraries and standards for developers to get an understanding of their use
- CI now runs `forc test` as well as `cargo test`

## Notes

- A separate application for a simple native token is also in development

## Related Issues

Closes #713

PR #748, opened 2023-10-05
Permanently remove old NFT and Fractional NFT App

bitzoic:
## Type of change

## Changes

- With the https://github.com/FuelLabs/sway-applications/pull/747 pull request, the NFT and Fractional NFT apps may be deleted to avoid confusion and prevent developers from using this as a reference.

## Notes

- Nothing about this app should be used.


Braqzen (commented):
Is the new NFT PR supporting the functionality of the F-NFT now that we have sub_id's? The F-NFT with its shares was cool and should be removed after a replacement for it exists.

Check the `.devops` to see if these need to be removed from there too.

bitzoic:
> Is the new NFT PR supporting the functionality of the F-NFT now that we have sub_id's? The F-NFT with its shares was cool and should be removed after a replacement for it exists.
> 
> Check the `.devops` to see if these need to be removed from there too.

The F-NFT will ONLY work with the outdated and bad practice NFT project and a non-existent library. 

There will be a new standard soon^tm where this can be implemented.

Braqzen:
> > Is the new NFT PR supporting the functionality of the F-NFT now that we have sub_id's? The F-NFT with its shares was cool and should be removed after a replacement for it exists.
> > Check the `.devops` to see if these need to be removed from there too.
> 
> The F-NFT will ONLY work with the outdated and bad practice NFT project and a non-existent library.
> 
> There will be a new standard soon^tm where this can be implemented.

Okay, [this](https://github.com/FuelLabs/sway-applications/blob/master/.devops/aurora/apps.txt) needs updating then can approve

Braqzen (approved):

PR #749, opened 2023-10-10
Introduce Token Application

bitzoic:
## Type of change

- New Application

## Changes

- Adds the example token application with implemented libraries and standards to the repo


native-assets/token/README.md
```diff
@@ -0,0 +1,101 @@
...
+    <a href="https://crates.io/crates/fuel-core/0.20.4" alt="fuel-core">
+        <img src="https://img.shields.io/badge/fuel--core-v0.20.4-yellow" />
+    </a>
+</p>
+
+## Overview
+
+A fungible token is a [Native Assets](https://fuellabs.github.io/sway/v0.46.1/book/blockchain-development/native_assets.html) on the Fuel Network.
+
+In this barebones asset example project, there are a maximum of 100,000,000 tokens for each asset that may be minted. There is no limit on the total individual assets a user may mint.
```
K1-R1, line 21:
```suggestion
In this project, there are a maximum of 100,000,000 tokens for each asset that may be minted. There is no limit on the total individual assets a user may mint.
```

K1-R1 (commented):
Cool to se the use of the storage maps to allow many unique assets from one contract. Left a few comments but it's looking good. 
- Can the the `token-contract/tests` dir be removed if unused? This should allow the associated `Cargo` files to be removed too.

bitzoic:
> * Can the the `token-contract/tests` dir be removed if unused? This should allow the associated `Cargo` files to be removed too.

No, removing this will cause CI to fail

K1-R1 (approved):

Issue #714, opened 2023-07-25
Update and reintroduce `fractional-NFT`


K1-R1:
The `fractional-NFT` application was temporarily archived awating the release of the [Sway multi-token standard](https://github.com/FuelLabs/rfcs/pull/17).  Once this standard has been finalised and the version of `forc` that supports this standard is released; the `fractional-NFT` app should be updated and reintroduced.

bitzoic:
This has been unblocked with the introduction of the SRC-6 vault standard https://github.com/FuelLabs/sway-standards/pull/24

PR #768, opened 2024-01-18
Delete AMM App

SwayStar123:
The original author of the AMM app has left the fuel organization, this plus the fact that the sway code has very little comments, and that the tests are written in an abstract way, makes this app very hard to maintain.

## Type of change

- Other (describe below)
- Deprecation of AMM app

## Changes

- Delete the AMM app directory


Braqzen:
"I find it difficult to maintain this" is not a valid reason to delete this.

SwayStar123:
@hummusonrails

hummusonrails (approved):

Braqzen (changes requested):
I haven't blocked anything in a long time but this is unacceptable reasoning and the disregard for the comment is disrespectful.

Blocking until a discussion has been had.

K1-R1:
If bumping the AMM has become a significant blocker, rather than delete I'd advocate that we archive the AMM temporarily as we did the `Fraction-NFT` (#714).  However the intention would then be to reintroduce the AMM at a later date. Perhaps with a refactor of the tests if they have become to difficult to maintain in their current form.
The AMM app is a great example app, especially for the usage of scripts, and I don't think we should delete it.

SwayStar123:
Closed in favour of #769

PR #783, opened 2024-03-22
Reorganize project directories

bitzoic:
## Type of change

## Changes

Changes the project structure for all projects to:
```
My_app
├── src/
│   └────── main.tsx
├── app-contract
│   ├────── src/main.sw
│   └────── tests/harness.rs
├── package.json
├── index.html
├── README.md
└── SPECIFICATION.md
```

This should allow for quicker access to the project's source code and not be buried in a million directories. 

## Notes

- Some projects such as TicTacToe/NFT have been removed from "games"/"native-assets" and moved into the parent directory
- This may break any other open pull requests that depend on this  project structure
- Empty UI folders have been removed


matt-user (changes requested):
This broke the vercel deployment.  I'll request/push changes to fix this.

TicTacToe/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
```
matt-user, line 1: Do we need this file?  seems like all rust stuff should be contained in the contract folder

TicTacToe/Forc.toml
```diff
@@ -0,0 +1,2 @@
+[workspace]
+members = ["./tictactoe-contract"]
```
matt-user, line 2: Do we need this file?

TicTacToe/Cargo.toml
bitzoic: These are included so CI can run `forc b` and `cargo t` in the root of the project's folder without special naming for each project's contracts. You can see this here: https://github.com/FuelLabs/sway-applications/blob/39d19079f8b4a4b147337eed33b2a9e8951023f2/.github/workflows/ci.yml#L43

TicTacToe/Forc.toml
bitzoic: Same as the comment above. While this project just has one contract, some have more such as the AMM.

K1-R1 (approved):

matt-user (approved):

